PR #4, opened 2017-11-09
Synced beam vcfio

mhsaul:
Synced beam vcfio changes to ensure future compatibility with beam distribution.


gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -1,34 +1,40 @@
...
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
```
mhsaul: Should these licenses change?

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -453,180 +500,209 @@ class _VcfSource(_TextSource):
...
-        skip_header_lines=0,
-        header_matcher_predicate=lambda x: x.startswith('#'))
-    self._header_lines_per_file = {}
+    super(_VcfSource, self).__init__(file_pattern,
+                                     compression_type=compression_type,
+                                     validate=validate)
 
-  def process_header_lines_matching_predicate(self, file_name, header_lines):
-    self._header_lines_per_file[file_name] = header_lines
+    self._header_lines_per_file = {}
```
arostamianfar: Maybe remove this unused line from here?

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -453,180 +500,209 @@ class _VcfSource(_TextSource):
...
-      # of whether it uses '|'.
-      if phaseset_from_format or sample.phased:
-        call.phaseset = (str(phaseset_from_format) if phaseset_from_format
-                         else DEFAULT_PHASESET_VALUE)
-      for field in sample.data._fields:
-        # Genotype and phaseset (if present) are already included.
-        if field in (GENOTYPE_FORMAT_KEY, PHASESET_FORMAT_KEY):
+        raise ValueError('Invalid record in VCF file. Error: %s' % str(e))
+
+    def _convert_to_variant_record(self, record, infos, formats):
```
arostamianfar: To confirm: no changes were actually made to this logic? Not sure why git can't figure out that there are actually no diffs :)

gcp_variant_transforms/testing/test_utils.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2017 Google Inc.  All Rights Reserved.
```
arostamianfar: So, we already have a similar idea in "base_test_case_with_temp_dir.py" ... this is used in "gcp_variant_transforms/libs/vcf_header_parser_test.py".
Do you think you can refactor that test to use this new context manager?
Also, please rename to file to something more specific (test_utils.py ends up being a bag of unrelated stuff over time :) ). Maybe just "temp_dir.py" (i think it should be ok given that we're already in /testing).

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: No changes were made to the logic, but the function itself was moved from _VcfSource to _VcfRecordIterator.

gcp_variant_transforms/testing/test_utils.py
mhsaul: Got it! Eventually this will class will also be removed from our repo and pulled from beam.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

PR #5, opened 2017-11-10
Implemented VcfSink

mhsaul:
Implemented transform for writing collections of Variants to .vcf files. Also updated test data and utils to parallel beam.

@arostamianfar


gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -166,6 +168,50 @@ def __ge__(self, other):
...
+    format_keys = ['GT']
+    if self.calls:
+      format_keys.extend([k for k in sorted(self.calls[0].info)])
+    format_field = ':'.join(format_keys)
+
+    calls_values = []
+    for call in self.calls:
+      genotype_field = MISSING_FIELD_VALUE
+      if call.genotype != [MISSING_GENOTYPE_VALUE]:
+        phase_char = '/' if call.phaseset is None else '|'
```
mhsaul: This will need to be updated to include phaseset in info if it is set and not DEFAULT_PHASESET value.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -166,6 +168,50 @@ def __ge__(self, other):
...
+        else:
+          call_info_fields.append(MISSING_FIELD_VALUE)
+
+      calls_values.append(':'.join(call_info_fields))
+
+    calls_field = '\t'.join(calls_values)
+
+    return '\t'.join(str(x) if x else MISSING_FIELD_VALUE for x in [
+        self.reference_name,
+        self.start,
```
mhsaul: We should add 1 to start.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -166,6 +168,50 @@ def __ge__(self, other):
...
     return other <= self
 
+  def to_vcf_line(self):
+    """Converts the variant back to a line that can be written to a .vcf."""
+    info_list = []
+    for k, v in self.info.iteritems():
+      if v.field_count == '1':
+        info_list.append('%s=%s' % (k, str(v.data)))
+      else:
+        info_list.append('%s=%s' % (k, ','.join([str(x) for x in v.data])))
```
mhsaul: end should be added to info in the case of non variants

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -509,6 +485,208 @@ def test_dynamic_work_rebalancing(self):
...
+                                            _get_sample_variant_3())
+
+  def _assert_variant_lines_equal(self, actual, expected):
+    actual_fields = actual.strip().split('\t')
+    expected_fields = expected.strip().split('\t')
+
+    self.assertEqual(len(actual_fields), len(expected_fields))
+    self.assertEqual(actual_fields[0], expected_fields[0])
+    if actual_fields[1] != '.':
+      self.assertEqual(int(actual_fields[1]) + 1, int(expected_fields[1]))
```
mhsaul: Remove the add 1 here once to_vcf_line is refactored

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -509,6 +485,208 @@ def test_dynamic_work_rebalancing(self):
...
+  def test_to_vcf_line(self):
+    variant_1, line_1 = _get_sample_variant_1()
+    self._assert_variant_lines_equal(variant_1.to_vcf_line(), line_1)
+    variant_2, line_2 = _get_sample_variant_2()
+    self._assert_variant_lines_equal(variant_2.to_vcf_line(), line_2)
+    variant_3, line_3 = _get_sample_variant_3()
+    self._assert_variant_lines_equal(variant_3.to_vcf_line(), line_3)
+    empty_variant = vcfio.Variant()
+    empty_line = '\t'.join(['.' for _ in range(8)] + ['GT'])
+    self._assert_variant_lines_equal(empty_variant.to_vcf_line(), empty_line)
```
mhsaul: Add a test case to check END info field behavior.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -166,6 +168,71 @@ def __ge__(self, other):
 
     return other <= self
 
+  def to_vcf_line(self):
```
arostamianfar: Consider moving this to _VcfSink. The reason is that once we add support to genomics protos, then it would be more natural to have the convert logic in one place (we can't add methods to the proto class).

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -166,6 +168,71 @@ def __ge__(self, other):
...
+    return '\t'.join(str(x) if x else MISSING_FIELD_VALUE for x in [
+        self.reference_name,
+        self.start + 1 if self.start else None,
+        ';'.join(self.names),
+        self.reference_bases,
+        ','.join(self.alternate_bases),
+        self.quality,
+        ';'.join(self.filters),
+        info_field,
+        format_field]) + ('\t' + calls_field if calls_field else '') + '\n'
```
arostamianfar: It's cleaner if you have a list of 'columns' where you just return '\t'.join(columns). You can conditionally append things to the list.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -715,3 +847,56 @@ def __init__(
...
   def expand(self, pvalue):
     return pvalue.pipeline | Read(self._source)
+
+
+class WriteToVcf(PTransform):
+  """A PTransform for writing to VCF files."""
+
+  def __init__(self,
+               file_path_prefix,
+               file_name_suffix='',
```
arostamianfar: I think the interface is cleaner if we don't provide this option. It will be .vcf unless compression_type is GZIP/BZIP (in which case we can change it).

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -715,3 +847,56 @@ def __init__(
 
   def expand(self, pvalue):
     return pvalue.pipeline | Read(self._source)
+
+
+class WriteToVcf(PTransform):
+  """A PTransform for writing to VCF files."""
+
+  def __init__(self,
+               file_path_prefix,
```
arostamianfar: considering that we dont provide the suffix option, then can this just be 'file_path'?

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -715,3 +847,56 @@ def __init__(
...
+               file_name_suffix='',
+               num_shards=0,
```
arostamianfar: I think this should be set to 1 by default. People generally expect one VCF file (unless you have a very large output that you like to shard it and join it yourself; but that's likely not common as people would filter them prior to exporting). We have to also see how long it would take to output a ~5GB file.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -715,3 +847,56 @@ def __init__(
...
+               shard_name_template=None,
+               coder=coders.ToStringCoder(),
+               compression_type=CompressionTypes.AUTO,
```
arostamianfar: Considering the above comment, i think this should be set to UNCOMPRESSED by default.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -715,3 +847,56 @@ def __init__(
...
+
+
+class WriteToVcf(PTransform):
+  """A PTransform for writing to VCF files."""
+
+  def __init__(self,
+               file_path_prefix,
+               file_name_suffix='',
+               num_shards=0,
+               shard_name_template=None,
```
arostamianfar: I think this is not necessary and confusing to have as option (it only supports two options and one of them is empty string that is only applicable to 1 shard).

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -715,3 +847,56 @@ def __init__(
...
+        number and shard count respectively.  This argument can be '' in which
+        case it behaves as if num_shards was set to 1 and only one file will be
+        generated. The default pattern used is '-SSSSS-of-NNNNN'.
+      coder: Coder used to encode each line.
+      compression_type: Used to handle compressed output files. Typical value
+          is CompressionTypes.AUTO, in which case the final file path's
+          extension (as determined by file_path_prefix, file_name_suffix,
+          num_shards and shard_name_template) will be used to detect the
+          compression.
+      header: String to write at beginning of file as a header.
```
arostamianfar: Please add more description about what this is and why it's needed (i.e. it should correspond to the VCF header).

arostamianfar: Did you figure out how to handle sample names? Should they be passed in as a separate argument or can we generate them within the pipeline?

gcp_variant_transforms/testing/testdata_util.py
```diff
@@ -22,10 +22,11 @@
...
+  """Returns the full path of the specified ``file_name`` from ``data``."""
+  return os.path.join(
+      os.path.dirname(__file__), '..', 'testing', 'data', 'vcf', file_name)
 
 
 def get_full_dir():
-  """Returns the full path of the  ``testdata`` directory."""
-  return os.path.join(os.path.dirname(__file__), 'testdata')
+  """Returns the full path of the  ``data`` directory."""
+  return os.path.join(os.path.dirname(__file__), '..', 'testing', 'data', 'vcf')
```
arostamianfar: can this just be os.path.join(os.path.dirname(__file__), 'data', 'vcf'? (without the ../testing)?

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: In beam, I think the best way to deal with sample names is to have the user create their own header and pass it in through this parameter. Since we are not sorting the samples at all, if they are working with a single file, the last header line should be the same as the original file.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Good point, keeping the Variant interface as simple as possible makes sense.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Should we actually pass '.vcf' into the super constructor? Or should we just assume the user passes in "filename.vcf" as the argument to `file_path`?

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/testing/testdata_util.py
mhsaul: Good point. I think in Beam this was necessary since the function was written in the io folder.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+    return ';'.join(info_list)
+
+  def _get_variant_format(self, variant):
+    format_keys = ['GT']
+    if variant.calls:
+      # If any calls have a set phaseset that is not `DEFAULT_PHASESET_VALUE`,
+      # the key will be added to the format field.
+      for call in variant.calls:
+        if call.phaseset and call.phaseset != DEFAULT_PHASESET_VALUE:
+          format_keys.append('PS')
```
arostamianfar: Please use PHASESET_FORMAT_KEY

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+    for k, v in variant.info.iteritems():
+      if v.field_count == '1':
+        info_list.append('%s=%s' % (k, str(v.data)))
+      else:
+        info_list.append('%s=%s' % (k, ','.join([str(x) for x in v.data])))
+    return ';'.join(info_list)
+
+  def _get_variant_format(self, variant):
+    format_keys = ['GT']
+    if variant.calls:
```
arostamianfar: Should check 'if not variant.calls: return []'

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+      info_list.append('END=%d' % variant.end)
+    # Set all other fields of info.
+    for k, v in variant.info.iteritems():
+      if v.field_count == '1':
+        info_list.append('%s=%s' % (k, str(v.data)))
+      else:
+        info_list.append('%s=%s' % (k, ','.join([str(x) for x in v.data])))
+    return ';'.join(info_list)
+
+  def _get_variant_format(self, variant):
```
arostamianfar: _get_variant_format_keys is a better name.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+      # If any calls have a set phaseset that is not `DEFAULT_PHASESET_VALUE`,
+      # the key will be added to the format field.
+      for call in variant.calls:
+        if call.phaseset and call.phaseset != DEFAULT_PHASESET_VALUE:
+          format_keys.append('PS')
+          break
+      format_keys.extend([k for k in sorted(variant.calls[0].info)])
+    return ':'.join(format_keys)
+
+  def _get_variant_calls(self, variant):
```
arostamianfar: I think it's more robust for this method to get variant_format_keys as an argument. (you can return a list of keys instead of a string in _get_variant_format_keys). That way, you can loop through the keys for each call and fill the corresponding value. You also won't rely on the sorted order of keys.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+  def _get_variant_format(self, variant):
+    format_keys = ['GT']
+    if variant.calls:
+      # If any calls have a set phaseset that is not `DEFAULT_PHASESET_VALUE`,
+      # the key will be added to the format field.
+      for call in variant.calls:
+        if call.phaseset and call.phaseset != DEFAULT_PHASESET_VALUE:
+          format_keys.append('PS')
+          break
+      format_keys.extend([k for k in sorted(variant.calls[0].info)])
```
arostamianfar: Given that we'd want to pass this as an argument to _get_variant_calls (see below), I think it's more robust to merge all keys. I think it may be that some fields (boolean fields) would be completely missing in some calls.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+    return ':'.join(format_keys)
+
+  def _get_variant_calls(self, variant):
+    calls_values = []
+    alternate_phaseset = False
+
+    for call in variant.calls:
+      if call.phaseset and call.phaseset != DEFAULT_PHASESET_VALUE:
```
arostamianfar: Once we pass in variant_format_keys, this can just become "if PS in keys".

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+        alternate_phaseset = True
+        break
+
+    for call in variant.calls:
+      # Set GT field.
+      genotype_field = MISSING_FIELD_VALUE
+      if call.genotype != [MISSING_GENOTYPE_VALUE]:
+        phase_char = '/' if call.phaseset is None else '|'
```
arostamianfar: Empty phaseset should also be ignored, so please use '|' if call.phaseset else '/'

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+      # Set PS field if there is a set phaseset that is not
+      # `DEFAULT_PHASET_VALUE`.
+      if alternate_phaseset:
+        if call.phaseset and call.phaseset != DEFAULT_PHASESET_VALUE:
+          call_info_fields.append(call.phaseset)
+        else:
+          call_info_fields.append(MISSING_FIELD_VALUE)
+      # Set all other fields of the call info.
+      for k in sorted(call.info):
+        if call.info[k]:
```
arostamianfar: We should check whether call.info[k] is None or empty instead (the value can be zero). Since we need to do the same below, how about creating an isNoneOrEmpty method? There may already be such a thing in some python library.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+        variant.quality,
+        ';'.join(variant.filters),
+        info_field,
+        format_field,
+    ]
+    if calls_field:
+      columns.append(calls_field)
+
+    return '\t'.join(
+        [str(c) if c else MISSING_FIELD_VALUE for c in columns]) + '\n'
```
arostamianfar: Should check with isNoneOrEmpty here as well.

gcp_variant_transforms/beam_io/vcfio.py
arostamianfar: Actually, it looks like we cant' have number=0 for format fields as it needs to be common among all calls. In any case, since we are already looping through all calls, then it may be ok to merge all keys just in case.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
         [str(s) for s in [self.name, self.genotype, self.phaseset, self.info]])
 
 
+class _VcfSink(filebasedsink.FileBasedSink):
+  """A sink for writing VCF files."""
+
+  def __init__(self,
+               file_path,
+               num_shards=1,
+               coder=coders.ToStringCoder(),
```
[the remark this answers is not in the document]
mhsaul: Great idea! The ToStringCoder is converting from str -> str right now that I think of it, so really serves no purpose.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+               compression_type=CompressionTypes.UNCOMPRESSED,
+               headers=None):
+    """Initialize a _TextSink.
+
+    Args:
+      file_path: The file path to write to. The files written will begin
```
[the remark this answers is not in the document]
mhsaul: Good catch, thanks!

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+        Constraining the number of shards is likely to reduce
+        the performance of a pipeline.  Setting this value is not recommended
+        unless you require a specific number of output files.
+      coder: Coder used to encode each line.
+      compression_type: Used to handle compressed output files. Typical value
+        for VCF files is CompressionTypes.UNCOMPRESSED. If set to
+        CompressionTypes.AUTO, file_path's extension will be used to detect
+        compression.
+      headers: A list of VCF meta-information lines describing the INFO,
+        FILTER, and FORMAT entries in each record and a header line describing
```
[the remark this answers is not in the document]
mhsaul: Hmm aren't INFO and FORMAT just required to read the .vcf file back into beam? It might be a bit confusing to include information essential for us to read the file in the sink.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+
+  def _get_variant_info(self, variant):
+    info_list = []
+    # Set end in info if this is a non-variant.
+    if (variant.start and variant.reference_bases and
+        variant.start + len(variant.reference_bases) < variant.end):
+      info_list.append('END=%d' % variant.end)
+    # Set all other fields of info.
+    for k, v in variant.info.iteritems():
+      if v.field_count == '1':
```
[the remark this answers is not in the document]
mhsaul: Sounds good thanks!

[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+class _VcfSink(filebasedsink.FileBasedSink):
+  """A sink for writing VCF files."""
+
+  def __init__(self,
+               file_path,
+               num_shards=1,
+               coder=coders.ToStringCoder(),
+               compression_type=CompressionTypes.UNCOMPRESSED,
+               headers=None):
+    """Initialize a _TextSink.
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+    # Set all other fields of info.
+    for k, v in variant.info.iteritems():
+      if v.field_count == '1':
+        info_list.append('%s=%s' % (k, str(v.data)))
+      else:
+        info_list.append('%s=%s' % (k, ','.join([str(x) for x in v.data])))
+    return ';'.join(info_list)
+
+  def _get_variant_format(self, variant):
+    format_keys = ['GT']
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+
+  def _variant_to_vcf_line(self, variant):
+    """Converts the variant back to a line that can be written to a .vcf."""
+    info_field = self._get_variant_info(variant)
+    format_field = self._get_variant_format(variant)
+    calls_field = self._get_variant_calls(variant)
+
+    columns = [
+        variant.reference_name,
+        variant.start + 1 if variant.start else None,
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+  def open(self, temp_path):
+    file_handle = super(_VcfSink, self).open(temp_path)
+    if self._headers is not None:
+      for header in self._headers:
+        file_handle.write(header)
+    return file_handle
+
+  def _get_variant_info(self, variant):
+    info_list = []
+    # Set end in info if this is a non-variant.
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,145 @@ def __repr__(self):
...
+          call_info_fields.append(str(call.info[k]))
+        else:
+          call_info_fields.append(MISSING_FIELD_VALUE)
+
+      calls_values.append(':'.join(call_info_fields))
+
+    return '\t'.join(calls_values)
+
+  def _variant_to_vcf_line(self, variant):
+    """Converts the variant back to a line that can be written to a .vcf."""
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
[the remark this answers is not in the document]
mhsaul: Since we actual control the value of field_count in _get_field_count_as_string, maybe we should case on our own values?

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,160 @@ def __repr__(self):
...
+
+    for call in variant.calls:
+      # Set GT field.
+      call_info_fields = []
+      for key in format_keys:
+        if key == GENOTYPE_FORMAT_KEY:
+          genotype_field = MISSING_FIELD_VALUE
+          if call.genotype != [MISSING_GENOTYPE_VALUE]:
+            phase_char = '|' if call.phaseset else '/'
+            genotype_field = phase_char.join([str(x) for x in call.genotype])
```
arostamianfar: actually, the missing value can be inside the genotype as well (e.g. "0/.").

gcp_variant_transforms/beam_io/vcfio.py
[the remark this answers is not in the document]
arostamianfar: I see. Good point. I'd still prefer to not include FILTER in this comment as it's a special case and totally optional.

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -509,6 +512,210 @@ def test_dynamic_work_rebalancing(self):
...
+      with open(file_name, 'r') as f:
+        read_result.extend(f.read().splitlines())
+
+    for actual, expected in zip(read_result, self.variant_lines):
+      self._assert_variant_lines_equal(actual, expected)
+
+  def test_write_dataflow_auto_compression(self):
+    pipeline = TestPipeline()
+    pcoll = pipeline | beam.core.Create(self.variants)
+    pcoll | 'Write' >> vcfio.WriteToVcf(self.path + '.gz',  # pylint: disable=expression-not-assigned
```
arostamianfar: You can actually just assign this to '_' to get rid of the lint error...e.g. _ = (pcoll | ...)

gcp_variant_transforms/beam_io/vcfio.py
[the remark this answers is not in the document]
mhsaul: Good point. Fixed it!

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,160 @@ def __repr__(self):
         [str(s) for s in [self.name, self.genotype, self.phaseset, self.info]])
 
 
+class ToVcfRecordCoder(coders.Coder):
+  """Coder for encoding :class:`Variant` objects as .vcf text lines."""
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,160 @@ def __repr__(self):
...
+
+  def _get_variant_info(self, variant):
+    info_list = []
+    # Set END in info if it doesn't match start+len(reference_bases). This is
+    # usually the case for non-variant regions.
+    if (variant.start and variant.reference_bases and
```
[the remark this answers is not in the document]
mhsaul: Good catch!

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,160 @@ def __repr__(self):
...
+        variant.start + len(variant.reference_bases) < variant.end):
```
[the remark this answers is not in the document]
mhsaul: Sounds good! Do we also want to add an `and variant.end`. Since `variant.end` isn't necessary to print vcf lines so this would allow the function to not break when it is None.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,160 @@ def __repr__(self):
...
+      else:
+        info_list.append('%s=%s' % (k, str(v.data)))
+    return info_list
+
+  def _is_alternate_phaseset(self, call):
+    return call.phaseset and call.phaseset != DEFAULT_PHASESET_VALUE
+
+  def _get_variant_format_keys(self, variant):
+    format_keys = []
+    if variant.calls:
```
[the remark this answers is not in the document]
mhsaul: Good call, thanks!

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,160 @@ def __repr__(self):
...
+      format_keys.append(GENOTYPE_FORMAT_KEY)
+      # If any calls have a set phaseset that is not `DEFAULT_PHASESET_VALUE`,
+      # the key will be added to the format field.
+      if any(self._is_alternate_phaseset(call) for call in variant.calls):
+        format_keys.append(PHASESET_FORMAT_KEY)
+
+      for call in variant.calls:
+        format_keys.extend([k for k in call.info if k not in format_keys])
+
+      sort_index = 2 if PHASESET_FORMAT_KEY in format_keys else 1
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,160 @@ def __repr__(self):
...
+
+    return format_keys
+
+  def _field_missing(self, value):
+    """Returns True for empty list, string or None and False otherwise."""
+    return not value and value != 0
+
+  def _get_variant_calls(self, variant, format_keys):
+    calls_values = []
+
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Can we change this condition to `MISSING_GENOTYPE_KEY not in call.genotype` then?

gcp_variant_transforms/beam_io/vcfio_test.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,168 @@ def __repr__(self):
...
+      return format_keys
+
+    def _get_variant_calls(self, variant, format_keys):
+      calls = []
+      for call in variant.calls:
+        call_info = []
+        # GT will always be the first format key.
+        genotype_field = MISSING_FIELD_VALUE
+        if MISSING_GENOTYPE_VALUE not in call.genotype:
```
mhsaul: @arostamianfar: Should these be applied element wise on the list? Like could genotype be [1, -1] in which case we would want to output "1/." instead?

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,156 @@ def __repr__(self):
...
+      # Sort all keys and remove duplicates after GT
+      format_keys[1:] = sorted(list(set(format_keys[1:])))
+
+      return format_keys
+
+    def _get_variant_calls(self, variant, format_keys):
+      calls = []
+      for call in variant.calls:
+        call_info = []
+        # GT will always be the first format key.
```
arostamianfar: Consider adding an assert for this (just in case the logic above changes and we don't know about it).

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,156 @@ def __repr__(self):
...
+        genotype = []
+        for allele in call.genotype:
```
arostamianfar: can also write:
formatted_genotype = [self._format_value(a) if a != MISSING_GENOTYPE_VALUE else MISSING_FIELD_VALUE for a in call.genotype]

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,156 @@ def __repr__(self):
...
+      format_keys[1:] = sorted(list(set(format_keys[1:])))
+
+      return format_keys
+
+    def _get_variant_calls(self, variant, format_keys):
+      calls = []
+      for call in variant.calls:
+        call_info = []
+        # GT will always be the first format key.
+        genotype = []
```
arostamianfar: Consider naming this 'formatted_genotype' to indicate that it's different from call.genotype.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,156 @@ def __repr__(self):
...
+          if allele == MISSING_GENOTYPE_VALUE:
+            genotype.append(MISSING_FIELD_VALUE)
+          else:
+            genotype.append(self._format_value(allele))
+        phase_char = '|' if call.phaseset else '/'
+        call_info.append(phase_char.join(genotype))
+
+        for key in format_keys[1:]:
+          if key == PHASESET_FORMAT_KEY:
+            if self._is_alternate_phaseset(call):
```
arostamianfar: I find it more compact and easier to read to write:
call_info.append(call.phaseset if self._is_alternate_phaseset(call) else MISSING_FIELD_VALUE)

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,156 @@ def __repr__(self):
...
+          ','.join(variant.alternate_bases),
+          variant.quality,
+          ';'.join(variant.filters),
+          ';'.join(info),
+          ':'.join(format_keys),
+      ]
+      if calls:
+        columns.append('\t'.join(calls))
+
+      for i, column in enumerate(columns):
```
arostamianfar: would "columns = [self._format_value(c) for c in columns]" also work?

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +209,156 @@ def __repr__(self):
...
+        call_info = []
+        # GT will always be the first format key.
+        genotype = []
+        for allele in call.genotype:
+          if allele == MISSING_GENOTYPE_VALUE:
+            genotype.append(MISSING_FIELD_VALUE)
+          else:
+            genotype.append(self._format_value(allele))
+        phase_char = '|' if call.phaseset else '/'
+        call_info.append(phase_char.join(genotype))
```
arostamianfar: there's the possibility of genotype itself being an empty list (it won't happen from our imports; but it's a reasonable thing to have). I think it would be fine if you just do: "call_info.append(phase_char.join(genotype) or MISSING_FIELD_VALUE)"

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: This is what I had initially but I found it was more readable to do it this way since the other version doesn't fit on a single line.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Good catch!

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Yes, definitely better!

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: @arostamianfar: I changed this and the other similar comment to be split across multiple lines. Let me know which format is more readable or if there is a better to do the split.

mhsaul:
@arostamianfar Thanks for the review! Please take another look when you get the chance.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Changed this again. I decided to keep them over multiple lines, but extract each into a function so that the getting call function is more readable.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -715,3 +836,46 @@ def __init__(
...
+        CompressionTypes.AUTO, file_path's extension will be used to detect
+        compression.
+      headers: A list of VCF meta-information lines describing the INFO and
+        FORMAT entries in each record and a header line describing the column
+        names. These lines will be written at the beginning of the file.
+    """
+    self._file_path = file_path
+    self._num_shards = num_shards
+    self._compression_type = compression_type
+    self._header = ''.join(headers) if headers else None
```
arostamianfar: Should this be '\n'.join?

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -715,3 +836,46 @@ def __init__(
...
+      num_shards: The number of files (shards) used for output. If not set, the
+        service will decide on the optimal number of shards.
+        Constraining the number of shards is likely to reduce
+        the performance of a pipeline.  Setting this value is not recommended
+        unless you require a specific number of output files.
+      compression_type: Used to handle compressed output files. Typical value
+        for VCF files is CompressionTypes.UNCOMPRESSED. If set to
+        CompressionTypes.AUTO, file_path's extension will be used to detect
+        compression.
+      headers: A list of VCF meta-information lines describing the INFO and
```
arostamianfar: nit: describing at least the INFO and FORMAT ...
They can have more info, but not less :)

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +210,124 @@ def __repr__(self):
...
+
+  def encode(self, variant):
+    """Converts a :class:`Variant` object back to a VCF line."""
+    encoded_info = self._encode_variant_info(variant)
+    format_keys = self._get_variant_format_keys(variant)
+    encoded_calls = self._encode_variant_calls(variant, format_keys)
+
+    columns = [
+        variant.reference_name,
+        variant.start if variant.start is None else variant.start + 1,
```
arostamianfar: nit: I think it's better to rephrase as "variant.start + 1 if variant.start is not None else None" (feels strange to assign it to variant.start if it's None).

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -207,6 +210,124 @@ def __repr__(self):
...
+    assert not format_keys or format_keys[0] == GENOTYPE_FORMAT_KEY
+    encoded_calls = []
+    for call in variant.calls:
+      encoded_call_info = [self._encode_genotype(call.genotype,
+                                                 call.phaseset)]
+      for key in format_keys[1:]:
+        if key == PHASESET_FORMAT_KEY:
+          encoded_call_info.append(self._encode_phaseset(call.phaseset))
+        else:
+          encoded_call_info.append(self._encode_call_info_value(call.info,
```
arostamianfar: nit: I think it's more readable to break it after .append and start a new line at self._encode...

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Good call!

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Hmm, how about calling strip() on each header and then joining with '\n'?

PR #6, opened 2017-11-10
Update vcf_to_bq.py

CNuge:
There was a stray r before the """comment """ on line 15 I removed.

### DIFF
--- a/gcp_variant_transforms/vcf_to_bq.py
+++ b/gcp_variant_transforms/vcf_to_bq.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


CNuge:
Hello google robot, I signed the form!

arostamianfar:
Thanks for the PR! However, the 'r' is actually there by design as the comment includes a backslash. See: https://docs.python.org/2/reference/lexical_analysis.html#string-literals

CNuge:
Ah I see my mistake... thanks for teaching me that!

arostamianfar:
np! thanks for reviewing our code :) please feel free to make PRs if you encounter any issues/bugs. We're just getting started, but I'll write some docs about how to report and fix bugs/features.

PR #9, opened 2017-11-17
Travis file for continuous integration

mhsaul:
Adding travis configuration to run test suite before merging.

@arostamianfar

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: python
+python:
+  - "2.7"
+install:
+  - python setup.py install
+script:
+  - python setup.py test


.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: python
+python:
+  - "2.7"
+install:
+  - python setup.py install
```
arostamianfar: Does this actually work? I haven't been able to get it working on my workstation. It complains about some package version mismatch, but maybe travis handles that somehow?

mhsaul: Hmm which package is failing? It works inside a clean virtual environment on my workstation and the travis build for my forked repo.

arostamianfar: Interesting. Looks like it's fine on travis then. I'll figure out why my instances didnt work later.

PR #12, opened 2017-11-21
Made transform to write Vcf Headers

mhsaul:
Created transform to write VcfHeader objects.

@arostamianfar

### DIFF
--- a/gcp_variant_transforms/beam_io/vcf_header_io.py
+++ b/gcp_variant_transforms/beam_io/vcf_header_io.py
@@ -25,6 +25,8 @@ from apache_beam.io.filesystems import FileSystems
 from apache_beam.io.iobase import Read
 from apache_beam.transforms import PTransform
 
+from gcp_variant_transforms.beam_io import vcfio
+
 __all__ = ['VcfHeader', 'ReadVcfHeaders', 'WriteVcfHeaders']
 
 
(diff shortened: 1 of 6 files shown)


gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -193,3 +194,86 @@ def __init__(
...
+  NUMBER_KEY = 'Number'
+  TYPE_KEY = 'Type'
+  DESCRIPTION_KEY = 'Description'
+
+  def __init__(self, file_path):
+    self._file_path = file_path
+
+  def process(self, header):
+    with FileSystems.create(self._file_path) as file_to_write:
+      self._write_headers_by_key(file_to_write, self.INFO_KEY, header.infos)
```
arostamianfar: I think it's better to treat high-level "key"s as "types" instead (there are too many keys in the header spec, so it's difficult to follow which is which). So, this would be _write_headers_by_type

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -193,3 +194,86 @@ def __init__(
...
+
+  def _write_headers_by_key(self, file_to_write, header_key, header):
+    if header_key == self.INFO_KEY:
+      self._remove_source_and_version(header)
+
+    for header_value in header.values():
+      file_to_write.write(self._to_vcf_header_line(header_key, header_value))
+
+  def _to_vcf_header_line(self, header_key, header):
+    headers = ['{}={}'.format(*self._map_header(header, key)) for key in header]
```
arostamianfar: As mentioned above, it's difficult to follow with nested 'key' variable names. I suggest changing 'header_key' to be 'header_type'.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -193,3 +194,86 @@ def __init__(
...
+      file_to_write.write(self.FINAL_HEADER_LINE)
+
+  def _write_headers_by_key(self, file_to_write, header_key, header):
+    if header_key == self.INFO_KEY:
+      self._remove_source_and_version(header)
+
+    for header_value in header.values():
+      file_to_write.write(self._to_vcf_header_line(header_key, header_value))
+
+  def _to_vcf_header_line(self, header_key, header):
```
arostamianfar: Consider renaming 'header' to something more specific (e.g. 'header_parts_dict').

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -193,3 +194,86 @@ def __init__(
...
+    headers = ['{}={}'.format(*self._map_header(header, key)) for key in header]
+    return self.HEADER_TEMPLATE.format(header_key, ','.join(headers))
+
+  def _map_header(self, header, key):
```
arostamianfar: Consider renaming this method to something more specific (e.g. _get_formatted_header_key_value).

gcp_variant_transforms/beam_io/vcf_header_io.py
arostamianfar: Instead of passing header, can you pass header[key]? and rename the argument to be 'key, value' inside the method?
I think it's more efficient to even use "method(key, value) for (key, value) in header.iteritems()"

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -193,3 +194,86 @@ def __init__(
...
+    value = header[key]
+    key = self._map_key(key)
+    if key == self.NUMBER_KEY:
+      value = self._map_num(value)
+    elif key == self.DESCRIPTION_KEY:
+      value = '"{}"'.format(value)
+    return (key, value)
+
+  def _map_key(self, key):
```
arostamianfar: Please rename to something more specific...._get_formatted_header_key?

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -193,3 +194,86 @@ def __init__(
...
+    elif key == 'num':
+      return self.NUMBER_KEY
+    elif key == 'desc':
+      return self.DESCRIPTION_KEY
+    elif key == 'type':
+      return self.TYPE_KEY
+    else:
+      raise ValueError('Unknown VCF header key {}.'.format(key))
+
+  def _map_num(self, num):
```
arostamianfar: There are also 'G', 'R' types :)
Looks like we need to use this method here: https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/beam_io/vcfio.py#L795 ... I'm ok to copy it here since we'll be merging that function with Beam eventually.
But I think it's better to do this conversion inside _VcfHeaderSource? This way we eliminate all dependencies on PyVCF in one transform. Not sure how easy it is to do it there though..

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -193,3 +194,86 @@ def __init__(
 
   def expand(self, pvalue):
     return pvalue.pipeline | Read(self._source)
+
+
+class _WriteVcfHeaderFn(beam.DoFn):
+  """A DoFn for writing VCF headers to a file."""
+
+  HEADER_TEMPLATE = '##{}=<{}>\n'
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/beam_io/vcf_header_io.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcf_header_io.py
mhsaul: Done.

mhsaul: Done.

gcp_variant_transforms/beam_io/vcf_header_io.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcf_header_io.py
mhsaul: Done.

gcp_variant_transforms/beam_io/vcf_header_io.py
mhsaul: Sounds good. Do you have any sample header lines with 'G' or 'R'? I think it would probably be good to add all variations into the get_sample_vcf_header_lines function so that we consistently test this throughout our code.

arostamianfar: We have them as special cases here: https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/beam_io/vcfio_test.py#L389
Also, instead of hardcoding the values (-1, -2, etc) could you use the same logic as '_get_field_count_as_string' in vcfio? I worry about hardcoding pyvcf's values in our code as they could change in the future and new ones could be added and we would not know about it.

gcp_variant_transforms/testing/testdata_util.py
```diff
@@ -30,3 +30,27 @@ def get_full_file_path(file_name):
 def get_full_dir():
   """Returns the full path of the  ``data`` directory."""
   return os.path.join(os.path.dirname(__file__), 'data', 'vcf')
+
+
+def get_sample_vcf_header_lines():
```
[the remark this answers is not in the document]
mhsaul: Is it okay to turn this into three functions where one returns the combined lines? I think its easier to call different functions than deal with the tuple output.

gcp_variant_transforms/beam_io/vcf_header_io.py
mhsaul: Sounds good! Just made the switch.

PR #19, opened 2017-11-24
Add ReadAllFromVcf to read from massive number of files.

arostamianfar:


gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -883,6 +884,65 @@ def expand(self, pvalue):
...
+      allow_malformed_records (bool): If true, malformed records from VCF files
+        will be returned as :class:`MalformedVcfRecord` instead of failing
+        the pipeline.
+    """
+    super(ReadAllFromVcf, self).__init__(**kwargs)
+    source_from_file = partial(
+        _create_vcf_source, compression_type=compression_type,
+        allow_malformed_records=allow_malformed_records)
+    self._read_all_files = filebasedsource.ReadAllFiles(
+        True, CompressionTypes.AUTO, desired_bundle_size, min_bundle_size,
```
mhsaul: It is a bit unclear what `True` is without reading the source code in ReadAllFiles. I think putting it on its own line with an inline comment `# splittable` would be helpful to readers.

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -492,6 +493,17 @@ def test_pipeline_read_single_file(self):
...
+  def test_pipeline_read_all_single_file(self):
+    with TempDir() as tempdir:
+      file_name = self._create_temp_vcf_file(_SAMPLE_HEADER_LINES +
+                                             _SAMPLE_TEXT_LINES, tempdir)
+      pipeline = TestPipeline()
+      pcoll = (pipeline
+               | 'Create' >> beam.Create([file_name])
+               | 'Read' >> ReadAllFromVcf())
+      assert_that(pcoll, asserts.count_equals_to(len(_SAMPLE_TEXT_LINES)))
+      pipeline.run()
```
mhsaul: These 6 lines are repeated in each test case. Consider moving them to a function like `_assert_read_all_line_count_equal` that takes a file pattern and line count.

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -500,6 +512,16 @@ def test_pipeline_read_single_file_large(self):
...
 
+  @unittest.skipIf(VCF_FILE_DIR_MISSING, 'VCF test file directory is missing')
+  def test_pipeline_read_all_single_file_large(self):
+    pipeline = TestPipeline()
+    pcoll = (pipeline
+             | 'Create' >> beam.Create(
+                 [testdata_util.get_full_file_path('valid-4.0.vcf')])
+             | 'Read' >> ReadAllFromVcf())
+    assert_that(pcoll, asserts.count_equals_to(5))
+    pipeline.run()
```
mhsaul: For the large test it may be better to run on valid-4.1-large.vcf since the 4.0 file isn't much different in size to the sample lines.

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -508,6 +530,38 @@ def test_pipeline_read_file_pattern_large(self):
...
+  def test_pipeline_read_all_multiple_files_large(self):
+    pipeline = TestPipeline()
+    pcoll = (pipeline
+             | 'Create' >> beam.Create(
+                 [testdata_util.get_full_file_path('valid-4.0.vcf'),
+                  testdata_util.get_full_file_path('valid-4.1-large.vcf'),
+                  testdata_util.get_full_file_path('valid-4.2.vcf')])
+             | 'Read' >> ReadAllFromVcf())
+    assert_that(pcoll, asserts.count_equals_to(9900))
+    pipeline.run()
```
mhsaul: Make sure this same functionality is tested even when the vcf directory is missing.

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -508,6 +530,38 @@ def test_pipeline_read_file_pattern_large(self):
...
+
+  @unittest.skipIf(VCF_FILE_DIR_MISSING, 'VCF test file directory is missing')
+  def test_pipeline_read_all_gzip_large(self):
+    pipeline = TestPipeline()
+    pcoll = (pipeline
+             | 'Create' >> beam.Create(
+                 [os.path.join(testdata_util.get_full_dir(), 'valid-*.vcf.gz')])
+             | 'Read' >> ReadAllFromVcf())
+    assert_that(pcoll, asserts.count_equals_to(9900))
+    pipeline.run()
```
mhsaul: Add equivalent test for when vcf directory is missing.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -164,11 +161,22 @@ def run(argv=None):
             'large number of files are specified by input_pattern. '
             'Note that each VCF file must still contain valid header files '
             'even if this is provided.'))
+  parser.add_argument(
+      '--allow_malformed_records',
+      type='bool', default=False, nargs='?', const=True,
+      help=('If true, failed VCF record reads will not raise errors. '
+            'Failed reads will be logged to as warnings and returned as '
```
mhsaul: Remove "to" between "logged to as"

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -164,11 +161,22 @@ def run(argv=None):
...
+            'MalformedVcfRecord objects'))
```
mhsaul: Add "."

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -237,12 +234,20 @@ def run(argv=None):
...
+      variants = (p
+                  | 'InputFilePattern' >> beam.Create(
+                      [known_args.input_pattern])
+                  | 'ReadAllFromVcf' >> vcfio.ReadAllFromVcf(
+                      allow_malformed_records=(
+                          known_args.allow_malformed_records)))
+    else:
+      variants = p | 'ReadFromVcf' >> vcfio.ReadFromVcf(
+          known_args.input_pattern,
+          allow_malformed_records=known_args.allow_malformed_records)
```
mhsaul: Consider moving this logic to a private method `_read_variants`.

gcp_variant_transforms/beam_io/vcfio.py
arostamianfar: Done! Good idea.

gcp_variant_transforms/beam_io/vcfio_test.py
arostamianfar: Done! Good idea!

gcp_variant_transforms/beam_io/vcfio_test.py
arostamianfar: Done.

gcp_variant_transforms/beam_io/vcfio_test.py
arostamianfar: Done.

gcp_variant_transforms/beam_io/vcfio_test.py
arostamianfar: Done.

gcp_variant_transforms/vcf_to_bq.py
arostamianfar: Done.

gcp_variant_transforms/vcf_to_bq.py
arostamianfar: Done.

gcp_variant_transforms/vcf_to_bq.py
arostamianfar: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -883,6 +884,66 @@ def expand(self, pvalue):
     return pvalue.pipeline | Read(self._source)
 
 
+def _create_vcf_source(
+    file_pattern=None, compression_type=None, allow_malformed_records=None):
```
mhsaul: It seems strange that any of these parameters would be optional since we supply them in the transform. This seems to be what was done in `ReadAllFromText` as well though...

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -883,6 +884,66 @@ def expand(self, pvalue):
...
+
+  This transform should be used when reading from massive (>70,000) number of
+  files.
+  """
+
+  DEFAULT_DESIRED_BUNDLE_SIZE = 64 * 1024 * 1024  # 64MB
+
+  def __init__(
+      self,
+      min_bundle_size=0,
```
mhsaul: Currently min_bundle_size is not configurable in _VcfSource. I think we should remove it as a parameter here and always set it to 0 since that is what it will be set to by default in FileBasedSource.

gcp_variant_transforms/testing/temp_dir.py
```diff
@@ -40,20 +44,38 @@ def get_path(self):
...
+  def create_temp_file(
+      self, suffix='', lines=None,
+      compression_type=filesystem.CompressionTypes.UNCOMPRESSED):
     """Creates a temporary file in the temporary directory.
+
     Args:
       suffix (str): The filename suffix of the temporary file (e.g. '.txt')
       lines (List[str]): A list of lines that will be written to the temporary
         file.
+      compression_type (``CompressionTypes``): Enum specifying compression type.
```
mhsaul: This is actually labeled as a (str) in other classes. I think this labeling it as an enum is probably more appropriate to not break the abstraction, but I'm not sure if we should favor that over consistency.

gcp_variant_transforms/testing/temp_dir.py
```diff
@@ -40,20 +44,38 @@ def get_path(self):
...
-    if lines:
-      for line in lines:
-        f.write(line)
+    if not lines:
+      return f.name
+    if compression_type in (filesystem.CompressionTypes.UNCOMPRESSED,
+                            filesystem.CompressionTypes.AUTO):
+      f.write(''.join(lines))
+    elif compression_type == filesystem.CompressionTypes.GZIP:
+      f.close()
```
mhsaul: temp files are a bit weird and can actually be read and written without being closed, so I think this line should be removed. In fact, the API specifies that the file will be destroyed upon being closed, so the current implementation is probably destroying and recreating a normal file with the same name.

gcp_variant_transforms/beam_io/vcfio.py
arostamianfar: yeah, I noticed that too. It's the same in both other ReadAll* files. I assume it has something to do with using partial or the way the method is instantiated. I prefer to keep it like this for now to be consistent with other sources.

gcp_variant_transforms/beam_io/vcfio.py
arostamianfar: Good point. Done.

gcp_variant_transforms/testing/temp_dir.py
arostamianfar: ah, I didn't realize it's a string in other classes. Changes here to be str as well.

gcp_variant_transforms/testing/temp_dir.py
arostamianfar: ah! nice catch! removed.

gcp_variant_transforms/beam_io/vcfio.py
mhsaul: Oh that's a good point. Since we're applying it with the second two parameters.

gcp_variant_transforms/testing/temp_dir.py
```diff
@@ -40,20 +44,37 @@ def get_path(self):
...
                                     dir=self._tempdir,
                                     suffix=suffix)
-    if lines:
-      for line in lines:
-        f.write(line)
+    if not lines:
+      return f.name
+    if compression_type in (filesystem.CompressionTypes.UNCOMPRESSED,
+                            filesystem.CompressionTypes.AUTO):
+      f.write(''.join(lines))
```
mhsaul: I wonder if we should actually change this to always append new lines in a standard way, before running anything we could just do `lines = '\n'.join([line.strip() for line in lines]).

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -164,11 +177,22 @@ def run(argv=None):
...
+      type='bool', default=False, nargs='?', const=True,
+      help=('If true, failed VCF record reads will not raise errors. '
+            'Failed reads will be logged as warnings and returned as '
+            'MalformedVcfRecord objects.'))
+  parser.add_argument(
+      '--optimize_for_large_inputs',
+      type='bool', default=False, nargs='?', const=True,
+      help=('If true, the pipeline runs in optimized way for handling large '
+            'inputs. Set this to true if you are loading more than 50,000 '
+            'files.'))
```
mhsaul: Could we remove this and automatically set it for > 50000 files?

gcp_variant_transforms/testing/temp_dir.py
arostamianfar: Actually, part of the tests were to test '\n', and '\r\n', so I think it's ok to leave as is.

gcp_variant_transforms/vcf_to_bq.py
arostamianfar: yeap :) once we have your header changes merged, we'll do this automatically for >50k input sizes. We'll keep this option for users who know that their input size will be large (it takes a while to get the list of files, which would be a waste of time).

PR #22, opened 2017-11-27
fixed bug in merge_with_non_variants_strategy that duplicated data

mhsaul:
Previously merge_with_non_varaints_strategy would duplicate samples if a record merged with one overlapping record and did not merge with another. This change fixes the bug.


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -65,7 +65,6 @@
 _VARIANT_MERGE_STRATEGIES = [
     _NONE_STRING,
     _MOVE_TO_CALLS_STRING,
-    _MERGE_WITH_NON_VARIANTS_STRING,
```
arostamianfar: nit: please add a short comment that this is under development in the comment above.

mhsaul: Is it okay to just delete the part about merge_with_non_variants and then add it back once things are finished?

arostamianfar: sg :)

PR #25, opened 2017-11-27
Add ReadAllVcfHeaders transform to handle massive number of files.

arostamianfar:


gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -198,6 +200,54 @@ def expand(self, pvalue):
...
+      compression_type (str): Used to handle compressed input files.
+        Typical value is :attr:`CompressionTypes.AUTO
+        <apache_beam.io.filesystem.CompressionTypes.AUTO>`, in which case the
+        underlying file_path's extension will be used to detect the compression.
+    """
+    super(ReadAllVcfHeaders, self).__init__(**kwargs)
+    source_from_file = partial(
+        _create_vcf_header_source, compression_type=compression_type)
+    self._read_all_files = filebasedsource.ReadAllFiles(
+        True,  # splittable
```
mhsaul: Since splitting doesn't support reading a subset of lines in a file, we don't want to split reading of VCF header files.

arostamianfar: Nice catch! Updated.

PR #29, opened 2017-11-29
Added read all vcf headers to pipeline

mhsaul:
Changed pipeline to use ReadAllVcfHeaders if more than 50000 files are input.

### DIFF
--- a/gcp_variant_transforms/vcf_to_bq.py
+++ b/gcp_variant_transforms/vcf_to_bq.py
@@ -34,6 +34,7 @@ python -m gcp_variant_transforms.vcf_to_bq \
 from __future__ import absolute_import
 
 import argparse
+import enum
 import logging
 import re
 import tempfile
@@ -70,9 +71,17 @@ _VARIANT_MERGE_STRATEGIES = [
 ]
 # If the # of files matching the input file_pattern exceeds this value, then
 # headers will be merged in beam.
-_USE_HEADER_MERGER_THRESHOLD = 100
+_SMALL_DATA_THRESHOLD = 100
+_LARGE_DATA_THRESHOLD = 50000
 _MERGE_HEADERS_FILE_NAME = 'merged_headers.vcf'
 _MERGE_HEADERS_JOB_NAME = 'merge-vcf-headers'
+
+
+class PipelineModes(enum.Enum):
+  """An Enum specifying the mode of the pipeline based on the data size."""
+  SMALL = 0
+  MEDIUM = 1
+  LARGE = 2
 
 
 def _get_variant_merge_strategy(known_args):
@@ -157,13 +166,53 @@ def _read_variants(pipeline, known_args):
   return variants
 
 
-def _use_beam_header_merger(known_args):
-  if known_args.representative_header_file:
-    return False
+def _get_pipeline_mode(known_args):
+  """Returns the mode the pipeline should operate in based on input size."""
+  if known_args.optimize_for_large_inputs:
+    return PipelineModes.LARGE
+
   match_results = FileSystems.match([known_args.input_pattern])
   if not match_results:
-    return False
-  return len(match_results[0].metadata_list) > _USE_HEADER_MERGER_THRESHOLD
+    raise ValueError('No files matched input_pattern: {}'.format(
+        known_args.input_pattern))
+
+  total_files = len(match_results[0].metadata_list)
+  if total_files > _LARGE_DATA_THRESHOLD:
+    return PipelineModes.LARGE
+  elif total_files > _SMALL_DATA_THRESHOLD:
+    return PipelineModes.MEDIUM
+
+  return PipelineModes.SMALL
+
+
+def _merge_headers(known_args, pipeline_args, pipeline_mode):
+  """Merges VCF headers using beam based on pipeline_mode."""
+  if (known_args.representative_header_file or
+      pipeline_mode == PipelineModes.SMALL):
+    return
+
+  pipeline_options = GoogleCloudOptions(pipeline_args)
+  if pipeline_options.job_name:
+    pipeline_options.job_name += '-' + _MERGE_HEADERS_JOB_NAME
+  else:
+    pipeline_options.job_name = _MERGE_HEADERS_FILE_NAME
+
+  temp_directory = pipeline_options.temp_location or tempfile.mkdtemp()
+  known_args.representative_header_file = FileSystems.join(
+      temp_directory, _MERGE_HEADERS_FILE_NAME)
+
+  with beam.Pipeline(options=pipeline_options) as p:
+    headers = p
+    if pipeline_mode == PipelineModes.LARGE:
+      headers |= (beam.Create([known_args.input_pattern])
+                  | vcf_header_io.ReadAllVcfHeaders())
+    else:
+      headers |= vcf_header_io.ReadVcfHeaders(known_args.input_pattern)
+
+    _ = (headers
+         | 'MergeHeaders' >> merge_headers.MergeHeaders()
+         | 'WriteHeaders' >> vcf_header_io.WriteVcfHeaders(
+             known_args.representative_header_file))
 
 
 def run(argv=None):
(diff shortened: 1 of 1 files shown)


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -157,13 +158,37 @@ def _read_variants(pipeline, known_args):
   return variants
 
 
-def _use_beam_header_merger(known_args):
+def _beam_merge_headers(known_args, pipeline_args):
```
arostamianfar: nit: just _merge_headers?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -157,13 +158,37 @@ def _read_variants(pipeline, known_args):
...
-    return False
+    return
+
   match_results = FileSystems.match([known_args.input_pattern])
   if not match_results:
-    return False
-  return len(match_results[0].metadata_list) > _USE_HEADER_MERGER_THRESHOLD
+    return None
+
+  paths = [f.path for f in match_results[0].metadata_list]
```
arostamianfar: consider only doing "len(match_results[0].metadata_list)". I'm afraid this would be unnecessary and slow if we're importing 1M+ files.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -157,13 +158,37 @@ def _read_variants(pipeline, known_args):
...
+      pipeline_options.job_name = _MERGE_HEADERS_FILE_NAME
+
+    temp_directory = pipeline_options.temp_location or tempfile.mkdtemp()
+    known_args.representative_header_file = FileSystems.join(
+        temp_directory, _MERGE_HEADERS_FILE_NAME)
+
+    with beam.Pipeline(options=pipeline_options) as p:
+      headers = p
+      if len(paths) > _USE_BEAM_READ_ALL_HEADER_THRESHOLD:
+        headers |= beam.Create(paths) | vcf_header_io.ReadAllVcfHeaders()
```
arostamianfar: this may actually break for very large number of files. I suggest to stick with input_pattern instead of paths

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -157,13 +158,37 @@ def _read_variants(pipeline, known_args):
   return variants
 
 
-def _use_beam_header_merger(known_args):
+def _beam_merge_headers(known_args, pipeline_args):
   if known_args.representative_header_file:
-    return False
+    return
+
   match_results = FileSystems.match([known_args.input_pattern])
```
arostamianfar: we should only do this if '--optimize_for_large_inputs' is false (it takes a while to get the list of files and would be unnecessary if the user has already hinted to use large inputs).

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -157,13 +158,37 @@ def _read_variants(pipeline, known_args):
...
   if not match_results:
-    return False
-  return len(match_results[0].metadata_list) > _USE_HEADER_MERGER_THRESHOLD
+    return None
+
+  paths = [f.path for f in match_results[0].metadata_list]
+  if len(paths) > _USE_BEAM_READ_HEADER_THRESHOLD:
```
arostamianfar: add "or pipeline_args.optimize_for_large_inputs"

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -157,13 +158,37 @@ def _read_variants(pipeline, known_args):
...
+    else:
+      pipeline_options.job_name = _MERGE_HEADERS_FILE_NAME
+
+    temp_directory = pipeline_options.temp_location or tempfile.mkdtemp()
+    known_args.representative_header_file = FileSystems.join(
+        temp_directory, _MERGE_HEADERS_FILE_NAME)
+
+    with beam.Pipeline(options=pipeline_options) as p:
+      headers = p
+      if len(paths) > _USE_BEAM_READ_ALL_HEADER_THRESHOLD:
```
arostamianfar: Consider just basing this on optimize_for_large inputs (see above for setting it if the threshold is higher than some value).

gcp_variant_transforms/vcf_to_bq.py
arostamianfar: Please add a logic to automatically set "pipeline_args.optimize_for_large_inputs = True" if the len is > _USE_BEAM_READ_ALL_HEADER_THRESHOLD

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -70,7 +70,8 @@
 ]
 # If the # of files matching the input file_pattern exceeds this value, then
 # headers will be merged in beam.
-_USE_HEADER_MERGER_THRESHOLD = 100
+_USE_BEAM_READ_HEADER_THRESHOLD = 100
+_USE_BEAM_READ_ALL_HEADER_THRESHOLD = 50000
```
arostamianfar: Consider renaming this to a more general _USE_BEAM_READ_ALL_THRESHOLD (and add a comment about what this does). This can be used for both VCFs and headers.

mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -70,11 +70,19 @@
 ]
 # If the # of files matching the input file_pattern exceeds this value, then
 # headers will be merged in beam.
-_USE_HEADER_MERGER_THRESHOLD = 100
+_USE_BEAM_READ_THRESHOLD = 100
```
arostamianfar: let's rename these to '_SMALL_DATA_THRESHOLD', '_LARGE_DATA_THRESHOLD'

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -70,11 +70,19 @@
...
+_USE_BEAM_READ_ALL_THRESHOLD = 50000
 _MERGE_HEADERS_FILE_NAME = 'merged_headers.vcf'
 _MERGE_HEADERS_JOB_NAME = 'merge-vcf-headers'
 
 
+class PipelineModes(object):
```
arostamianfar: nit: use enum.Enum class.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -157,13 +165,53 @@ def _read_variants(pipeline, known_args):
...
+    return PipelineModes.LARGE
+  elif total_files > _USE_BEAM_READ_THRESHOLD:
+    return PipelineModes.MEDIUM
+
+  return PipelineModes.SMALL
+
+
+def _merge_headers(known_args, pipeline_args, pipeline_mode):
+  """Merges VCF headers using beam based on pipeline_mode."""
+  if (known_args.representative_header_file
```
arostamianfar: nit: we've been putting 'or' , 'and' etc on the previous line as a convention so far.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

PR #30, opened 2017-11-29
Added classes that define vcf_to_bq options

mhsaul:
Created abstract class for adding transform options and several derivatives to encapsulate already implemented options.

### DIFF
--- a/gcp_variant_transforms/options/variant_transform_options.py
+++ b/gcp_variant_transforms/options/variant_transform_options.py
@@ -1,0 +1,213 @@
+# Copyright 2017 Google Inc.  All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+import re
+
+from apache_beam.io.gcp.internal.clients import bigquery
+from apitools.base.py import exceptions
+from oauth2client.client import GoogleCredentials
+
+
+__all__ = ['VariantTransformsOptions', 'VcfReadOptions', 'BigQueryWriteOptions',
+           'FilterOptions', 'MergeOptions']
+
+
+class VariantTransformsOptions(object):
+  """Base class for defining groups of options for Variant Transforms.
+
+  Transforms should create a derived class of ``VariantTransformsOptions``
+  and override the ``add_arguments`` and ``validate`` methods. ``add_arguments``
+  should create an argument group for the transform and add all necessary
+  command line arguments necessary for the transform. ``validate`` should
+  validate the resulting arguments after they are parsed.
+  """
+
+  def add_arguments(self, parser):
+    """Adds all options of this transform to parser."""
+    raise NotImplementedError
+
+  def validate(self, parsed_args):
+    """Validates this group's options parsed from the command line."""
+    pass
+
+
+class VcfReadOptions(VariantTransformsOptions):
+  """Options for reading VCF files."""
+
+  def add_arguments(self, parser):
+    """Adds all options of this transform to parser."""
+    parser.add_argument('--input_pattern',
+                        required=True,
+                        help='Input pattern for VCF files to process.')
+    parser.add_argument(
+        '--allow_malformed_records',
+        type='bool', default=False, nargs='?', const=True,
+        help=('If true, failed VCF record reads will not raise errors. '
+              'Failed reads will be logged as warnings and returned as '
+              'MalformedVcfRecord objects.'))
+    parser.add_argument(
+        '--optimize_for_large_inputs',
+        type='bool', default=False, nargs='?', const=True,
+        help=('If true, the pipeline runs in optimized way for handling large '
+              'inputs. Set this to true if you are loading more than 50,000 '
+              'files.'))
+    parser.add_argument(
+        '--representative_header_file',
+        default='',
+        help=('If provided, header values from the provided file will be used '
+              'as representative for all files matching input_pattern. In '
+              'particular, this will be used to generate the BigQuery schema. '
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -0,0 +1,215 @@
...
+_MOVE_TO_CALLS_STRING = 'MOVE_TO_CALLS'
+_MERGE_WITH_NON_VARIANTS_STRING = 'MERGE_WITH_NON_VARIANTS'
+# List of supported merge strategies for variants.
+# - NONE: Variants will not be merged across files.
+# - MOVE_TO_CALLS: uses libs.variant_merge.move_to_calls_strategy
+#   for merging. Please see the documentation in that file for details.
+_VARIANT_MERGE_STRATEGIES = [
+    _NONE_STRING,
+    _MOVE_TO_CALLS_STRING,
+]
```
mhsaul: What would be the best way to avoid duplication of these globals between our options and the main pipeline?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -37,27 +37,30 @@
...
+from gcp_variant_transforms.libs.variant_merge import move_to_calls_strategy
+from gcp_variant_transforms.options import variant_transform_options
 from gcp_variant_transforms.transforms import filter_variants
+from gcp_variant_transforms.transforms import merge_headers
 from gcp_variant_transforms.transforms import merge_variants
 from gcp_variant_transforms.transforms import variant_to_bigquery
-from gcp_variant_transforms.transforms import merge_headers
+
+OPTIONS = [
+    variant_transform_options.VcfReadOptions(),
```
arostamianfar: I think it's not recommended to initialize classes in module level variables (IIRC, there is even a styleguide violation about this b/c of the way assignment happens for these variables). Consider just setting the module name "variant_transform_options.VcfReadOptions" and initializing it in run (e.g. you can do command_line_options = [option() for option in _COMMAND_LINE_OPTIONS] and then pass that list to your methods).

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -0,0 +1,215 @@
...
+  should create an argument group for the transform and add all necessary
+  command line arguments necessary for the transform. ``validate`` should
+  validate the resulting arguments after they are parsed.
+  """
+  __metaclass__ = abc.ABCMeta
+
+  @abc.abstractmethod
+  def add_arguments(self, parser):
+    """Adds all options of this transform to parser."""
+    pass
```
arostamianfar: It's better to "raise NotImplementedError"

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -0,0 +1,215 @@
...
+
+  @abc.abstractmethod
```
arostamianfar: I think we can remove the abstraction from this method. Some of the options don't have a validate method and it's ok (you can delete the trivial validate method from those classes as well).

gcp_variant_transforms/options/variant_transform_options.py
arostamianfar: I think we can move them as public constants inside the MergeOption class....get_variant_merge_strategy can then use those constants. WDYT?

mhsaul: Sounds good to me. Using the constants in the pipeline is a bit verbose right now though (e.g. `variant_transform_options.MergeOptions.MERGE_WITH_NON_VARIANTS_STRING`). Is there a good way around this?

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -0,0 +1,215 @@
...
+# - NONE: Variants will not be merged across files.
+# - MOVE_TO_CALLS: uses libs.variant_merge.move_to_calls_strategy
+#   for merging. Please see the documentation in that file for details.
+_VARIANT_MERGE_STRATEGIES = [
+    _NONE_STRING,
+    _MOVE_TO_CALLS_STRING,
+]
+
+
+class VariantTransformOptions(object):
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -0,0 +1,215 @@
...
+  """Abstract class for defining groups of options for variant transforms.
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/options/variant_transform_options.py
mhsaul: Done.

gcp_variant_transforms/options/variant_transform_options.py
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -37,27 +37,30 @@
...
 from gcp_variant_transforms.libs.variant_merge import merge_with_non_variants_strategy
+from gcp_variant_transforms.libs.variant_merge import move_to_calls_strategy
+from gcp_variant_transforms.options import variant_transform_options
 from gcp_variant_transforms.transforms import filter_variants
+from gcp_variant_transforms.transforms import merge_headers
 from gcp_variant_transforms.transforms import merge_variants
 from gcp_variant_transforms.transforms import variant_to_bigquery
-from gcp_variant_transforms.transforms import merge_headers
+
+OPTIONS = [
```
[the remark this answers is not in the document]
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
mhsaul: Done.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -87,14 +79,17 @@ class PipelineModes(enum.Enum):
 
 def _get_variant_merge_strategy(known_args):
   if (not known_args.variant_merge_strategy or
-      known_args.variant_merge_strategy == _NONE_STRING):
+      known_args.variant_merge_strategy ==
+      variant_transform_options.MergeOptions.NONE_STRING):
```
arostamianfar: I think you can just create an alias for this in the def.
e.g. merge_options = variant_transform_options.MergeOptions
then do merge_options.NONE_STRING etc

PR #36, opened 2017-12-11
Add special logic for handling unkonwn number of END keys.

arostamianfar:


gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -778,7 +779,17 @@ def _convert_to_variant_record(self, record, infos, formats):
...
+          # of one, which parses into a list in PyVCF.
+          if isinstance(v, list):
+            if len(v) == 1:
+              variant.end = v[0]
+            else:
+              raise ValueError(
+                  'END INFO field has more than one item: {}'.format(
+                      self._last_record))
+          else:
+            variant.end = v
```
mhsaul: Maybe we should move this to a private method like `_get_variant_end`? That way if we ever decide to change the behavior here or the parser changes we can easily refactor.

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -444,10 +444,29 @@ def test_end_info_key(self):
...
+    read_data = self._create_temp_file_and_read_records(
+        [end_info_header_line] + _SAMPLE_HEADER_LINES[1:] + record_lines)
+    self.assertEqual(1, len(read_data))
+    self._assert_variants_equal([variant_1], read_data)
+
+    # Test with invalid number of fields in END (should only have 1 number).
+    record_lines.append('19	124	.	A	.	.	.	END=11,12	GT	1/0	0/1\n')
+    with self.assertRaises(ValueError):
+      self._create_temp_file_and_read_records(
+          [end_info_header_line] + _SAMPLE_HEADER_LINES[1:] + record_lines)
```
mhsaul: I think we should break this into two separate cases so each case tests just one thing.

gcp_variant_transforms/beam_io/vcfio.py
arostamianfar: Good idea. Done.

gcp_variant_transforms/beam_io/vcfio_test.py
arostamianfar: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -843,6 +846,19 @@ def _get_field_count_as_string(self, field_count):
...
+        return end_info_value
+      elif (isinstance(end_info_value, list) and len(end_info_value) == 1 and
+            isinstance(end_info_value[0], (int, long))):
+        # Some VCF files set the number of END info fields to unknown instead
+        # of one, which parses into a list in PyVCF.
+        return end_info_value[0]
+      else:
+        raise ValueError('Invalid END INFO field in record: {}'.format(
+            self._last_record))
+
```
mhsaul: This doesn't need to be done now, but I think we should add a TODO about refactoring convert_to_variant_record. It should probably be broken into several methods like the `_get_variant_end_from_info` like you did here.

PR #37, opened 2017-12-12
Fix path to test vcf file

nmousavi:
Modify Readme file:
  - Fix path to test vcf file
  - Add '&&' so one can run multiple shell cmds at once

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Example command for DirectRunner:
 
 ```bash
 python -m gcp_variant_transforms.vcf_to_bq \
-  --input_pattern gcp_variant_transforms/testing/testdata/valid-4.0.vcf \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,8 +101,8 @@ In addition to using the docker image, you may run the pipeline directly from
 source. First install git, python, pip, and virtualenv:
 
 ```bash
-sudo apt-get install -y git python-pip python-dev build-essential
-sudo pip install --upgrade pip
+sudo apt-get install -y git python-pip python-dev build-essential  &&
+sudo pip install --upgrade pip &&
```
mhsaul: Although the `&&` works, I think its fine to also enter line by line. I'd rather not include it in the docs.

PR #38, opened 2017-12-15
Updating grpcio to pass build

mhsaul:
Updated dependency so package can build.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,9 @@
 import setuptools
 
 REQUIRED_PACKAGES = [
+    # TODO(arostamianfar):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -17,6 +17,7 @@
 import setuptools
 
 REQUIRED_PACKAGES = [
+    'grpcio==1.7.3',
```
arostamianfar: nit: can you please reference https://issues.apache.org/jira/browse/BEAM-3357 in a comment?

mhsaul: Done.

PR #39, opened 2017-12-15
Remove duplicate merge-header logic.

nmousavi:
Remove duplicate merge-header logic. This is achieved by using the merge_headers pipeline with DirectRunner for small inputs. This allows us to modify this logic easier (DRY).

In the next cl, we will modify the merge logic to accept
compatible type conflicts (i.e., int vs double).

### DIFF
--- a/gcp_variant_transforms/libs/vcf_header_parser.py
+++ b/gcp_variant_transforms/libs/vcf_header_parser.py
@@ -32,68 +32,27 @@ __all__ = ['HeaderFields', 'get_merged_vcf_headers']
 HeaderFields = namedtuple('HeaderFields', ['infos', 'formats'])
 
 
-def get_merged_vcf_headers(input_pattern):
-  """Returns merged VCF headers (FORMAT and INFO) from ``input_pattern``.
-
-  If multiple files are specified by ``input_pattern`` then header fields will
-  be merged by key and only one will be chosen as representative (in no
-  particular order). If the same key is used in multiple files, then all types
-  and numbers must be the same.
+def get_vcf_headers(input_file):
+  """Returns VCF headers (FORMAT and INFO) from ``input_file``.
 
   Args:
-    input_pattern (str): A string specifying the path to VCF input file(s). They
-      can be local or remote (e.g. on GCS).
+    input_file (str): A string specifying the path to the representative VCF
+    file, i.e., the VCF file that contains a header representative of all VCF
+    files matching the input_pattern of the job. It can be local or remote (e.g.
+    on GCS).
   Returns:
-    `HeaderFields`` specifying merged header info from all files in
-    ``input_pattern``.
+    ``HeaderFields`` specifying header info.
   Raises:
-    ValueError: If the VCF headers are invalid or incompatible (e.g. the same
-      key if defined with different types in multiple files).
+    ValueError: If ``input_file`` is not a valid VCF file (e.g. bad format,
+    empty, non-existent).
   """
-  merged_info_fields = {}
-  merged_format_fields = {}
-  # The match implementation supports one pattern, but the API technically
-  # supports multiple patterns, hence the need for a list (in both request
-  # and response).
-  match_results = FileSystems.match([input_pattern])
-  if not match_results:
-    return HeaderFields(merged_info_fields, merged_format_fields)
-  for file_metadata in match_results[0].metadata_list:
-    file_name = file_metadata.path
-    if not file_metadata.size_in_bytes:
-      continue  # Ignore empty files.
-    try:
-      vcf_reader = vcf.Reader(fsock=_line_generator(file_name))
-    except SyntaxError as e:
-      raise ValueError('Invalid VCF header: %s' % str(e))
-    _merge_header_fields(merged_info_fields, vcf_reader.infos)
-    _merge_header_fields(merged_format_fields, vcf_reader.formats)
-  return HeaderFields(merged_info_fields, merged_format_fields)
-
-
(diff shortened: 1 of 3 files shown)


gcp_variant_transforms/libs/vcf_header_parser.py
```diff
@@ -32,68 +32,29 @@
...
-  particular order). If the same key is used in multiple files, then all types
-  and numbers must be the same.
+def get_vcf_headers(input_file):
+  """Returns VCF headers (FORMAT and INFO) from ``input_file``.
 
   Args:
-    input_pattern (str): A string specifying the path to VCF input file(s). They
-      can be local or remote (e.g. on GCS).
+    input_file (str): A string specifying the path to a representative VCF
+    file. It can be local or remote (e.g. on GCS).
```
mhsaul: I think it would be good to clarify why the VCF file is representative here. i.e. the VCF file contains a header representative of all VCF files matching the input_pattern of the job.

gcp_variant_transforms/libs/vcf_header_parser.py
```diff
@@ -32,68 +32,29 @@
...
-      source_value = source[key]
-      if (source_value.num != to_merge_value.num or
-          source_value.type != to_merge_value.type):
-        raise ValueError(
-            'Incompatible number of types in header fields: %s, %s' % (
-                source_value, to_merge_value))
-
+  # Return empty if file doesn't exist.
+  if not FileSystems.exists(input_file):
+    return HeaderFields({}, {})
```
mhsaul: It seems better to raise an error in this case because it either means that the merge_headers pipeline failed or the user input the path to their representative header file incorrectly in which case the job is likely to fail later in the pipeline.

gcp_variant_transforms/libs/vcf_header_parser.py
```diff
@@ -32,68 +32,29 @@
...
+  try:
+    vcf_reader = vcf.Reader(fsock=_line_generator(input_file))
+  except SyntaxError as e:
+    raise ValueError('Invalid VCF header: %s' % str(e))
+  except StopIteration:
+    # Return empty if file is empty.
+    return HeaderFields({}, {})
```
mhsaul: This should potentially be an error too. I'm not sure if the schema can properly be generated with empty header fields. Also, if the merge headers pipeline generates an empty file, it means that some error likely occurred before records could be written.

gcp_variant_transforms/libs/vcf_header_parser_test.py
```diff
@@ -78,42 +54,28 @@ def test_invalid_file(self):
...
-    header_fields = vcf_header_parser.get_merged_vcf_headers(file_path)
+    header_fields = vcf_header_parser.get_vcf_headers(file_path)
     self.assertGreater(len(header_fields.infos), 1)
     self.assertGreater(len(header_fields.formats), 1)
 
   def test_non_existent_input_pattern(self):
     expected_header_fields = vcf_header_parser.HeaderFields({}, {})
     self.assertEqual(expected_header_fields,
-                     vcf_header_parser.get_merged_vcf_headers('randompath/*'))
+                     vcf_header_parser.get_vcf_headers('randompath/*'))
```
mhsaul: Since we no longer use an input pattern, I think we should change this to something like `randompath.vcf`

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -134,11 +134,14 @@ def _get_pipeline_mode(known_args):
 
 def _merge_headers(known_args, pipeline_args, pipeline_mode):
   """Merges VCF headers using beam based on pipeline_mode."""
-  if (known_args.representative_header_file or
-      pipeline_mode == PipelineModes.SMALL):
+  if (known_args.representative_header_file):
```
mhsaul: nit: parentheses can be removed.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -134,11 +134,14 @@ def _get_pipeline_mode(known_args):
...
     return
 
   pipeline_options = GoogleCloudOptions(pipeline_args)
+  # Always run pipeline locally if data is small.
+  if (pipeline_mode == PipelineModes.SMALL):
```
mhsaul: nit: remove parentheses.

gcp_variant_transforms/libs/vcf_header_parser.py
nmousavi: Done!

gcp_variant_transforms/libs/vcf_header_parser.py
nmousavi: Agreed. The current implementation returns empty in this case, and I was not sure if this is intended or not, otherwise I would raise an error.  Done!

gcp_variant_transforms/libs/vcf_header_parser.py
nmousavi: Ditto! Done!

gcp_variant_transforms/libs/vcf_header_parser_test.py
nmousavi: Done!

gcp_variant_transforms/vcf_to_bq.py
nmousavi: Done!

gcp_variant_transforms/vcf_to_bq.py
nmousavi: Done!

PR #41, opened 2017-12-20
Resolve Integer vs Float type conflicts in VCF header fields

nmousavi:


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -77,6 +77,12 @@ def add_arguments(self, parser):
               'large number of files are specified by input_pattern. '
               'Note that each VCF file must still contain valid header files '
               'even if this is provided.'))
+    parser.add_argument(
+        '--force_merge_header_conflicts',
+        type='bool', default=False, nargs='?', const=True,
+        help=('If true, all the resolvable conflicts (e.g, Integer vs Float)'
+              'in in the header fields of the given VCF files are resolved.'
+              'If not set, such conflicts raise errors.'))
```
mhsaul: If type conversions do not affect data integrity, we might want to always do these conversions. I would consider defaulting the parameter to True or even removing it altogether to avoid the added complexity.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -28,10 +29,63 @@
...
+  """ A class for resolving mistmatch between header fields
+      (e.g. type, num).
+  """
+
+  # List of all resolvable conflicts with their resolution.
+  # For example, entry [a, b, c] means mistmatch between a and b is resolved to
+  # c (where c is usually either a or b).
+  _RESOLVABLE_CONFLICTS = [
+      # Conflict between 'Integer' and 'Float' is resolved to 'Float'.
+      ['Integer', 'Float', 'Float'],
```
mhsaul: I love the extensibility of this class, but if we don't add any more resolvable conflicts to this list it might be more complex than is necessary.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -28,10 +29,63 @@
...
+    is resolvable.
+
+    Order of args does not matter.
+
+    Args:
+      first_value (string): first given value.
+      second_value (string): second given value.
+    """
+    if not self.can_resolve(first, second):
+      return None
```
mhsaul: I think in this case we should raise the `'Incompatible number or types in header fields: ' {}, {}'.format(source_value, to_merge_value)` error since only resolvable values should be passed to the method.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -117,16 +171,39 @@ def _merge_header_fields(self, source, to_merge):
...
+      merged_value = OrderedDict()
+      for source_field_key, source_field_value in source_value.iteritems():
+        # We only care about mistmach in 'num' and 'type' fields.
+        if (to_merge_value[source_field_key] == source_field_value or
+            source_field_key not in ['num', 'type']):
+          merged_value.update({source_field_key : source_field_value})
+          continue
+        # There is a conflict in header fields. Try to resolve it.
+        can_resolve_conflict = resolver.can_resolve(
+            source_field_value, to_merge_value[source_field_key])
```
mhsaul: With above change this value can be removed. This if statement can just be `if force_merge_conflicts`

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -117,16 +171,39 @@ def _merge_header_fields(self, source, to_merge):
...
+        if force_merge_conflicts and can_resolve_conflict:
+          resolution_field_value = resolver.resolve(
+              source_field_value, to_merge_value[source_field_key])
+          merged_value.update({source_field_key : resolution_field_value})
+        elif not force_merge_conflicts and can_resolve_conflict:
```
mhsaul: This statement can become `elif resolver.can_resolve(source_field_value, to_merge_value[source_field_key])`.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -61,7 +115,7 @@ def __init__(self,
     self.formats = self._values_asdict(formats or {})
     self.contigs = self._values_asdict(contigs or {})
 
-  def update(self, to_merge):
+  def update(self, to_merge, force_merge_conflicts):
```
mhsaul: We can have `force_merge_conflicts` default to False so that test cases do not need to specify it.

gcp_variant_transforms/transforms/merge_headers_test.py
```diff
@@ -102,19 +103,110 @@ def test_combine_multiple_headers_as_accumulators(self):
...
+    combiner_fn = merge_headers._MergeHeadersFn(force_merge_conflicts=False)
+
+    try:
+      merged_headers = combiner_fn.create_accumulator()
+      merged_headers = combiner_fn.add_input(merged_headers, headers_1)
+      merged_headers = combiner_fn.add_input(merged_headers, headers_2)
+      merged_headers = combiner_fn.extract_output(merged_headers)
+      self.fail('Incompatible VCF headers must throw an exception.')
+    except ValueError:
+      pass
```
mhsaul: instead of doing the try except for value error, you can write `with self.assertRaises(ValueError):` and then have the rest of the calls nested within there.

gcp_variant_transforms/beam_io/vcf_header_io.py
nmousavi: Asha mentioned another type of conflicts we want to resolve, e.g., 'num' fields with any value greater than 1 are compatible. I plan to add regex to this list for this case (in a separate cl). Apart from these two cases, I am not sure if there is more.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -28,10 +29,63 @@
...
+    Args:
+      first_value (string): first given value.
+      second_value (string): second given value.
+    """
+    return (first == second or
+            (first in self._resolvable_conflicts_dict and
+             second in self._resolvable_conflicts_dict[first]))
+
+  def resolve(self, first, second):
+    """ Returns the resolution if the conflict between `first` and `second`
```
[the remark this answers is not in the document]
nmousavi: Done!

gcp_variant_transforms/beam_io/vcf_header_io.py
nmousavi: Done!

gcp_variant_transforms/beam_io/vcf_header_io.py
nmousavi: Done!

gcp_variant_transforms/beam_io/vcf_header_io.py
nmousavi: Done!

gcp_variant_transforms/beam_io/vcf_header_io.py
nmousavi: Done!

gcp_variant_transforms/options/variant_transform_options.py
nmousavi: That's a good point. I am not sure why Asha suggested having a flag for this. Should we keep it as is and change it later once Asha is back?

gcp_variant_transforms/transforms/merge_headers_test.py
nmousavi: Done!

gcp_variant_transforms/options/variant_transform_options.py
mhsaul: Keeping it as I think is okay for now. Maybe make a note or TODO or github issue of it to address it after break.

nmousavi: Done! PTAL.

PR #51, opened 2018-01-10
Fix some PEP8 style guide violations and other warnings in IntelliJ

bashir2:
In addition to fixing these warnings, this change does the followings:
* Adds a Project_Default.xml file under the .idea directory such that future setup of IntelliJ is easier.
* Adds some TODOs for fixes that are less trivial.
* Adds tests for a bug that was earlier fixed when pylint was enabled.


bashir2:
I signed it!

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,21 @@
+<component name="InspectionProjectProfileManager">
```
arostamianfar: Please add a README file in this dir (or upstream) to indicate what this config is for and where the spec is defined.

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,21 @@
...
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
```
arostamianfar: nit: is it possible to change myName to something more meaningful?

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,21 @@
...
+    <inspection_tool class="PyAbstractClassInspection" enabled="false" level="WEAK WARNING" enabled_by_default="false" />
+    <inspection_tool class="PyMethodMayBeStaticInspection" enabled="false" level="WEAK WARNING" enabled_by_default="false" />
+    <inspection_tool class="PyPep8Inspection" enabled="true" level="WEAK WARNING" enabled_by_default="true">
+      <option name="ignoredErrors">
+        <list>
+          <option value="E111" />
```
arostamianfar: Are these error codes documented anywhere? Please add a link to it in the README file as well.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -346,7 +347,7 @@ def __init__(self, file_path):
     self._file_path = file_path
     self._file_to_write = None
 
-  def process(self, header):
+  def process(self, header, *args, **kwargs):
```
arostamianfar: I don't think we should add args/kwargs. These are used for optional side-inputs (they get replaced with appropriate variables, see [here](https://beam.apache.org/documentation/programming-guide/#passing-side-inputs-to-pardo)) and none of the existing python implementations (even outside of this project) have them.

gcp_variant_transforms/libs/variant_merge/variant_merge_strategy.py
```diff
@@ -20,7 +20,7 @@
 class VariantMergeStrategy(object):
   """Interface for a variant merge strategy."""
 
-  def get_merged_variants(self, variants):
+  def get_merged_variants(self, variants, key):
```
arostamianfar: nice catch! Could you please add 'key' to the args section as well? It's a string representing the key used to merge ``variants``.

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -137,6 +136,8 @@ def __enter__(self):
     self.dataset_id = 'integration_tests_{}'.format(
         datetime.now().strftime('%Y%m%d_%H%M%S'))
     dataset_ref = client.dataset(self.dataset_id)
+    # TODO(bashir): This does not seem to be right, the client argument is
```
arostamianfar: I haven't run this myself yet, but looks like this has the same structure as https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/cloud-client/quickstart.py ?

gcp_variant_transforms/transforms/filter_variants.py
```diff
@@ -60,4 +60,7 @@ def _apply_filters(self, variant):
       yield variant
 
   def expand(self, pcoll):
+    # TODO(bashir): It seems there is a type check violation here as the
```
arostamianfar: This actually gets converted to a [CallableWrapperDoFn](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/transforms/core.py#L394). This style is sometimes used if the function is trivial and doesn't have any context. Python magic :)

.idea/inspectionProfiles/Project_Default.xml
bashir2: Done. I just added some notes on how this is created but did not bother to link to IntelliJ documentation for the spec (because it is not hand made).

.idea/inspectionProfiles/Project_Default.xml
bashir2: Done.

.idea/inspectionProfiles/Project_Default.xml
bashir2: This file is generated by IntelliJ, I tried to rename the config (which changes "Project Default") but it creates other files so I prefer to keep it as simple as possible. If you feel strongly about it, I will try to figure how this can be fixed.

gcp_variant_transforms/beam_io/vcf_header_io.py
bashir2: I realized that we don't use them in our code but without these, we get a warning that this signature "does not match signature of base method" and that is because DoFn.process has them. So from a pure base class implementation perspective I thought they should be there (BTW, is there any harm in having them?). I do see that even in Beam documentation examples, they are not always present so I removed them.

gcp_variant_transforms/libs/variant_merge/variant_merge_strategy.py
bashir2: Done.

gcp_variant_transforms/testing/integration/run_tests.py
bashir2: Updated the TODO per our offline discussion. For the record, the issue is the incompatibility between different versions of bigquery API.

gcp_variant_transforms/transforms/filter_variants.py
bashir2: I see, apparently IntelliJ code inspectors did not realize this magic :-)
Looking closer at this, I now see that [the constructor of the base class PTransformWithSideInputs](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/transforms/ptransform.py#L520) explicitly does this conversion. I think the problem is the [docstring of the ParDo constructor](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/transforms/core.py#L763) which says it expects a DoFn (which is not necessarily true).

bashir2:
I am not sure when/how GitHub sends a notification email to the reviewers telling them that their comments are addressed; but Asha if you see this, please let me know.

.idea/inspectionProfiles/Project_Default.xml
arostamianfar: Sounds good. Thanks!

.idea/inspectionProfiles/Project_Default.xml
arostamianfar: ah, it's ok then.

.idea/inspectionProfiles/README.md
```diff
@@ -0,0 +1,6 @@
+The `Project_Default.xml` file is generated by IntelliJ and is included here
+for easier workspace setup in future. It is mostly added for the list of
+Python inspetion checks that are disabled.
```
arostamianfar: nit: inspection

gcp_variant_transforms/transforms/filter_variants.py
arostamianfar: Please remove the TODO?

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -137,6 +136,11 @@ def __enter__(self):
     self.dataset_id = 'integration_tests_{}'.format(
         datetime.now().strftime('%Y%m%d_%H%M%S'))
     dataset_ref = client.dataset(self.dataset_id)
+    # TODO(bashir): There is a mismatch between the bigquery API versions
```
arostamianfar: nit: should this be bashir2?

gcp_variant_transforms/transforms/merge_headers.py
```diff
@@ -26,16 +26,16 @@ class _MergeHeadersFn(beam.CombineFn):
   def create_accumulator(self):
     return vcf_header_io.VcfHeader()
 
-  def add_input(self, source, to_merge):
+  def add_input(self, source, to_merge, *args, **kwargs):
```
arostamianfar: Please also remove args/kwargs from here. This is also not a common convention (e.g. https://beam.apache.org/documentation/programming-guide/#advanced-combinations-using-combinefn)

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, split_alternate_allele_info_fields=True):
     self._split_alternate_allele_info_fields = (
         split_alternate_allele_info_fields)
 
-  def process(self, record):
+  def process(self, record, *args, **kwargs):
```
arostamianfar: ditto here.

gcp_variant_transforms/beam_io/vcf_header_io.py
arostamianfar: I prefer not having them as they are distracting and not a convention used anywhere else. You can consider disabling this warning at the cost of potentially masking other legit warnings (I'm ok with disabling if the warnings are too annoying).

gcp_variant_transforms/transforms/merge_headers.py
bashir2: Done

gcp_variant_transforms/transforms/variant_to_bigquery.py
bashir2: Done

bashir2:
It seems some of my previous changes were not pushed to the review branch, sorry about that. I am actually in the process of reworking my git workflow as it seems the GitHub flow is too different than the git5/citc flow I am used to. I may create a new review branch to squash all the commits in this pull request (BTW, can we enable [commit squashing](https://github.com/blog/2141-squash-your-commits) for this repo?). I will ping you again once this is ready to take another look at.

bashir2:
So once #57 is merged, I will close this pull request as it is not needed anymore.

Issue #62, opened 2018-01-13
Fix and enable continuous run of the integration tests.


bashir2:
There is currently a version of integration tests implemented (see [run_tests.py](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/testing/integration/run_tests.py)) and at some point in the past it has worked. But currently it does not run. We need to fix it, improve the tests and add some sort of validation (currently it does not validate the generated table). Finally we should implement some sort of continuous run for this integration test (it is not a small test).

bashir2:
All the required changes to fix integration tests, add some minimal verifications to them, and run them automatically on every commit are done. Once the documentation changes of PR #97 is merged, this issue will be closed. The remaining nice to have but not critical enhancements are tracked in Issue #98.

PR #63, opened 2018-01-13
Added dependencies to satisfy googleapiclient modules.

bashir2:
While running run_tests.py, it complained about missing googleapiclient package. This is only to fix the dependency issues (part of fixing #62).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,16 @@ REQUIRED_PACKAGES = [
     'apache-beam[gcp]>=2.2',
     'intervaltree>=2.1.0,<2.2.0',
     'pyvcf<0.7.0',
+    'google-api-python-client>=1.6',
+    # TODO(bashir2): It seems that setuptools cannot find the intersection of
+    # dependency requirements. For example the following dependencies are
+    # listed only because installing google-api-python-client brings versions
+    # of these packages that are newer than versions required in apache-beam
+    # or its dependencies. I think there should be a better of doing this that
+    # I am not aware of!
+    'six<1.11',
+    'oauth2client<4.0.0',
+    'httplib2<0.10',
     ]
 
 REQUIRED_SETUP_PACKAGES = [


setup.py
```diff
@@ -23,6 +23,16 @@
     'apache-beam[gcp]>=2.2',
     'intervaltree>=2.1.0,<2.2.0',
     'pyvcf<0.7.0',
+    'google-api-python-client>=1.6',
+    # TODO(bashir2): It seems that setuptools cannot find the intersection of
+    # dependency requirements. For example the following dependencies are
+    # listed only because installing google-api-python-client brings versions
+    # of these packages that are newer than versions required in apache-beam
+    # or its dependencies. I think there should be a better of doing this that
+    # I am not aware of!
```
bashir2: @arostamianfar please comment if you know a better way for doing this; thanks.

arostamianfar: I don't think these are needed. I was able to run the tests on a clean VM after fixing some other issues (although, table deletion still fails at the end of the run). Sent https://github.com/googlegenomics/gcp-variant-transforms/pull/70 . Added some more TODOs for you as well :)

bashir2: Hmm, then I might be missing something here, because I also tried in a clean VM and got similar errors as before (in my clean git directory on my local machine). Here are the steps:

- Getting a new VM on GCP and SSH into it.
- $ sudo apt-get install git
- $ git clone https://github.com/googlegenomics/gcp-variant-transforms
- $ cd gcp-variant-transforms/
- $ sudo apt-get install python-pip python-dev build-essential
- $ sudo pip install --upgrade pip
- $ sudo pip install --upgrade virtualenv
- $ virtualenv venv
- $ . venv/bin/activate
- $ python setup.py install

Running tests (python setup.py test) works fine at this point but running the integration test (with no command line options just to get the import error) gives me this:

$ python gcp_variant_transforms/testing/integration/run_tests.py
Traceback (most recent call last):
  File "gcp_variant_transforms/testing/integration/run_tests.py", line 25, in <module>
    from googleapiclient import discovery
ImportError: No module named googleapiclient

How do you get googleapiclient module in your setup?

arostamianfar: ah, sorry for the confusion... I only meant the additional dependencies related to your TODO here. We definitely need googleapiclient and bigquery api (I added these in #70), but I don't think we need others. I think you can close this PR and continue once #70 is merged (I decided to run the tests myself and ended up fixing some of the other issues as well).

bashir2: So for future reference, I created Issue #71 and updated the development guide as well. Closing this PR and ignoring the dependency conflicts for now.

PR #70, opened 2018-01-15
Fix basic errors in integration/run_tests.py

arostamianfar:
Note: The tests still do not run successfully e2e (table deletion fails when cleaning up), but they progress further :)

Issue #62

### DIFF
--- a/gcp_variant_transforms/testing/integration/run_tests.py
+++ b/gcp_variant_transforms/testing/integration/run_tests.py
@@ -12,19 +12,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Integration testing runner."""
+r"""Integration testing runner for Variant Transforms' VCF to BigQuery pipeline.
+
+You may run this test in any project (the test files are publicly accessible).
+Execute the following command from the root source directory:
+python gcp_variant_transforms/testing/integration/run_tests.py \
+  --project myproject \
+  --staging_location gs://mybucket/staging \
+  --temp_location gs://mybucket/temp \
+  --logging_location gs://mybucket/temp/integration_test_logs
+
+By default, it runs all integration tests inside
+`gcp_variant_transforms/testing/integration/`.
+"""
 
 import argparse
 import json
 import multiprocessing
 import os
+import time
 
 from datetime import datetime
+# TODO(bashir2): Figure out why pylint can't find this.
+# pylint: disable=no-name-in-module,import-error
 from google.cloud import bigquery
 
 from googleapiclient import discovery
 from oauth2client.client import GoogleCredentials
-import time
+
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -21,6 +21,8 @@
     # This is needed due to https://issues.apache.org/jira/browse/BEAM-3357.
     'grpcio>=1.3.5,<=1.7.3',
     'apache-beam[gcp]>=2.2',
+    'google-api-python-client>=1.6',
```
bashir2: nit: Do you mind adding a note with a reference to Issue #71 here?

setup.py
```diff
@@ -21,6 +21,8 @@
...
+    'google-cloud-bigquery>=0.25',
```
bashir2: Are you sure that version 0.25 works or should this be > 0.25 (see my TODO above)?

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -129,12 +146,13 @@ def __init__(self, args):
...
     self.credentials = GoogleCredentials.get_application_default()
-    client = bigquery.Client(project=self.project)
     self.dataset_id = 'integration_tests_{}'.format(
         datetime.now().strftime('%Y%m%d_%H%M%S'))
+    self.dataset = None
+
+  def __enter__(self):
+    client = bigquery.Client(project=self.project)
     dataset_ref = client.dataset(self.dataset_id)
     # TODO(bashir2): There is a mismatch between the bigquery API versions
```
bashir2: I think you can remove this TODO now. BTW, I thought we need a version strictly greater than 0.25.

arostamianfar: Done. And good catch! I changed to be >0.25.

setup.py
arostamianfar: Nice catch. Done. It installed v0.29.0 on my machine.

setup.py
arostamianfar: Done.

setup.py
```diff
@@ -21,6 +21,11 @@
     # This is needed due to https://issues.apache.org/jira/browse/BEAM-3357.
     'grpcio>=1.3.5,<=1.7.3',
     'apache-beam[gcp]>=2.2',
+    # Note that adding 'google-api-python-client>=1.6' causes some dependency
+    # mismatch issues if using 'setup.py install'. See Issue #71.
```
bashir2: nit: dependency issues are there no matter we use `setup.py install` or not, the difference seems to be that `setup.py install` stops while `pip install` continues but still installs conflicting versions.

arostamianfar: Done.

PR #75, opened 2018-01-21
A script for deploying and integration test running from the local client.

bashir2:
Please note that there are two main scenarios for running integration tests:
* First is before the code is submitted. Ideally this should be done as a presubmit check (e.g., by Travis) but it seems there are many credential setting issues. So instead, I have taken the approach of this PR.

* Second is to make sure that the code on the master branch is in a healthy condition. For that, I am also working on [Build Triggers](https://cloud.google.com/container-builder/docs/running-builds/automate-builds) but that is a separate change.

Tested:
Successfully ran:
./deploy_and_run_integration_tests.sh
./deploy_and_run_integration_tests.sh --keep_image --keep_tables --skip_build --image_tag test_2018-01-23-20-15-54

Issue: #62

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -15,13 +15,15 @@
 # Builds docker image for gcp-variant-transforms:
 # Run using:
 # $ gcloud container builds submit --config cloudbuild.yaml --timeout 1h .
+substitutions:
+  _CUSTOM_TAG_NAME: 'latest'
 steps:
   - name: 'gcr.io/cloud-builders/docker'
     args:
       - 'build'
-      - '--tag=gcr.io/$PROJECT_ID/gcp-variant-transforms'
+      - '--tag=gcr.io/$PROJECT_ID/gcp-variant-transforms:${_CUSTOM_TAG_NAME}'
       - '--file=docker/Dockerfile'
       - '.'
     dir: '.'
 images:
-  - 'gcr.io/$PROJECT_ID/gcp-variant-transforms'
+  - 'gcr.io/$PROJECT_ID/gcp-variant-transforms:${_CUSTOM_TAG_NAME}'
--- a/deploy_and_run_integration_tests.sh
+++ b/deploy_and_run_integration_tests.sh
@@ -1,0 +1,187 @@
+#!/bin/bash
+set -euo pipefail
+
+# Copyright 2018 Google Inc.  All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This script builds a Docker image from the current state of the local code,
+# pushes the image to a test registry (project 'gcp-variant-transforms-test')
+# and runs integration tests against that image.
+#
+# To run this test successfully:
+# - The user's gcloud credentials should be set; follow the steps at:
+#   https://cloud.google.com/genomics/install-genomics-tools
+#
+# - The user should have access to the 'gcp-variant-transforms-test' project.
+
+PROJECT="gcp-variant-transforms-test"
+GS_DIR="integration_test_runs"
+GREEN="\e[32m"
+RED="\e[31m"
+TEST_ARGUMENTS=""
+KEEP_IMAGE_OPT="--keep_image"
+IMAGE_TAG_OPT="--image_tag"
+SKIP_BUILD_OPT="--skip_build"
+image_tag=""
+keep_image=""
+skip_build=""
+
+#################################################
+# Prints a given message with a color.
+# Arguments:
+#   $1: The message
+#   $2: The text for the color, e.g., "\e[32m" for green.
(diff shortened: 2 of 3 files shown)


deploy_and_run_integration_tests.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
```
arostamianfar: Please add "set -euo pipefail" to be more robust.

deploy_and_run_integration_tests.sh
```diff
@@ -0,0 +1,95 @@
...
+fi
+docker_file=$(ls docker/Dockerfile)
+echo ${docker_file}
+if [[ ${docker_file} != "docker/Dockerfile" ]]; then
+  echo "Cannot find docker/Dockerfile!"
+  exit
+fi
+
+# Building the image
+time_tag=$(date +%F-%H-%M-%S)
```
arostamianfar: Consider making this a more general flag (e.g. TAG_NAME) that defaults to current time.

deploy_and_run_integration_tests.sh
```diff
@@ -0,0 +1,95 @@
...
+  echo -e "${GREEN}Success! ${NC}"
+else
+  echo -e "${RED}FAILED! ${NC}"
+fi
+
+# Removing the image and virtualenv int_test_venv
+echo "Deleting the test image from the registry"
+# TODO(bashir2): Find a way to mark these images as temporary such that they are
+# garbage collected automatically if the test fails before this line.
+gcloud container images delete --quiet ${image_tag}
```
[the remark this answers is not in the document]
arostamianfar: And I think this should be optional (maybe even keep by default?)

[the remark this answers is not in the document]
bashir2: I added the trap for virtualenv but decided to leave the image deletion outside. The reason being that if this fails for whatever reason and the image is created already (which takes a few minutes to be created) it makes sense not to delete the image on exit as it probably needed even if the --keep_image option is not provided.

bashir2:
BTW PTAL :-)

deploy_and_run_integration_tests.sh
```diff
@@ -0,0 +1,187 @@
...
+#################################################
+# Prints the usage.
+#################################################
+usage() {
+  echo "Usage: $0 [${KEEP_IMAGE_OPT}] [${IMAGE_TAG_OPT} image_tag] "
+  echo "    [${SKIP_BUILD_OPT}] [... test script options ...]"
+  echo "  With ${KEEP_IMAGE_OPT} the tested image is not deleted at the end."
+  echo "  If ${IMAGE_TAG_OPT} is not set, a new image with the current time"
+  echo "    tag is created and used in tests, otherwise the given tag is used."
+  echo "  ${SKIP_BUILD_OPT} skips the build step so it has to be used with "
```
nmousavi: Can we drop this flag and instead build iff 'image_tag' is empty? In other words, what's the use case for having non_empty 'image_tag' and skip_build=False in the integration test.

deploy_and_run_integration_tests.sh
```diff
@@ -0,0 +1,187 @@
...
+fi
+
+# Running integration tests in a temporary virtualenv
+temp_dir="$(mktemp -d)"
+color_print "Setting up integration test environment in ${temp_dir}" "${GREEN}"
+# Since we have no prompt we need to disable prompt changing in virtualenv.
+export VIRTUAL_ENV_DISABLE_PROMPT="something"
+virtualenv "${temp_dir}"
+source ${temp_dir}/bin/activate;
+trap clean_up_virtualenv EXIT
```
nmousavi: do we need to clean up images here?

cloudbuild.yaml
```diff
@@ -15,13 +15,15 @@
 # Builds docker image for gcp-variant-transforms:
 # Run using:
 # $ gcloud container builds submit --config cloudbuild.yaml --timeout 1h .
+substitutions:
```
arostamianfar: Nice!

deploy_and_run_integration_tests.sh
```diff
@@ -0,0 +1,187 @@
...
+  echo "  This script should be run from the root of source tree."
+}
+
+#################################################
+# Makes sure that we are in the root of the source tree and cloudbuild.yaml
+# file exists.
+#################################################
+check_dir() {
+  dir_ok="$(pwd | sed -e 's/.*gcp-variant-transforms$/MATCHED/')"
+  if [[ ${dir_ok} != "MATCHED" ]]; then
```
arostamianfar: nit: please also quote  ${dir_ok} ... there are a few other missing quotes below as well.

deploy_and_run_integration_tests.sh
bashir2: I agree with you that in the integration test scenarios (e.g., before sending a PR) the --image_tag is enough but I can imagine other scenarios that this is useful. For example, when we want to make a release candidate, we may want to use this script, tag the image as RC (rather than with a test time tag) _but also_ build it.

deploy_and_run_integration_tests.sh
bashir2: Done.
I fixed a few more cases but please note that not all cases should be quoted, for example ${TEST_ARGUMENTS} in the run_test.py call has to be unquoted, or when we are comparing numbers.

deploy_and_run_integration_tests.sh
bashir2: Done.
When Asha suggested to have the trap I thought I leave the image clean-up out because when something goes wrong, it is very likely that we need the image (to test again). But now that you also suggested it, I think it makes more sense to more people to delete it, so I did it :-)

PR #95, opened 2018-01-24
Enable coveralls.

arostamianfar:


.travis.yml
```diff
@@ -4,5 +4,8 @@ python:
 install:
   - pip install --upgrade .
   - pip install pylint
+  - pip install python-coveralls
 script:
   - ./run_presubmit.sh
+after_success:
+  - coveralls
```
bashir2: nit: I suppose this is the call that uploads the coverage results to coveralls.io, right? Maybe add a comment for this unless it is obvious for anyone knowing coveralls.

arostamianfar: Done. PTAL :)

PR #96, opened 2018-01-25
Adds a cloudbuild_CI.yaml for Build Triggers with related fixes to test scripts.

bashir2:
Please note that by default the container builder service account (by whom the cloudbuild steps will be run) only has Cloud Container Builder access. Of course that is not enough to run the integration tests. I tried to enable the accesses one by one (e.g., BigQuery, Genomics, GCP, etc.) but then decided it is easier to give this service account "Editor" access (like some other service accounts). That fixed all remaining permission issues.

Tested:
Created a Build Trigger in my project monitoring my forked git repo (using the new cloudbuild_CI.yaml) and tested both success and failure scenarios.

Issue: #62

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -24,6 +24,5 @@ steps:
       - '--tag=gcr.io/$PROJECT_ID/gcp-variant-transforms:${_CUSTOM_TAG_NAME}'
       - '--file=docker/Dockerfile'
       - '.'
-    dir: '.'
 images:
   - 'gcr.io/$PROJECT_ID/gcp-variant-transforms:${_CUSTOM_TAG_NAME}'
--- a/cloudbuild_CI.yaml
+++ b/cloudbuild_CI.yaml
@@ -1,0 +1,50 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This is meant to be used by the Build Triggers set in Container Builder Builds
+# of a cloud project. This is what it does:
+# * Builds an image using the last GitHub commit hash as the tag.
+# * Pushes that image to Container Registry.
+# * Runs the integration test from that very image.
+
+steps:
+  - name: 'gcr.io/cloud-builders/docker'
+    args:
+      - 'build'
+      - '--tag=gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
+      - '--file=docker/Dockerfile'
+      - '.'
+  # We have to push now since we are using this image in the next step.
+  - name: 'gcr.io/cloud-builders/docker'
+    args:
+      - 'push'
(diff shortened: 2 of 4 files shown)


cloudbuild.yaml
```diff
@@ -24,6 +24,5 @@ steps:
       - '--tag=gcr.io/$PROJECT_ID/gcp-variant-transforms:${_CUSTOM_TAG_NAME}'
       - '--file=docker/Dockerfile'
       - '.'
-    dir: '.'
```
nmousavi: Hmm, so it's not needed? I remember we had to add it but don't remember why? :)

cloudbuild_CI.yaml
```diff
@@ -0,0 +1,50 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
```
nmousavi: How and Where is this file used?

cloudbuild_CI.yaml
```diff
@@ -0,0 +1,50 @@
...
+# of a cloud project. What it does is:
+# * Build an image using the last GitHub commit hash as the tag.
+# * Push that image to Container Registry.
+# * Runs the integration test from that very image.
+
+steps:
+  - name: 'gcr.io/cloud-builders/docker'
+    args:
+      - 'build'
+      - '--tag=gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
```
nmousavi: Optional: I prefer 's/COMMIT_SHA/TAG_NAME', and then populating TAG_NAME with COMMIT_SHA value in the script calling gcloud build.  COMMIT_SHA doesn't look to belong to here.

deploy_and_run_integration_tests.sh
```diff
@@ -25,17 +25,21 @@ set -euo pipefail
...
+GS_DIR_OPT="--gs_dir"
 KEEP_IMAGE_OPT="--keep_image"
 IMAGE_TAG_OPT="--image_tag"
+PROJECT_OPT="--project"
+RUN_UNIT_TEST_OPT="--run_unit_tests"
 SKIP_BUILD_OPT="--skip_build"
 image_tag=""
 keep_image=""
 skip_build=""
+gs_dir="integration_test_runs"  # default GS dir to store logs, etc.
```
nmousavi: should update usage() too?

deploy_and_run_integration_tests.sh
```diff
@@ -168,20 +197,18 @@ export VIRTUAL_ENV_DISABLE_PROMPT="something"
 virtualenv "${temp_dir}"
 source ${temp_dir}/bin/activate;
 trap clean_up EXIT
+if [[ -n "${run_unit_tests}" ]]; then
```
nmousavi: a bit counterintuitive: the script is called deploy_and_run_integration test, but it runs unit test too.

cloudbuild.yaml
bashir2: Correct, AIUI, dir is the path relative to /workspace where everything runs. '.' is the default value.

cloudbuild_CI.yaml
bashir2: This is used when setting up a Build Trigger config for a project to monitor GitHub changes.
The comment below makes a reference to it.

cloudbuild_CI.yaml
bashir2: Like discussed offline, these are "default substitutions" in Container Builder:
https://cloud.google.com/container-builder/docs/configuring-builds/substitute-variable-values
So COMMIT_SHA is the commit hash value at which the image was created and tested.

deploy_and_run_integration_tests.sh
bashir2: Done.
Heh, I actually updated usage last night, after sending the PR, but forgot to push it to GitHub, I guess I was too tired :-)

deploy_and_run_integration_tests.sh
bashir2: I changed the script name to deploy_and_run_tests

PR #97, opened 2018-01-26
Updated development doc to reflect new integration test related changes, plus some minor tweaks.

### LINKED ISSUE #62: Fix and enable continuous run of the integration tests.

bashir2:
Final documentation to close Issue #62.

Fixes #62.

### DIFF
--- a/docs/development_guide.md
+++ b/docs/development_guide.md
@@ -62,7 +62,7 @@ git checkout -b <branch-name> origin/master
 ```
 
 ### Testing
-To run all tests:
+To run all unit tests:
 
 ```bash
 python setup.py test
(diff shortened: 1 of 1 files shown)


docs/development_guide.md
```diff
@@ -122,4 +140,29 @@ git commit --amend
...
+```bash
+git checkout master
+git checkout -b foo_review
+git merge --squash foo
+git push origin foo_review
+```
+This approach is specially useful if you tend to do a lot of small commits
+during your feature development and like to keep them as checkpoints.
+
+### Continuous integration
```
arostamianfar: Do we want to make the docker images in the test project also public? I'm ok with that. It may even be a good idea for experimental images.
Otherwise, this is kind of confusing for an external user (since they don't have access to any of the images) and we should rephrase.

bashir2: I don't mind if we make those images public but even if we do that, can people who don't have access to the test project check the status of the builds? IIUC, making the images public is not enough to look at the build history dashboard. Or did you mean give broader public access to the test project itself?

I see your point about the potential confusion for an external contributor. I made the reference to the test project here for an internal contributor. I added a note to clarify the access issue.

arostamianfar: Thanks for rewording. I think this is good for now.
You are right about the build process not being public anyway.

PR #100, opened 2018-01-30
Added a comment for the test script run step of cloudbuild_CI.yaml

bashir2:
This is to prevent confusion about how the test script is run.

### DIFF
--- a/cloudbuild_CI.yaml
+++ b/cloudbuild_CI.yaml
@@ -31,6 +31,13 @@ steps:
       - 'push'
       - 'gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
   # Run the test script from the image we just built and pushed.
+  #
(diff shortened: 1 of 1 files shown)


cloudbuild_CI.yaml
```diff
@@ -31,6 +31,13 @@ steps:
       - 'push'
       - 'gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
   # Run the test script from the image we just built and pushed.
+  #
+  # Note that the only reason that we use the built image in the 'name' argument
```
arostamianfar: Thanks, Bashir! This makes sense now!
Consider rephrasing to be more concise. Something like:
Note that './deploy_and_run_tests.sh' actually references the source downloaded from GitHub and not a file in the docker image we just built. The only reason that we use the built image in the 'name' argument is to avoid reinstalling python, virtualenv, etc on the image.

bashir2:
PTAL

PR #103, opened 2018-01-31
Use hash of entries for merge key.

### LINKED ISSUE #102: Use a hash for reference/alternate_bases when making the merge key
We currently just concatenate reference/alternate_bases together as the merge key (see [here](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy.py#L120)).

This may break Dataflow if the length of reference/alternate_bases are large (larger than 1MB limit for the key). We should create a hash of these fields and use that instead. This ensures that our keys do not get arbitrarily large.

As a result, the key would be `reference_name:start_position:end_position:<hash of reference_name:alternate_bases>`. Given that we include reference/start/end as part of the key (excluding the hash) then the risk of collision is extremely low. However, we can still add an assert [here](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy.py#L99) just to be sure.

nmousavi:
Sha256 of reference_bases and alternate_bases is used for the merge key.

fixes #102

### DIFF
--- a/gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy.py
+++ b/gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy.py
@@ -16,6 +16,7 @@
 
 from __future__ import absolute_import
 
+import hashlib
 import re
 
 from gcp_variant_transforms.beam_io.vcfio import Variant
@@ -103,6 +104,17 @@ class MoveToCallsStrategy(variant_merge_strategy.VariantMergeStrategy):
                                  end=variant.end,
                                  reference_bases=variant.reference_bases,
                                  alternate_bases=variant.alternate_bases)
+      # Since we use hash function in generating the merge key, there is
+      # a chance (extremely low though) to have variants with different
+      # `reference_bases` or `alternate_base` here due to a collision in
+      # the hash function.
+      assert variant.reference_bases == merged_variant.reference_bases, (
+          'Cannot merge variants with different reference bases. {} vs {}'
+          .format(variant.reference_bases, merged_variant.reference_bases))
+      assert variant.alternate_bases == merged_variant.alternate_bases, (
+          'Cannot merge variants with different alternate bases. {} vs {}'
+          .format(variant.alternate_bases, merged_variant.alternate_bases))
+
       merged_variant.names.extend(variant.names)
       merged_variant.filters.extend(variant.filters)
       merged_variant.quality = max(merged_variant.quality, variant.quality)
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy.py
```diff
@@ -118,13 +130,14 @@ def get_merged_variants(self, variants, unused_key=None):
...
-    yield ':'.join(
+    key = ':'.join(
         [str(x) for x in [
             variant.reference_name or '',
             variant.start or '',
             variant.end or '',
-            variant.reference_bases or '',
-            ','.join(variant.alternate_bases or [])]])
+            hashlib.sha256(variant.reference_bases or '').hexdigest(),
+            hashlib.sha256(','.join(variant.alternate_bases or [])).hexdigest()]])
```
arostamianfar: Consider making a private method _get_hash_key and use it here.

gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy.py
```diff
@@ -118,13 +130,14 @@ def get_merged_variants(self, variants, unused_key=None):
...
   def get_merge_keys(self, variant):
-    yield ':'.join(
+    key = ':'.join(
         [str(x) for x in [
             variant.reference_name or '',
             variant.start or '',
             variant.end or '',
-            variant.reference_bases or '',
-            ','.join(variant.alternate_bases or [])]])
+            hashlib.sha256(variant.reference_bases or '').hexdigest(),
```
arostamianfar: I wonder if md5 is good enough here. 256 seems like an overkill :p

gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy_test.py
```diff
@@ -225,20 +226,28 @@ def test_get_merged_variants_move_everything_to_calls(self):
   def test_get_merge_keys(self):
     strategy = move_to_calls_strategy.MoveToCallsStrategy(None, None, None)
 
+    empty_string_hash = hashlib.sha256('').hexdigest()
```
arostamianfar: Consider making a helper method here as well. I think it's ok to use the one in the private lib too.

gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy.py
nmousavi: Used  md5 instead.

gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy.py
nmousavi: Done!

gcp_variant_transforms/libs/variant_merge/move_to_calls_strategy_test.py
nmousavi: Done!

nmousavi:
Job finished successfully on platinum dataset. 

Thanks for the review!

PR #105, opened 2018-02-03
Add new --omit_empty_sample_calls option.

snarfed:
...to skip empty calls, e.g. samples that don't have a given variant. Useful for multi-sample VCFs, especially when most variants are sparse and are only called in a small fraction of samples.

This switches from translating VCFs literally to using BigTable's support for repeated columns and sparse storage more effectively.

I'm using this to load our CNVs, which generally have hundreds of samples but usually just one sample per call. This makes the resulting rows in BigTable easier to work with, since each `call` row is an actual call.

I'm flexible on pretty much all the details and code here. Open to other ideas. Thanks in advance!

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_vcf_schema.py
+++ b/gcp_variant_transforms/libs/bigquery_vcf_schema.py
@@ -216,10 +216,11 @@ def generate_schema_from_header_fields(header_fields, variant_merger=None,
 
 
 # TODO: refactor this to use a class instead.
-def get_rows_from_variant(variant, split_alternate_allele_info_fields=True):
+def get_rows_from_variant(variant, split_alternate_allele_info_fields=True,
+                          omit_empty_sample_calls=False):
   """Yields BigQuery rows according to the schema from the given variant.
 
-  There is a 10MB limit for each BigQuqery row, which can exceed by having
+  There is a 10MB limit for each BigQuery row, which can be exceeded by having
   a large number of calls. This method may split up a row into multiple rows if
   it exceeds 10MB.
 
@@ -229,6 +230,8 @@ def get_rows_from_variant(variant, split_alternate_allele_info_fields=True):
       `Number=A` (i.e. one value for each alternate allele) will be stored under
       the `alternate_bases` record. If false, they will be stored with the rest
       of the INFO fields.
+    omit_empty_sample_calls (bool): If true, samples that don't have a given
+      call will be omitted.
   Yields:
     A dict representing a BigQuery row from the given variant. The row may have
     a subset of the calls if it exceeds the maximum allowed BigQuery row size.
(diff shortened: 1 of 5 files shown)


gcp_variant_transforms/libs/bigquery_vcf_schema.py
```diff
@@ -244,6 +247,20 @@ def get_rows_from_variant(variant, split_alternate_allele_info_fields=True):
   row = copy.deepcopy(base_row)  # Keep base_row intact.
   for call in variant.calls:
     call_record = _get_call_record(call)
+
+    if omit_empty_sample_calls:
+      # Omit empty calls, e.g. samples that don't have this variant.
+      for key, val in call_record.items():
+        if key == 'genotype':
+          if val and set(val) != set((vcfio.MISSING_GENOTYPE_VALUE,)):
```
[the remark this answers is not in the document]
arostamianfar: Oops! I just noticed you *were* using the correct value and I misread it! Please ignore my comment.

snarfed:
thanks for the review! these suggestions all sound good. I'll ping once I've incorporated them.

snarfed:
done. definitely cleaner this way, thanks for the suggestions.

i kept the boolean coercion in `_is_empty_field()` and just check for 0 explicitly. apart from genotype, input values all come from `_get_bigquery_sanitized_field`, so i'm not too worried about unusual types accidentally boolean evaluating to False. otherwise, checking for an explicit set of known empty values might be the next best thing, e.g. `None, {}, [], (), '', set(), frozenset()`.

i'm also happy to squash the two commits if you want, just let me know.

gcp_variant_transforms/libs/bigquery_vcf_schema.py
```diff
@@ -258,17 +264,26 @@ def get_rows_from_variant(variant, split_alternate_allele_info_fields=True):
 
 
 def _get_call_record(call):
-  """A helper method for ``get_rows_from_variant`` to get a call as JSON."""
+  """A helper method for ``get_rows_from_variant`` to get a call as JSON.
+
+  Args:
+    call: PyVCF Variant call to convert.
```
arostamianfar: nit: this is actually  ` ``VariantCall`` ` inside vcfio.
Please use:
call (` ``VariantCall`` `): Variant call to convert.

gcp_variant_transforms/libs/bigquery_vcf_schema.py
```diff
@@ -258,17 +264,26 @@ def get_rows_from_variant(variant, split_alternate_allele_info_fields=True):
...
+
+  Returns:
+    BigQuery call value (dict).
+  """
   call_record = {
       ColumnKeyConstants.CALLS_NAME: _get_bigquery_sanitized_field(call.name),
       ColumnKeyConstants.CALLS_PHASESET: call.phaseset,
       ColumnKeyConstants.CALLS_GENOTYPE: call.genotype or []
   }
+  is_empty = _is_empty_field('genotype', call.genotype)
```
arostamianfar: What do you think of treating genotype as a special field and just using is_empty_field for other info fields? This avoids special-casing genotype inside is_empty_field
In other words, something like: `is_empty = set(call.genotype) == set((vcfio.MISSING_GENOTYPE_VALUE,))`

snarfed: sure! done. i actually considered this originally, but didn't do it because the boolean value check gets duplicated. i'm fine either way though!

snarfed:
thanks! all done, and squashed.

btw, we also use coveralls, and had the same concern with it as a check on github, since it's overly sensitive and marks any drop as a failure. we still use it, but we disconnected it from github and track it more offline, over the longer term.

snarfed:
woo, thanks for merging!

arostamianfar:
I'll release a new docker image with this change once our integration tests pass :)

arostamianfar:
FYI, the docker image is updated as well!

Issue #81, opened 2018-01-23
Add native annotation support


arostamianfar:
This task is to add native annotation support (e.g. gene ID, variant harmfulness, clinvar, dbSNP, etc) as part of the VCF->BigQuery pipeline.

Initially, we can have a pre-processing step that adds [VEP](https://useast.ensembl.org/info/docs/tools/vep/index.html) or [SnpEff](http://snpeff.sourceforge.net/SnpEff_manual.html) annotations to the VCF file prior to import. The annotations would be added to a new INFO field that can then be parsed to load each annotation as a separate column to BigQuery. The annotation field presumably follows [this schema](http://snpeff.sourceforge.net/VCFannotationformat_v1.0.pdf).

Having a more generic INFO parser will also be beneficial for other pre-annotated VCF files such as [gnomAD](http://gnomad.broadinstitute.org/).

We can integrate [Verily's annotation pipeline](https://github.com/verilylifesciences/variant-annotation) to load annotations via VEP.

Note: adding annotations as part of the pipeline without pre-processing is outside the scope of this task. We can consider that option if the pre-processing step is not feasible for certain use cases.

bashir2:
We can now close this super-issue as the original goal is achieved: We can (i) properly parse annotation fields and (ii) run VEP as a preprocess step of VT; this is fully integrated with VT and only needs supplying some flags.

Other remaining issues like documentation/integration-test/refactoring have their own tracking issues. The ones targeted for Q2 are under [this project](https://github.com/googlegenomics/gcp-variant-transforms/projects/3).

PR #107, opened 2018-02-07
Adds an option to split annotation fields.

bashir2:
Please note that some more unit-tests need to be added and there are a few TODOs (some for follow-up PRs), but this is in a reasonable shape for the first pass review.

Tested:
Added some unit-tests and one integration test (valid_4_2_VEP.json).
More interesting query: https://bigquery.cloud.google.com/results/gcp-variant-transforms-test:bquijob_7d798216_1616efded18

Issue #81

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_vcf_schema.py
+++ b/gcp_variant_transforms/libs/bigquery_vcf_schema.py
@@ -26,6 +26,7 @@ import vcf
 
 from apache_beam.io.gcp.internal.clients import bigquery
 from gcp_variant_transforms.beam_io import vcfio
+from gcp_variant_transforms.libs import vcf_header_parser
 
 
 __all__ = ['generate_schema_from_header_fields', 'get_rows_from_variant',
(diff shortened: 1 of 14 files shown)


gcp_variant_transforms/libs/bigquery_vcf_schema.py
```diff
@@ -82,8 +83,15 @@ class _TableFieldConstants(object):
 _JSON_CONCATENATION_OVERHEAD_BYTES = 5
 
 
+# TODO(bashir2): Using type identifiers like ``HeaderFields`` does not seem
```
arostamianfar: yeah, the type identifiers have been on my mind for some time. I inherited it from Beam conventions, but have not actually set up automated documentation generation to verify that it finds the correct class.
I think the 'correct' way is something like `` :class:`HeaderFields` ``. In any case, I filed Issue #108 for this. You can remove the TODO from here as it's awkward to single out this particular field.

bashir2:
To document follow-up discussion on the design decisions:

* The reason that we need to touch vcfio is that it is the only place that we have access to _original_ headers. Original headers are needed to resolve inconsistency between annotation headers, e.g., Conseq|Gene in one file and Gene|Conseq in another (see [this TODO](https://github.com/googlegenomics/gcp-variant-transforms/pull/107/files#diff-eddcc9d3e28546df1dfd0f63c0a47053R815) for a code pointer).

* We decided that we will _not_ support resolving these kind of conflicts. With that, we can rely on the _merged_ header and do not attach annotation format to annotation fields of each variant.

* This also applies to the `field_count` which is already in [`VariantInfo`](https://github.com/googlegenomics/gcp-variant-transforms/pull/107/files#diff-eddcc9d3e28546df1dfd0f63c0a47053R54).

* To make the annotation extraction support more decoupled from the rest of the code, we need more refactoring to use the merged header as a side input to a PTransform that processes Variant. With this we will make VCF reading (vcfio) and BigQuery writing (bigquery_vcf_schema) dumb transforms that do mostly reading/parsing and writing, not much more processing.

* Supporting multiple annotation fields is simple, both in the current form and the future refactored form. We just need  to change --annotation_field to be a comma separated list of field names.

bashir2:
This was superseded by PR #109 .

Issue #59, opened 2018-01-12
Refactor bigquery_vcf_schema.py to use a class


arostamianfar:
We currently use basic functions for getting the schema and converting variants to BQ rows. This has made the code less readable especially when making subtle changes (e.g. Issue #48 ). We should first refactor the file to use a class. This also makes checking the schema while the rows are being generated much easier as the schema would be a member variable of that class.

bashir2:
As discussed offline, I can do this while I am working on Issue #81.

bashir2:
With PR #118 merged, I do not have any immediate plans for making more refactoring to this module. Assigning this to @nmousavi as he may make such changes as he is working on Issue #101 .

PR #109, opened 2018-02-09
Adds annotation field support and some related refactoring.

bashir2:
Follow up (i.e., replacement) of PR #107.

Please note that this PR needs more work to fix current unit-test and add new ones (plus other minor clean-ups and documentations) but I am sharing at this point to get early feedback on the design and make sure we are on the same page. In particular, please look at processed_variant module and how Variant objects are transformed into ProcessedVariant.

Tested:
Manually ran the pipeline locally on the new integration test file (see valid_4_2_VEP.json) and checked the output.
Also `./deploy_and_run_tests.sh --keep_image --keep_tables` passed.
Sample query:
`SELECT start_position, end_position, alt.AF, annotations.Consequence
FROM integration_tests_20180210_010224.valid_4_2_VEP AS t, t.alternate_bases as alt LEFT JOIN alt.CSQ AS annotations
ORDER BY start_position, alt.AF, annotations.Consequence`

Issue #81, #59

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_vcf_schema.py
+++ b/gcp_variant_transforms/libs/bigquery_vcf_schema.py
@@ -21,11 +21,13 @@ import json
 import math
 import re
 import sys
+from typing import Dict, Any  #pylint: disable=unused-import
 
 import vcf
 
 from apache_beam.io.gcp.internal.clients import bigquery
 from gcp_variant_transforms.beam_io import vcfio
+from gcp_variant_transforms.libs import processed_variant
 
 
 __all__ = ['generate_schema_from_header_fields', 'get_rows_from_variant',
@@ -83,7 +85,8 @@ _JSON_CONCATENATION_OVERHEAD_BYTES = 5
 
 
 def generate_schema_from_header_fields(header_fields, variant_merger=None,
-                                       split_alternate_allele_info_fields=True):
+                                       split_alternate_allele_info_fields=True,
+                                       annotation_fields=None):
   """Returns a ``TableSchema`` for the BigQuery table storing variants.
 
   Args:
@@ -97,6 +100,8 @@ def generate_schema_from_header_fields(header_fields, variant_merger=None,
       `Number=A` (i.e. one value for each alternate allele) will be stored under
       the `alternate_bases` record. If false, they will be stored with the rest
       of the INFO fields.
+    annotation_fields (List[str]): If provided, it is the list of annotation
+      INFO fields.
   """
   schema = bigquery.TableSchema()
   schema.fields.append(bigquery.TableFieldSchema(
@@ -141,6 +146,27 @@ def generate_schema_from_header_fields(header_fields, variant_merger=None,
             type=_get_bigquery_type_from_vcf_type(field.type),
             mode=_TableFieldConstants.MODE_NULLABLE,
             description=_get_bigquery_sanitized_field(field.desc)))
+
+  for annot_field in annotation_fields or []:
+    if not annot_field in header_fields.infos:
+      raise ValueError('Annotation field {} not found'.format(annot_field))
+    annotation_names = processed_variant.extract_annotation_names(
+        header_fields.infos[annot_field].desc)
+    annotation_record = bigquery.TableFieldSchema(
+        name=_get_bigquery_sanitized_field(annot_field),
+        type=_TableFieldConstants.TYPE_RECORD,
+        mode=_TableFieldConstants.MODE_REPEATED,
+        description='List of {} annotations for this alternate.'.format(
+            annot_field))
+    for annotation_name in annotation_names:
+      annotation_record.fields.append(bigquery.TableFieldSchema(
+          name=_get_bigquery_sanitized_field(annotation_name),
+          type=_TableFieldConstants.TYPE_STRING,
+          mode=_TableFieldConstants.MODE_NULLABLE,
+          # TODO(bashir2): Add descriptions of well known annotations, e.g.,
+          # from VEP.
+          description=''))
+    alternate_bases_record.fields.append(annotation_record)
   schema.fields.append(alternate_bases_record)
 
   schema.fields.append(bigquery.TableFieldSchema(
@@ -198,11 +224,13 @@ def generate_schema_from_header_fields(header_fields, variant_merger=None,
 
   # Add info fields.
   info_keys = set()
+  annotation_fields_set = set(annotation_fields or [])
   for key, field in header_fields.infos.iteritems():
     # END info is already included by modifying the end_position.
     if (key == vcfio.END_INFO_KEY or
         (split_alternate_allele_info_fields and
-         field.num == vcf.parser.field_counts[_FIELD_COUNT_ALTERNATE_ALLELE])):
+         field.num == vcf.parser.field_counts[_FIELD_COUNT_ALTERNATE_ALLELE]) or
+        key in annotation_fields_set):
       continue
     schema.fields.append(bigquery.TableFieldSchema(
         name=_get_bigquery_sanitized_field_name(key),
@@ -216,8 +244,8 @@ def generate_schema_from_header_fields(header_fields, variant_merger=None,
 
 
 # TODO: refactor this to use a class instead.
-def get_rows_from_variant(variant, split_alternate_allele_info_fields=True,
-                          omit_empty_sample_calls=False):
+def get_rows_from_variant(variant, omit_empty_sample_calls=False):
+  # type: (processed_variant.ProcessedVariant, bool) -> Dict
   """Yields BigQuery rows according to the schema from the given variant.
 
   There is a 10MB limit for each BigQuery row, which can be exceeded by having
@@ -225,11 +253,7 @@ def get_rows_from_variant(variant, split_alternate_allele_info_fields=True,
   it exceeds 10MB.
 
   Args:
-    variant (``Variant``): Variant to process.
-    split_alternate_allele_info_fields (bool): If true, all INFO fields with
-      `Number=A` (i.e. one value for each alternate allele) will be stored under
-      the `alternate_bases` record. If false, they will be stored with the rest
-      of the INFO fields.
+    variant (``ProcessedVariant``): Variant to convert to a row.
     omit_empty_sample_calls (bool): If true, samples that don't have a given
       call will be omitted.
   Yields:
@@ -240,8 +264,7 @@ def get_rows_from_variant(variant, split_alternate_allele_info_fields=True,
   """
   # TODO: Add error checking here for cases where the schema defined
   # by the headers does not match actual records.
-  base_row = _get_base_row_from_variant(
-      variant, split_alternate_allele_info_fields)
+  base_row = _get_base_row_from_variant(variant)
   base_row_size_in_bytes = _get_json_object_size(base_row)
   row_size_in_bytes = base_row_size_in_bytes
   row = copy.deepcopy(base_row)  # Keep base_row intact.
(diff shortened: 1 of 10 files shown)


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -217,14 +218,20 @@ def run(argv=None):
     if variant_merger:
       variants |= (
           'MergeVariants' >> merge_variants.MergeVariants(variant_merger))
-    _ = (variants |
+    proc_variants = variants | 'ProcessVaraints' >> beam.Map(
+        processed_variant.create_processed_variant, header_fields,
```
bashir2: @arostamianfar when running the pipeline with DataflowRunner, I get a serialization related error at this stage which says:
PicklingError: Can't pickle <type 'function'>: attribute lookup __builtin__.function failed

I am still trying to figure out what causes this and have some theories, but have not got to the root cause/solution yet. If you have had similar issues before, I appreciate your comments.

bashir2: I spent some time on this and I have found _a_ solution. The problem seems to be that Beam uses cPickle as the default serializer and that cannot serialize the new class ProcessedVariant (I think because it has a list of another class SingleAlternateData).

One fix I have found is to specify the coder in coder_registry, i.e., this new line that I just updated:
`beam.coders.registry.register_coder(processed_variant.ProcessedVariant, beam.coders.coders.DillCoder)`

That said, dill is slower than cPickle so this has some performance implications (see the TODO before this line).

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,244 @@
...
+    self._infos[field_name] = data
+
+  def _add_annotations(self, field_name, annotations, annotation_names):
+    annotation_dict = {}
+    for i in range(len(annotations)):
+      annotation_dict[annotation_names[i]] = annotations[i]
+    self._annotations[field_name].append(annotation_dict)
+
+
+def create_processed_variant(variant, header_fields,
```
arostamianfar: Consider writing this as a PTransform where input is a Variant and output is a ProcessedVariant. The ProcessedVariant class itself should just be a data storage class without any data manipulation logic (similar to the Variant class; think of it as a proto) and all such logic should be inside that PTransform or a helper library class ([recommended by Beam](https://beam.apache.org/contribute/ptransform-style-guide/#exposing-a-ptransform-vs-something-else)) I think it would be even cleaner to add a 'modify_schema' method to that library so that we can remove annotations and "split alt info logic" from bigquery_vcf_schema altogether (this would be similar to variant_merger). That way, if we decide to add a new annotation field in the future, all of the logic would be in one place.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,244 @@
...
+            annotations[0], field_name, annotation_str))
+
+
+class SingleAlternateData(object):
+
+  def __init__(self, alt_bases):
+    self._alt_bases = alt_bases
+    self._infos = {}
+    # Each list item in this '_annotations' dictionary is dictionary itself.
+    self._annotations = defaultdict(lambda: [])
```
arostamianfar: I don't think we should special-case annotations. They are the same as 'infos' as far as this class is concerned.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,244 @@
...
+  def get_quality(self):
+    return self._variant.quality
+
+  def get_filters(self):
+    return self._variant.filters
+
+  def get_calls(self):
+    return self._variant.calls
+
+  def get_non_alt_infos(self):
```
arostamianfar: nit: just `info`, it's cleaner :) We can assume that this returns the top-level info and not per-alt or per-call info.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,244 @@
...
+      # TODO(bashir2): Currently we only check exact matches of alternate bases
+      # which is not enough. We should implement the whole standard for finding
+      # alternate bases for an annotation list.
+      if not found_alt:
+        logging.warning('Could not find matching alternate bases for {} in '
+                        'annotation filed {} with content {}'. format(
+            annotations[0], field_name, annotation_str))
+
+
+class SingleAlternateData(object):
```
arostamianfar: Consider renaming this class to be more descriptive. Perhaps VariantAlternateBase? (similar to VariantCall). I think this should also just be a simple storage class (similar to a proto).

bashir2:
Thanks Asha for the comments and the offline discussion. I think I have addressed most of the design related comments and this is ready for another round of review. PTAL

(Needless to say that there are TODOs about documentation and unit-tests that needs to be added which will all be done once we settle on the design; for now I just reran with direct and dataflow runners and checked the output tables manually).

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,283 @@
...
+    # `vcfio.VariantInfo.data`).
+    self._infos = {}
+
+  @property
+  def alternate_bases(self):
+    # type: () -> str
+    return self._alt_bases
+
+  @property
+  def infos(self):
```
arostamianfar: I actually used 'info' in both Variant and VariantCall to be consistent with the proto definitions. I realize that this is more about proto semantics (which is why I was ok with using infos in VcfHeader), but you may consider being consistent with that notation just so that we dont have 'info' and 'infos' in the same Variant context.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,283 @@
...
+    for i in range(len(annotations)):
+      annotation_dict[annotation_names[i]] = annotations[i]
+    alt._infos[field_name].append(annotation_dict)
+
+
+def _extract_annotation_list_with_alt(annotation_str):
+  # type: (str) -> List[str]
+  """Extracts annotations from an annotation INFO field.
+
+  This works by dividing the ``annotation_str`` on '|'. The first element is
```
arostamianfar: Please provide a small example of what this string looks like and what the return value is like.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+# TODO(bashir2): Once the refactoring of bigquery_vcf_schema is complete, these
+# functions should not be used outside this module (mark them as 'private').
+
+def extract_annotation_names(description):
+  # type: (str) -> List[str]
+  """Extracts annotation list from the description of an annotation INFO field.
+
+  This is similar to extract_extract_annotation_list_with_alt with the
```
arostamianfar: Please provide a small example of what this string looks like and what the return value is like.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,283 @@
...
+      field_name (str): The name of the annotation field.
+      data(List[str]): The data part of the field separated on comma.
+    """
+    if not self._header_fields.infos.has_key(field_name):
+      raise ValueError('{} INFO not found in the header, variant: {}'.format(
+          field_name, proc_var))
+    header_desc = self._header_fields.infos[field_name].desc
+    annotation_names = extract_annotation_names(header_desc)
+    for annotation_str in data:
+      annotations = _extract_annotation_list_with_alt(annotation_str)
```
arostamianfar: Given that the first element is always the alternate base, I think it's more readable and a bit more efficient to switch the orders: first convert `data` to be a map of alternate_base -> list of split annotations [1]. Then loop through each alternate_base and add the necessary data. This also avoids having the 1-offset in the code (e.g. 'len - 1' below).

[1] I think it would be even cleaner to do the formatting here as well, so the returned data already contains the annotation names together with the data.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,283 @@
...
+      for alt in proc_var._alternate_datas:
+        if alt.alternate_bases == annotations[0]:
+          found_alt = True
+          self._add_annotation_for_alt(
+              alt, field_name, annotations[1:], annotation_names)
+          break
+      # TODO(bashir2): Currently we only check exact matches of alternate bases
+      # which is not enough. We should implement the whole standard for finding
+      # alternate bases for an annotation list.
+      if not found_alt:
```
arostamianfar: nit: you can use python's "for else" statement for these cases, but this may not be needed anyway if we change how we process the annotations (see above).

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,283 @@
...
+      proc_var (ProcessedVariant): The object to which the annotations are being
+        added.
+      field_name (str): The name of the annotation field.
+      data(List[str]): The data part of the field separated on comma.
+    """
+    if not self._header_fields.infos.has_key(field_name):
+      raise ValueError('{} INFO not found in the header, variant: {}'.format(
+          field_name, proc_var))
+    header_desc = self._header_fields.infos[field_name].desc
+    annotation_names = extract_annotation_names(header_desc)
```
arostamianfar: I think it's cleaner if this is also done in init, so that you have a map of annotation_field -> annotation_names and just use it here.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,244 @@
...
+    self._variant = variant
+    self._non_alt_infos = None
+    self._single_alternate_datas = []
+    for a in variant.alternate_bases:
+      self._single_alternate_datas.append(SingleAlternateData(a))
+
+  def __repr__(self):
+    return self._variant.__repr__()
+
+  def get_reference_name(self):
```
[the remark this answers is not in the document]
arostamianfar: I think @property is the more pythonic way of doing things like this and the API matches the regular Variant class.

gcp_variant_transforms/libs/processed_variant.py
arostamianfar: I agree, but this is unfortunately a fundamental issue with VCF itself :) We can revisit once we move to protos.

gcp_variant_transforms/libs/bigquery_vcf_schema.py
```diff
@@ -216,32 +244,29 @@ def generate_schema_from_header_fields(header_fields, variant_merger=None,
...
+    variant (``ProcessedVariant``): Variant to convert to a row.
     omit_empty_sample_calls (bool): If true, samples that don't have a given
       call will be omitted.
   Yields:
     A dict representing a BigQuery row from the given variant. The row may have
     a subset of the calls if it exceeds the maximum allowed BigQuery row size.
   Raises:
     ValueError: If variant data is inconsistent or invalid.
+
+  :type variant: :class:`processed_variant.ProcessedVariant`
```
arostamianfar: nit: is this still needed now that we have the type: comment?

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -0,0 +1,307 @@
...
+  split and attached to alternates. This is not inherited from
+  :class:``vcfio.Variant`` as an encapsulation layer and to prefer composition.
+  """
+
+  def __init__(self, variant):
+    # type: (vcfio.Variant) -> None
+    if not isinstance(variant, vcfio.Variant):
+      raise ValueError('Expected an instance of vcfio.Variant.')
+    self._variant = variant
+    self._non_alt_info = {} # type: Dict[str, Any]
```
arostamianfar: nit: can these type comments also have two spaces before #?

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -105,6 +105,15 @@ def add_arguments(self, parser):
         '--omit_empty_sample_calls',
         type='bool', default=False, nargs='?', const=True,
         help=("If true, samples that don't have a given call will be omitted."))
+    parser.add_argument(
+        '--annotation_fields',
+        default=None, nargs='+',
```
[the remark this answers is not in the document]
arostamianfar: Good point! I think this may be why we had None in --reference_names too :p

Issue #112, opened 2018-02-15
"variant merging" documentation


smrgit:
It would be great to have some more information about variant merging -- I have sometimes as many as 12 VCF files for a single tumor/normal pair and I'd like to understand how variants are merged.  I also noted that the main Google documentation page instructions suggests using ``--variant_merge_strategy MOVE_TO_CALLS`` while the documentation here on github doesn't mention that.  When I tried to import a set of 23 VCF files representing two tumor/normal pairs, I got the following error:
```
    ERROR:root:Exception at bundle <apache_beam.runners.direct.bundle_factory._Bundle object at 0x7f64a98d5450>, due to an exception.
     Traceback (most recent call last):
      File "/opt/gcp_variant_transforms/venv/local/lib/python2.7/site-packages/apache_beam/runners/direct/executor.py", line 312, in call
        side_input_values)
      File "/opt/gcp_variant_transforms/venv/local/lib/python2.7/site-packages/apache_beam/runners/direct/executor.py", line 347, in attempt_call
        evaluator.process_element(value)
      File "/opt/gcp_variant_transforms/venv/local/lib/python2.7/site-packages/apache_beam/runners/direct/transform_evaluator.py", line 551, in process_element
        self.runner.process(element)
      File "/opt/gcp_variant_transforms/venv/local/lib/python2.7/site-packages/apache_beam/runners/common.py", line 390, in process
        self._reraise_augmented(exn)
      File "/opt/gcp_variant_transforms/venv/local/lib/python2.7/site-packages/apache_beam/runners/common.py", line 388, in process
        self.do_fn_invoker.invoke_process(windowed_value)
      File "/opt/gcp_variant_transforms/venv/local/lib/python2.7/site-packages/apache_beam/runners/common.py", line 281, in invoke_process
        self._invoke_per_window(windowed_value)
      File "/opt/gcp_variant_transforms/venv/local/lib/python2.7/site-packages/apache_beam/runners/common.py", line 307, in _invoke_per_window
        windowed_value, self.process_method(*args_for_process))
      File "/opt/gcp_variant_transforms/venv/local/lib/python2.7/site-packages/apache_beam/typehints/typecheck.py", line 63, in process
        return self.wrapper(self.dofn.process, args, kwargs)
      File "/opt/gcp_variant_transforms/venv/local/lib/python2.7/site-packages/apache_beam/typehints/typecheck.py", line 81, in wrapper
        result = method(*args, **kwargs)
[...]

Issue #76, opened 2018-01-23
Add the remaining documentation pages


arostamianfar:
We still need to add documentation for the merging logic and handling invalid data.

arostamianfar:
We also need a higher level overview of what the tool does for people who are not even familiar with BigQuery.

arostamianfar:
Our documentation is now in a much better shape and we have docs about the main features (#116, #153 , #160 ). Closing this issue. Keeping the docs up to date with new features is an ongoing effort though and we should create new issues for them.

PR #116, opened 2018-02-20
Add Variant merging documentation.

arostamianfar:
Fixes Issue #112.
Also, adds documentation for Issue #76.

Preview is available at: https://github.com/arostamianfar/gcp-variant-transforms/blob/docs/docs/variant_merging.md

### DIFF
--- a/docs/variant_merging.md
+++ b/docs/variant_merging.md
@@ -1,3 +1,237 @@
 # Variant merging
 
-TODO
+You can use Variant Transforms to load multiple files at once
+by providing a pattern like `--input_pattern gs://mybucket/*.vcf`. This
+will create a merged BigQuery schema that contains all variants from all
+matching files. Note that the VCF fields must be compatible across files (e.g.
+`INFO` and `FORMAT` fields with the same key have compatible types and numbers
+across files).
+
+However, by default, calls between files will not be merged. For instance,
+consider the following VCF files:
+
+*`gs://mybucket/a.vcf`*:
+
+```
+## <headers omitted>
+#CHROM  POS   ID    REF   ALT   QUAL  FILTER  INFO    FORMAT  S1      S2
+1       100   rs1   A     T     29    PASS    DP=14   GT:GQ   0|1:48  0/0:30
+```
+
+*`gs://mybucket/b.vcf`*:
+
+```
+## <headers omitted>
+#CHROM  POS   ID    REF   ALT   QUAL  FILTER  INFO         FORMAT  S3
+1       100   rs1   A     T     7     q10     DP=11;AA=G   GT:GQ   1/1:2
+```
+
+If you load both files together (i.e. by specifying
+`--input_pattern gs://mybucket/*.vcf`), you would end up with 2 rows in
+BigQuery. One with calls `S1` and `S2`, and another with just call `S3` even
+though all calls have the same variant.
+
+Variant merging can be used to merge calls across files. There are different
+merging strategies and you can specify any of the following
+strategies by adding `--variant_merge_strategy <merge_strategy>` to the
+arguments.
+
+## MOVE_TO_CALLS strategy
+
+This strategy merges calls that have the same `reference_name`,
+`start_position`, `end_position`, `reference_bases`, and all `alternate_bases`
+across files. By default, the variant fields are merged as follows:
+
+* `Calls` are merged together in no particular order.
+* The highest `quality` value is chosen among all variants being merged.
+* All `filter` values are merged and deduplicated.
+* All `names` (`ID` column in the VCF spec) are merged and deduplicated.
+* For each `INFO` field, one is chosen as representative for the *entire* merged
+  variant record in *no particular order*. Note that for repeated fields, a
+  single set is chosen as representative for that particular field.
+
+Depending on your use case, you may customize the merge strategy as follows:
+
+* Adding `--copy_quality_to_calls` and/or `--copy_filter_to_calls` to the
+  arguments will copy the `quality` and/or `filter` values in each file to the
+  set of calls specified in that file. This option is useful when merging
+  single-call VCF files as `quality` and `filter` values usually correspond
+  to call-level quality/filter. Note that the variant-level `quality` and
+  `filter` values will still be merged according to the above logic.
+* Adding `--info_keys_to_move_to_calls_regex <regex>` will move `INFO` fields
+  that match the provided regular expression to the associated calls. This is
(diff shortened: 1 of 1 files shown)


docs/variant_merging.md
```diff
@@ -1,3 +1,227 @@
...
+* For each `INFO` field, one is chosen as representative for the *entire* merged
+  variant record in *no particular order*.
+
+Depending on your use case, you may customize the merge strategy as follows:
+
+* Adding `--copy_quality_to_calls` and/or `--copy_filter_to_calls` to the
+  arguments will copy the `quality` and/or `filter` values in each file to the
+  set of calls specified in that file. This option is useful when merging
+  single-call VCF files as `quality` and `filter` values usually correspond
+  the call-level quality/filter. Note that the variant-level `quality` and
```
bashir2: nit: \*to\* the call level?

docs/variant_merging.md
```diff
@@ -1,3 +1,227 @@
...
+  `filter` values will still be merged according to the above logic.
+* Adding `--info_keys_to_move_to_calls_regex <regex>` will move `INFO` fields
```
bashir2: Please add a hint re. what the spec of this regex is, e.g., by providing a link

docs/variant_merging.md
```diff
@@ -1,3 +1,227 @@
...
+This strategy merges calls that have the same `reference_name`,
+`start_position`, `end_position`, `reference_bases`, and all `alternate_bases`
+across files. By default, the variant fields are merged as follows:
+
+* `Calls` are merged together in no particular order.
+* The highest `quality` value is chosen among all variants being merged.
+* All `filter` values are merged and deduplicated.
+* All `names` (`ID` column in the VCF spec) are merged and deduplicated.
+* For each `INFO` field, one is chosen as representative for the *entire* merged
+  variant record in *no particular order*.
```
bashir2: Maybe it worth clarifying this for Number='.' cases even more.

docs/variant_merging.md
```diff
@@ -1,3 +1,227 @@
...
+
+* Adding `--copy_quality_to_calls` and/or `--copy_filter_to_calls` to the
+  arguments will copy the `quality` and/or `filter` values in each file to the
+  set of calls specified in that file. This option is useful when merging
+  single-call VCF files as `quality` and `filter` values usually correspond
+  the call-level quality/filter. Note that the variant-level `quality` and
+  `filter` values will still be merged according to the above logic.
+* Adding `--info_keys_to_move_to_calls_regex <regex>` will move `INFO` fields
+  that match the provided regular expression to the associated calls. This is
+  useful to ensure all `INFO` fields are kept when merging variants.
```
bashir2: What happens in FORMAT of two calls are different? Should you document that too?

docs/variant_merging.md
```diff
@@ -1,3 +1,227 @@
...
+  [
+    name: "S1",
+    genotype: [0, 1]
+    phaseset: "*"
+    GQ: 48
+  ],
+  [
+    name: "S2",
+    genotype: [0, 0]
+    phaseset: null
```
bashir2: Do we care pointing out why phaseset is null in S2 but is missing in S3?

docs/variant_merging.md
```diff
@@ -1,3 +1,227 @@
...
+1       50    .     T     <NON_REF>  .     .       END=150  GT:GQ   0/0:40
+```
+
+If you specify `--input_pattern gs://mybucket/*.vcf` with `MOVE_TO_CALLS` merge
+strategy, the non-variant region from call `S4` will not be merged with calls
+`S1`, `S2`, and `S3`. However, we know that call `S4` has genotype `0/0` in
+chr1:100 since it matches the reference from position 50-150 in chr1.
+The `MERGE_WITH_NON_VARIANTS` strategy will infer `0/0` for call `S4` at
+chr1:100 so that the resulting BigQuery table has a single row with all calls
+`S1-4`.
```
bashir2: In this case, do we want to break the [50, 150] segment above into [50, 99] and [101, 150]? Or do we keep the original [50, 150] region (hence S4 will have two entries for 1:100 which are consistent)?

docs/variant_merging.md
```diff
@@ -1,3 +1,227 @@
 # Variant merging
 
-TODO
+You can use Variant Transforms to load multiple files at once
+by providing a pattern like `--input_pattern gs://mybucket/**.vcf`. This
```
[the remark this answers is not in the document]
arostamianfar: I was trying to illustrate recursive search, but you're right, putting it here is confusing :) I'll write another doc on how to load multiple files in more detail.

docs/variant_merging.md
arostamianfar: Not sure if I understand your comment correctly, but I added a statement about how repeated fields are handled in general.

docs/variant_merging.md
arostamianfar: thanks!

docs/variant_merging.md
arostamianfar: Done. Also clarified when INFO and FORMAT keys are the same.

docs/variant_merging.md
arostamianfar: Good point, but I think this is independent of the merging logic, as it can happen when loading multiple files without merging. I'm planning to write a doc on 'importing multiple files' and will explain this as well.

docs/variant_merging.md
arostamianfar: ah, good catch! It should be null everywhere :)

docs/variant_merging.md
arostamianfar: Logic is TBD :) Ideally, we should break it into 3 sections (what our existing code tries to do actually), but it gets very complicated, so I'm thinking for the initial version we'd just output SNPs and totally ignore non-variants.
Anyway, I don't want to go into too much detail here as users can't use this feature now anyway.

docs/variant_merging.md
bashir2: Yes, that is what I meant.

docs/variant_merging.md
bashir2: Great.

docs/variant_merging.md
bashir2: Okay makes sense.

arostamianfar:
ah, I just noticed that my documentation on how duplicate keys are handled was incorrect :P We actually throw an error if the same key is present in both INFO and FORMAT and one is said to be moved to calls. Updated the doc. PTAL.

Issue #122, opened 2018-02-27
Cast Number=A/G/R fields to a list in FORMAT fields


arostamianfar:
We currently [cast Number=. fields to a list](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/beam_io/vcfio.py#L861) for FORMAT fields. We should do the same for Number=A/G/R fields.

This is blocking imports on Issue #119.

Note: ultimately, the schema checks (Issue #101), will fix the schema errors, but I still think it makes sense to always return a list natively for Number=A/G/R fields as they should be the same logic as INFO fields (this is also done by PyVCF).

nmousavi:
New image with fix PR #123  is released.

PR #123, opened 2018-02-27
Convert special FORMAT fields to list.

nmousavi:
FORMAT fileds with Number=A,G,R are converted to list even when they are
single valued.

Fixes issue  #122 

TESTED:
  unit test


nmousavi:
Np :)

PR #124, opened 2018-02-27
Update the development guide doc

allieychen:
Update the development guide doc.
- Add IntelliJ IDE setup.
- Add more details.

### DIFF
--- a/docs/development_guide.md
+++ b/docs/development_guide.md
@@ -17,7 +17,7 @@ for more information on using pull requests.
 ### Fork the repository on Github
 
 Visit the
-[gcp-variant-transforms repository](https://github.com/googlegenomics/gcp-variant-transforms) 
+[gcp-variant-transforms repository](https://github.com/googlegenomics/gcp-variant-transforms)
 to create your own fork of the repository. See
 [https://guides.github.com/activities/forking/](https://guides.github.com/activities/forking/)
 for more information. Do not create branches on the main repository. Meanwhile,
--- a/docs/development_guide.md
+++ b/docs/development_guide.md
@@ -154,8 +154,8 @@ script.
 
 Before pushing changes, make sure the pylint checks pass. To install pylint:
 ```bash
-.[PATH_TO_VENV]/bin/activate
-pip install pylint
+source [PATH_TO_VENV]/bin/activate
+pip install --upgrade pylint
 ```
 Then run:
 ```bash
--- a/docs/development_guide.md
+++ b/docs/development_guide.md
@@ -15,8 +15,13 @@ for more information on using pull requests.
 ## Setup
 
 ### Fork the repository on Github
-Visit the [gcp-variant-transforms repository](https://github.com/googlegenomics/gcp-variant-transforms) to create your own fork of the repository. See [https://guides.github.com/activities/forking/](https://guides.github.com/activities/forking/)
-for more information. Do not create branches on the main repository. Meanwhile, do not commit anything to the master of the forked repository to keep the syncing process simple.
+Visit the 
+[gcp-variant-transforms repository](https://github.com/googlegenomics/gcp-variant-transforms) 
+to create your own fork of the repository. See 
+[https://guides.github.com/activities/forking/](https://guides.github.com/activities/forking/)
+for more information. Do not create branches on the main repository. Meanwhile, 
+do not commit anything to the master of the forked repository to keep the 
+syncing process simple.
 
 
 ### Setup dev environment
@@ -51,34 +56,50 @@ installed packages which is currently safe to ignore. For details see
 
 ### Setup IDE
 
-You may choose any IDE as you like. The following steps are intended for IntelliJ users.
+You may choose any IDE as you like. The following steps are intended for 
+IntelliJ users.
 
 #### Install IntelliJ IDE
 
-Download [IntelliJ IDEA Community Edition](https://www.jetbrains.com/idea/download/#section=linux) and install.
+Download 
+[IntelliJ IDEA Community Edition](https://www.jetbrains.com/idea/download/#section=linux) 
+and install.
 
 #### Install Python plugin
 
 Choose File | Settings on the main menu, and then go to Plugins.
 Click the Install JetBrains plugin button.
-In the dialog that opens, search for Python Community Edition and then install the plugin.
+In the dialog that opens, search for Python Community Edition and then install 
+the plugin.
 
-For more details, refer to [Install plugins](https://www.jetbrains.com/help/idea/installing-updating-and-uninstalling-repository-plugins.html).
+For more details, refer to 
+[Install plugins](https://www.jetbrains.com/help/idea/installing-updating-and-uninstalling-repository-plugins.html).
 
 #### Setup IntelliJ SDK
 
 1. Choose File | Project Structure on the main menu, and then go to Project.
-2. Create a new project SDK by clicking the New button, choose Python SDK then add Local.
-3. In the dialog that opens, click the Virtual Environment node. Select New environment, and specify the location of the new virtual environment. Note that the folder where the new virtual environment should be located must be empty! For the Base interpreter, add the python path gcp-variant-transforms/venv/bin/python under the created virtualenv.
+2. Create a new project SDK by clicking the New button, choose Python SDK then 
+add Local.
+3. In the dialog that opens, click the Virtual Environment node. Select New 
+environment, and specify the location of the new virtual environment. Note that 
+the folder where the new virtual environment should be located must be empty! 
+For the Base interpreter, add the python path 
+gcp-variant-transforms/venv/bin/python under the created virtualenv.
 
 #### Code Inspection
 
-The inspection profile in .idea/inspectionProfiles/Project_Default.xml is checked into the git repository and can be imported into File | Settings | Editor | Inspections.
+The inspection profile in .idea/inspectionProfiles/Project_Default.xml is 
+checked into the git repository and can be imported into 
+File | Settings | Editor | Inspections.
 
-Code inspections can be run from the Analyze menu. The result window can be accessed from View > Tool Windows.
+Code inspections can be run from the Analyze menu. The result window can be 
+accessed from View > Tool Windows.
 
 #### Code Style
-To comply with pylint coding style, you may change the default line length in File | Settings | Editor | Code Style. Set the hard wrap at 80 columns and check Wrap on typing. Further, go to Python in the dropdown list, you can set the indent to 2 and continuation indent to 4.
+To comply with pylint coding style, you may change the default line length in 
+File | Settings | Editor | Code Style. Set the hard wrap at 80 columns and 
+check Wrap on typing. Further, go to Python in the dropdown list, you can 
+set the indent to 2 and continuation indent to 4.
 
 ## Making Changes
 ### Create a branch in your forked repository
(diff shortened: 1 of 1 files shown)


docs/development_guide.md
```diff
@@ -15,11 +15,9 @@ for more information on using pull requests.
 ## Setup
 
 ### Fork the repository on Github
-Visit the
-[gcp-variant-transforms repository](https://github.com/googlegenomics/gcp-variant-transforms)
-to create your own fork of the repository. See
-[https://guides.github.com/activities/forking/](https://guides.github.com/activities/forking/)
-for more information.
+Visit the [gcp-variant-transforms repository](https://github.com/googlegenomics/gcp-variant-transforms) to create your own fork of the repository. See [https://guides.github.com/activities/forking/](https://guides.github.com/activities/forking/)
```
arostamianfar: Please keep each line <80 chars. Links that are longer than 80chars should be in separate lines.
There is a google style guide on .md docs somewhere as well (i'll ping you when I've found it).

docs/development_guide.md
```diff
@@ -101,8 +138,12 @@ git push -u origin <branch name>
 To commit and push those changes to your branch.
 
 ### Syncing your branch
-If you want to pull in changes from the target branch (i.e. googlegenomic:master),
-run:
+To update the forked repository, you may follow instructions on [syncing a fork](https://help.github.com/articles/syncing-a-fork/), or simply set upstream to the main repository and do:
+```bash
```
arostamianfar: Do you need this section? This is the same as the instructions below except it doesn't have --rebase (and I think we needed to always pull with rebase for things to work smoothly).

allieychen: I am thinking there are two scenarios. 
Before checking out a new branch from the forked master branch, we would like to keep the forked master branch to be synced with the remote master branch, so that the commits from other people will apply. In that case, we update the forked master repository.
The second portion is after we made changes to local branch, we pull in the latest commits, then replay our local changes.
But you are right, the second portion is sufficient for the sync. I will remove this part.

docs/development_guide.md
```diff
@@ -15,11 +15,14 @@ for more information on using pull requests.
 ## Setup
 
 ### Fork the repository on Github
-Visit the
-[gcp-variant-transforms repository](https://github.com/googlegenomics/gcp-variant-transforms)
-to create your own fork of the repository. See
+Visit the
```
arostamianfar: nit: please remove trailing space from here and elsewhere.

docs/development_guide.md
```diff
@@ -51,7 +55,56 @@ Note that after running the above command we get some dependency conflicts in
...
+#### Setup IntelliJ SDK
+
+1. Choose File | Project Structure on the main menu, and then go to Project.
+2. Create a new project SDK by clicking the New button, choose Python SDK then
+add Local.
+3. In the dialog that opens, click the Virtual Environment node. Select New
+environment, and specify the location of the new virtual environment. Note that
+the folder where the new virtual environment should be located must be empty!
+For the Base interpreter, add the python path
+gcp-variant-transforms/venv/bin/python under the created virtualenv.
```
bashir2: I put my venv under gcp-variant-transforms but I think some people don't (@arostamianfar do you do this?). So maybe replace "gcp-variant-transforms" with [PATH_TO_VENV] and mention that this 'venv' directory is what we created in "Setup virtualenv" above.

docs/development_guide.md
```diff
@@ -51,7 +55,56 @@ Note that after running the above command we get some dependency conflicts in
...
+File | Settings | Editor | Inspections.
+
+Code inspections can be run from the Analyze menu. The result window can be
+accessed from View > Tool Windows.
+
+#### Code Style
+
+To comply with pylint coding style, you may change the default line length in
+File | Settings | Editor | Code Style. Set the hard wrap at 80 columns and
+check Wrap on typing. Further, go to Python in the dropdown list, you can
```
bashir2: I have not done this, how annoying is it when you want to have a line over 80? (e.g., some of our "# pylint" comments for imports are like this).

docs/development_guide.md
```diff
@@ -88,6 +142,15 @@ For other projects you can use the `--project` and `--gs_dir` options of the
...
 
 ### Pushing changes to your fork's branch
+
+Before pushing changes, make sure the pylint checks pass. To install pylint:
+```bash
+sudo apt-get install pylint
+```
+Then run:
+```bash
+pylint gcp_variant_transforms
```
bashir2: Hmm, I think you should use the `.pylintrc` file that is checked into the git repo not the default RC file. I usually run `pylint --rcfile=.pylintrc gcp_variant_transforms/`

docs/development_guide.md
```diff
@@ -88,6 +142,15 @@ For other projects you can use the `--project` and `--gs_dir` options of the
 script.
 
 ### Pushing changes to your fork's branch
+
+Before pushing changes, make sure the pylint checks pass. To install pylint:
+```bash
+sudo apt-get install pylint
```
bashir2: Did you successfully use the pylint that comes with your linux distribution? I had problems with mine originally (Trusty at the time) and I tried again on Rodete but it gives me a lot of warnings (it is an older version). I use a  local pylint in my venv: `pip install --upgrade pylint`

docs/development_guide.md
```diff
@@ -101,6 +164,7 @@ git push -u origin <branch name>
 To commit and push those changes to your branch.
 
 ### Syncing your branch
+
```
bashir2: Are these blank lines needed? If not, consider removing them here and elsewhere.

docs/development_guide.md
```diff
@@ -51,7 +55,56 @@ Note that after running the above command we get some dependency conflicts in
...
+Click the Install JetBrains plugin button.
+In the dialog that opens, search for Python Community Edition and then install
+the plugin.
+
+For more details, refer to
+[Install plugins](https://www.jetbrains.com/help/idea/installing-updating-and-uninstalling-repository-plugins.html).
+
+#### Setup IntelliJ SDK
+
+1. Choose File | Project Structure on the main menu, and then go to Project.
```
bashir2: Please add an initial step on how the project is created. IIRC, the project could have been created in the default location but then there was somewhere that sources were being set, correct?

docs/development_guide.md
```diff
@@ -51,7 +55,56 @@ Note that after running the above command we get some dependency conflicts in
...
+gcp-variant-transforms/venv/bin/python under the created virtualenv.
+
+#### Code Inspection
+
+The inspection profile in .idea/inspectionProfiles/Project_Default.xml is
+checked into the git repository and can be imported into
+File | Settings | Editor | Inspections.
+
+Code inspections can be run from the Analyze menu. The result window can be
+accessed from View > Tool Windows.
```
bashir2: Should you add any comments about code inclusion/exclusion? I have excluded everything other than gcp_variant_transforms (see screenshot) and IIRC this was important to trim the long list of inspection warnings.
project_structure

docs/development_guide.md
allieychen: I referred to g3doc style guide, and it suggests newlines before and after the heading. If we do differently, I can remove it.
https://g3doc.corp.google.com/corp/g3doc/docs/reference/style.md#heading_spacing

docs/development_guide.md
allieychen: Done.

docs/development_guide.md
allieychen: I am not aware of a better solution other than manually adjust the line <= 80 without setting this. I would say it would be more annoying without it since it happens more often than the other case.

docs/development_guide.md
allieychen: Sorry, I tried both and it seems that they are doing the same checking. Does it mean that the default file is .pylintrc?

docs/development_guide.md
allieychen: Yes. I tried pip install --upgrade pylint in the first place and somehow it did not work.

docs/development_guide.md
allieychen: Done. Thanks.

docs/development_guide.md
allieychen: Done.

allieychen:
PTAL :)

allieychen:
I edited the --amend part.

docs/development_guide.md
bashir2: Agreed.

docs/development_guide.md
bashir2: So please update as we discussed.

docs/development_guide.md
bashir2: ditto

docs/development_guide.md
bashir2: Oh, I see, then maybe consider updating it for every section in this doc.

docs/development_guide.md
```diff
@@ -131,8 +204,11 @@ description of how you have tested your change. As a minimum you should have
 unit-test coverage for your change and make sure integration tests pass.
 
 ### Updating changes
-After making changes, you must again add, commit, and push those changes. Rather
-than creating new commits for related changes please run the following:
+
+After making changes, you must again add, commit, and push those changes.
+If you are making changes after a review process, you may create new commits for
+related changes by following the steps stated in "Pushing changes to your fork's
+branch". Otherwise, please run the following:
```
bashir2: Please add a note like: "Prefer to have one commit per each review round such that your reviewers can easily check what you have changed since last time."

docs/development_guide.md
```diff
@@ -51,7 +55,62 @@ Note that after running the above command we get some dependency conflicts in
...
+In the dialog that opens, search for Python Community Edition and then install
+the plugin.
+
+For more details, refer to
+[Install plugins](https://www.jetbrains.com/help/idea/installing-updating-and-uninstalling-repository-plugins.html).
+
+#### Create Project
+
+Choose File | New | Project on the main menu, and create a new Python project
+in the dialog that opens. To setup the Project SDK, follow the following steps.
```
bashir2: Please add a step before this which says where Project SDK is, i.e., from File | Project Structure go to the Project sub-menu ...

docs/development_guide.md
allieychen: Done.

docs/development_guide.md
allieychen: Done.

docs/development_guide.md
allieychen: That is one way to setup the SDK. Also, we can setup it directly in the new project dialog.

docs/development_guide.md
allieychen: Done.

allieychen:
PTAL :)

docs/development_guide.md
bashir2: Oh, I see.

docs/development_guide.md
```diff
@@ -88,6 +151,16 @@ For other projects you can use the `--project` and `--gs_dir` options of the
 script.
 
 ### Pushing changes to your fork's branch
+
+Before pushing changes, make sure the pylint checks pass. To install pylint:
+```bash
+.[PATH_TO_VENV]/bin/activate
```
bashir2: You need a space after '.'; also in general I prefer `source` to using '.', i.e.,
`source [PATH_TO_VENV]/bin/activate`
as it is more clear what is being done.

docs/development_guide.md
```diff
@@ -88,6 +151,16 @@ For other projects you can use the `--project` and `--gs_dir` options of the
...
+pip install pylint
```
bashir2: nit: consider adding `--upgrade` to be consistent with elsewhere.

docs/development_guide.md
allieychen: Thanks. Done.

docs/development_guide.md
allieychen: Done.

allieychen:
Thanks, PTAL :)

PR #126, opened 2018-03-01
 Extends variant annotation alt matching.

bashir2:
Please note that this is on top of PR #125 and as you can see it has two commits. The first one is the previous PR and can be skipped. So please only review the second commit.

Tested:
Updated unit-tests; ran integration test; manually checked the output table for valid-4.2_VEP.vcf case. For example this query used to return 11 but now returns 26:
`SELECT COUNT(0) FROM (
SELECT start_position, alt.alt, CSQ.Consequence
FROM vcf_test_imports.VEP_result_for_valid_4_2_fixed AS T, T.alternate_bases AS alt, alt.CSQ as CSQ)`
But even with these changes we are not covering all annotation corner cases in gnomAD!

Issue #81

### DIFF
--- a/gcp_variant_transforms/libs/processed_variant.py
+++ b/gcp_variant_transforms/libs/processed_variant.py
@@ -24,6 +24,8 @@ from __future__ import absolute_import
 
 import enum
 import logging
+import os
+import re
 
 from collections import defaultdict
 from typing import Dict, List, Any  # pylint: disable=unused-import
@@ -180,13 +182,13 @@ class ProcessedVariantFactory(object):
     """Sets the internal state of the factory class.
 
     Args:
-      header_fields (:class:`vcf_header_parser.HeaderFields`): Header
-        information used for parsing and splitting INFO fields of thei variant.
-      split_alternate_allele_info_fields (bool): If True, splits fields with
+      header_fields: Header information used for parsing and splitting INFO
+        fields of thei variant.
+      split_alternate_allele_info_fields: If True, splits fields with
         field_count='A' (i.e., one value for each alternate) among alternates.
-      annotation_fields (List[str]): If provided, this is the list of
-        INFO field names that store variant annotations. The format of how
-        annotations are stored and their names are extracted from header_fields.
+      annotation_fields: If provided, this is the list of INFO field names that
+        store variant annotations. The format of how annotations are stored and
+        their names are extracted from header_fields.
     """
     self._header_fields = header_fields
     self._split_alternate_allele_info_fields = (
@@ -199,6 +201,9 @@ class ProcessedVariantFactory(object):
         _CounterEnum.ANNOTATION.value)
     self._annotation_alt_mismatch_counter = cfactory.create_counter(
         _CounterEnum.ANNOTATION_ALT_MISMATCH.value)
+    self._annotation_processor = _AnnotationProcessor(
+        annotation_fields, self._header_fields, self._annotation_counter,
+        self._annotation_alt_mismatch_counter)
 
   def create_processed_variant(self, variant):
     # type: (vcfio.Variant) -> ProcessedVariant
@@ -216,7 +221,8 @@ class ProcessedVariantFactory(object):
           variant_info.field_count == _FIELD_COUNT_ALTERNATE_ALLELE):
         self._add_per_alt_info(proc_var, key, variant_info.data)
       elif key in self._annotation_field_set:
-        self._add_annotation(proc_var, key, variant_info.data)
+        self._annotation_processor.add_annotation_data(
+            proc_var, key, variant_info.data)
       else:
         proc_var._non_alt_info[key] = variant_info.data
     return proc_var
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -235,51 +244,18 @@ def _add_annotation(self, proc_var, field_name, data):
...
-    # type: (List[str], List[str]) -> Dict[str, str]
-    if len(annotation_names) != len(annotations) - 1:
-      raise ValueError('Expected {} annotations, got {}'.format(
-          len(annotation_names), len(annotations) - 1))
-    annotation_dict = {}
-    for index, name in enumerate(annotation_names):
-      annotation_dict[name] = annotations[index + 1]
-    return annotation_dict
+    annotation_processor = _AnnotationProcessor(
+        proc_var, field_name, self._annotation_counter,
```
arostamianfar: I'm starting to feel uneasy about the mutable proc_var getting passed around like this as it's no longer clear what is being mutated and in what way.
Given that the only mutable part is the alternate base, then does it make sense to redesign this in such a way that you only pass alternate_base? Looks like you also need reference_base for calculating the common prefix, but that can be done outside that class as a 'util' method?

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -235,51 +244,18 @@ def _add_annotation(self, proc_var, field_name, data):
...
-  def _create_annotation_map(self, annotations, annotation_names):
-    # type: (List[str], List[str]) -> Dict[str, str]
-    if len(annotation_names) != len(annotations) - 1:
-      raise ValueError('Expected {} annotations, got {}'.format(
-          len(annotation_names), len(annotations) - 1))
-    annotation_dict = {}
-    for index, name in enumerate(annotation_names):
-      annotation_dict[name] = annotations[index + 1]
-    return annotation_dict
+    annotation_processor = _AnnotationProcessor(
```
arostamianfar: From the API of this class, it's not really clear why some args are in the constructor vs. in the add_annotation_data method (e.g. field_name and annotation_names). It's also not clear why a new instance of the AnnotationProcessor class is needed for every annotation. I think a better design may be to have a single AnnotationProcessor (that has counters) and you'd use that instance for processing any annotation? This paradigm also resolves the issue of why some fields would be in the constructor vs. in the methods.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -343,6 +319,101 @@ def info_is_in_alt_bases(self, info_field_name):
...
+    # Handling symbolic ALTs.
+    id_match = _SYMBOLIC_ALT_RE.match(alt_bases)
+    if id_match and id_match.group('ID') == annotation_alt:
+      return True
+    # Handling breakend ALTs.
+    # TODO(bashir2): Check if the following logic is documented anywhere! I
+    # could not find it explicitly in any documentation but that's how I saw
+    # VEP does it in some examples I ran.
+    breakend_match = _BREAKEND_ALT_RE.match(alt_bases)
+    if (breakend_match and breakend_match.group('up_to_chr') == annotation_alt):
```
arostamianfar: nit: please remove the surrounding brackets.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -343,6 +319,101 @@ def info_is_in_alt_bases(self, info_field_name):
...
+    else:
+      self._mismatch_counter.inc()
+      logging.warning(
+          'Could not find matching alternate bases for %s in '
+          'annotation filed %s', alt_bases, self._anntation_field_name)
+
+  def _alt_matches_annotation_alt(self, alt_bases, annotation_alt):
+    # type: (str, str) -> bool
+    """Returns true if `alt_bases` matches`annotation_alt`
+    See: https://useast.ensembl.org/info/docs/tools/vep/vep_formats.html
```
arostamianfar: Please clarify which section in the doc talks about this logic. Also, please summarize the logic here as well (links can become broken over time) and provide a few examples.

bashir2:
To check the changes since your last review, please check [this commit](https://github.com/bashir2/gcp-variant-transforms/commit/1d3fa83f8c338ccfc7768b8004e88d2c6caa7462).

bashir2:
All of the 357 mismatch cases, that I reported for that gnomAD small sample, were covered by implementing what [--minimal option of VEP](https://useast.ensembl.org/info/docs/tools/vep/script/vep_options.html#opt_minimal) does to annotation ALTs. So apparently gnomAD annotation has been done by that feature of VEP which is unfortunate because it can add ambiguous cases (details to come in the next PR which needs schema changes too that I have not done yet). Anyways, a run with no mismatch can be found [here](https://pantheon.corp.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2018-03-03_15_07_06-11713068003808116107?project=gcp-variant-transforms-test&authuser=1&organizationId=433637338589).

bashir2:
PR updated with merging/conflict resolution of changes on `master` and PR #125, plus some more `_AnnotationProcessor` shuffling for better performance.

bashir2:
Just a reminder that this PR (and the next one on top of this) are ready for another round of review for a while now.

Thanks

bashir2:
Added the TODO for the annotation class and merged new changes from master.

Please approve again as the `push -f` overwrote the approval bit; thanks.

PR #127, opened 2018-03-02
Uses the macros to replace the common queries.

allieychen:
Define  NUM_ROWS, SUM_START, SUM_END in QueryFormatter, and replaces them in the query to avoid duplicate code.

TESTED:
deploy_and_run_tests.

### DIFF
--- a/gcp_variant_transforms/testing/integration/run_tests.py
+++ b/gcp_variant_transforms/testing/integration/run_tests.py
@@ -41,6 +41,7 @@ import sys
 import time
 
 from datetime import datetime
+from typing import List  # pylint: disable=unused-import
 # TODO(bashir2): Figure out why pylint can't find this.
 # pylint: disable=no-name-in-module,import-error
 from google.cloud import bigquery
@@ -202,13 +203,12 @@ class QueryFormatter(object):
     self._table_name = table_name
 
   def format_query(self, query):
-    # type: (list[str]) -> str
+    # type: (List[str]) -> str
     """Formats the given ``query``.
 
     Formatting logic is as follows:
     - Concatenates ``query`` parts into one string.
-    - Replaces macro NUM_ROWS_QUERY/SUM_START_QUERY/SUM_END_QUERY with the
-    corresponding value defined in _QueryMacros.
+    - Replaces macro with the corresponding value defined in _QueryMacros.
     - Replaces TABLE_NAME with the table associated for the query.
     """
     return self._replace_variables(self._replace_macros((' ').join(query)))
--- a/gcp_variant_transforms/testing/integration/run_tests.py
+++ b/gcp_variant_transforms/testing/integration/run_tests.py
@@ -148,7 +148,6 @@ class TestCase(object):
   def validate_table(self):
     """Runs queries against the output table and verifies results."""
     client = bigquery.Client(project=self._project)
-    # TODO(yifangchen): Create macros for common queries
     query_formatter = QueryFormatter(self._table_name)
     for assertion_config in self._assertion_configs:
       query = query_formatter.format_query(assertion_config['query'])
@@ -188,22 +187,31 @@ class QueryAssertion(object):
 class QueryFormatter(object):
   """Formats a query.
 
-  Replaces keyword TABLE_NAME and eventually macros in the query.
+  Replaces keyword TABLE_NAME and macros in the query.
   """
+
+  NUM_ROWS = 'SELECT COUNT(0) AS num_rows FROM {TABLE_NAME}'
+  SUM_START = 'SELECT SUM(start_position) AS sum_start FROM {TABLE_NAME}'
+  SUM_END = 'SELECT SUM(end_position) AS sum_end FROM {TABLE_NAME}'
 
   def __init__(self, table_name):
     # type: (str) -> None
     self._table_name = table_name
 
   def format_query(self, query):
-    # type: (List[str]) -> str
+    # type: (list[str]) -> str
     """Formats the given ``query``.
 
     Formatting logic is as follows:
     - Concatenates ``query`` parts into one string.
+    - Replaces NUM_ROWS/SUM_START/SUM_END with the corresponding query.
     - Replaces TABLE_NAME with the table associated for the query.
     """
-    return (' ').join(query).format(TABLE_NAME=self._table_name)
+    return (' ').join(query).format(NUM_ROWS=self.NUM_ROWS,
+                                    SUM_START=self.SUM_START,
+                                    SUM_END=self.SUM_END,
+                                    TABLE_NAME=self._table_name).format(
+                                        TABLE_NAME=self._table_name)
 
 
 class TestContextManager(object):
--- a/gcp_variant_transforms/testing/integration/run_tests.py
+++ b/gcp_variant_transforms/testing/integration/run_tests.py
@@ -33,6 +33,7 @@ and populating) and only do the validation, use --revalidation_dataset_id, e.g.,
 """
 
 import argparse
+import enum
 import json
 import multiprocessing
 import os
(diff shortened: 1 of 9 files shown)


allieychen:
Advice is needed for line 213 and line 214 in run_tests.py. It is quite ugly to replace TABLE_NAME twice.

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -188,22 +187,31 @@ def run_assertion(self):
...
+    # type: (list[str]) -> str
     """Formats the given ``query``.
 
     Formatting logic is as follows:
     - Concatenates ``query`` parts into one string.
+    - Replaces NUM_ROWS/SUM_START/SUM_END with the corresponding query.
     - Replaces TABLE_NAME with the table associated for the query.
     """
-    return (' ').join(query).format(TABLE_NAME=self._table_name)
+    return (' ').join(query).format(NUM_ROWS=self.NUM_ROWS,
```
arostamianfar: Consider breaking this up into two helper methods: `_replace_macros` and `_replace_variables`.
so, it would be `return _replace_variables(_replace_macros(' '.join(query))` 
you can optionally create meaningful variable names for the intermediate strings.

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -188,22 +187,31 @@ def run_assertion(self):
 class QueryFormatter(object):
   """Formats a query.
 
-  Replaces keyword TABLE_NAME and eventually macros in the query.
+  Replaces keyword TABLE_NAME and macros in the query.
   """
 
+  NUM_ROWS = 'SELECT COUNT(0) AS num_rows FROM {TABLE_NAME}'
```
arostamianfar: nit: to make these distinct from the rest of the variables (e.g. TABLE_NAME), what do you think of adding a `_QUERY` suffix to all of them?

arostamianfar: I think it's cleaner to create an enum of these macros. It both provides a natural grouping of these constants and also is easier to loop through them (e.g. adding a new macro is just as easy as adding a new entry in that Enum).
So, it would be something like:
```
import enum
class _QueryMacros(enum.Enum):  # nested under the QueryFormatter class
  NUM_ROWS_QUERY = 'SELECT ...'
  SUM_START_QUERY = 'SELECT ...'

def _replace_macros(query):
  for macro in _QueryMacros:
    if macro.name == query:
      return macro.value
  else:
    return query 
``` 

This also means that you'd no longer need the surrounding `{}` brackets in the test def, which further distinguishes macros from regular variable names.

gcp_variant_transforms/testing/integration/small_tests/valid_4_0.json
```diff
@@ -6,14 +6,25 @@
   "assertion_configs": [
     {
       "query": [
-        "SELECT COUNT(0) AS num_rows, ",
-        "  SUM(start_position) AS sum_start, ",
-        "  SUM(end_position) AS sum_end ",
-        "FROM {TABLE_NAME}"
+        "{NUM_ROWS}"
```
arostamianfar: nit: I think it's cleaner to just put these on one line now that you don't have multiple lines.
e.g.
```
{
  "query": [ "NUM_ROWS_QUERY" ],
  "expected_result": { "num_rows": 5 }
},
{
  "query": [ "SUM_START_QUERY" ],
  "expected_result": { "sum_start": 55 }
},

```

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -188,22 +187,31 @@ def run_assertion(self):
...
+  SUM_START = 'SELECT SUM(start_position) AS sum_start FROM {TABLE_NAME}'
+  SUM_END = 'SELECT SUM(end_position) AS sum_end FROM {TABLE_NAME}'
+
   def __init__(self, table_name):
     # type: (str) -> None
     self._table_name = table_name
 
   def format_query(self, query):
-    # type: (List[str]) -> str
+    # type: (list[str]) -> str
```
arostamianfar: Interesting. Is the correct format `list` instead of `List`?

allieychen: I think so, List actually gives a warning.

arostamianfar: @bashir2 FYI, looks like we should be using `list` instead of `List`

gcp_variant_transforms/testing/integration/small_tests/valid_4_0.json
allieychen: Good point!

gcp_variant_transforms/testing/integration/run_tests.py
allieychen: Done.

gcp_variant_transforms/testing/integration/run_tests.py
allieychen: Done.

gcp_variant_transforms/testing/integration/run_tests.py
allieychen: Thanks for such a detailed comment! DONE.

allieychen: I actually got a warning on "for macro in _QueryMacros", says "Expected expected collections.iterable got _QueryMacros instead. Any idea how to fix this? I have found one way which loops through the __members__.items, but I don't think it is as readable as current code.

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -188,22 +188,39 @@ def run_assertion(self):
...
     """Formats the given ``query``.
 
     Formatting logic is as follows:
     - Concatenates ``query`` parts into one string.
+    - Replaces macro NUM_ROWS_QUERY/SUM_START_QUERY/SUM_END_QUERY with the
```
bashir2: Consider not listing the members of _QueryMacros here such that adding a new query or changing the name of one in _QueryMacros does not an update of this documentation (which will probably not happen making the documentation obsolete in future).

gcp_variant_transforms/testing/integration/run_tests.py
allieychen: I see. I will correct it.

gcp_variant_transforms/testing/integration/run_tests.py
allieychen: Done.

allieychen:
PTAL.

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -188,9 +189,15 @@ def run_assertion(self):
 class QueryFormatter(object):
   """Formats a query.
 
-  Replaces keyword TABLE_NAME and eventually macros in the query.
+  Replaces macros and variable TABLE_NAME in the query.
   """
 
+  class _QueryMacros(enum.Enum):
+    NUM_ROWS_QUERY = 'SELECT COUNT(0) AS num_rows FROM {TABLE_NAME}'
+    SUM_START_QUERY = ('SELECT SUM(start_position) AS sum_start FROM {'
```
arostamianfar: nit: consider the following format as it's more readable (in general, try to avoid breaking lines as much as possible).
```
   SUM_START_QUERY = (
       'SELECT SUM(start_position) AS sum_start FROM {TABLE_NAME}')
```

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -201,9 +208,19 @@ def format_query(self, query):
 
     Formatting logic is as follows:
     - Concatenates ``query`` parts into one string.
+    - Replaces macro with the corresponding value defined in _QueryMacros.
     - Replaces TABLE_NAME with the table associated for the query.
     """
-    return (' ').join(query).format(TABLE_NAME=self._table_name)
+    return self._replace_variables(self._replace_macros((' ').join(query)))
```
arostamianfar: nit: do you need the brackets around `' '`? i.e. does `..._macros(' '.join(query)))` work?

gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -188,9 +189,15 @@ def run_assertion(self):
 class QueryFormatter(object):
   """Formats a query.
 
-  Replaces keyword TABLE_NAME and eventually macros in the query.
+  Replaces macros and variable TABLE_NAME in the query.
```
arostamianfar: nit: by the same token as Bashir's comment, please remove 'table_name' from here as well. (just "Replaces macros and variables in the query").

gcp_variant_transforms/testing/integration/run_tests.py
allieychen: Thanks, done.

gcp_variant_transforms/testing/integration/run_tests.py
allieychen: Done.

gcp_variant_transforms/testing/integration/run_tests.py
allieychen: Done.

allieychen:
Thanks. PTAL :)

PR #128, opened 2018-03-02
Added an INFO message for the full command.

bashir2:
Tested: Ran manually and checked the new log message.


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -209,6 +210,10 @@ def _validate_args(options, parsed_args):
 
 def run(argv=None):
   """Runs VCF to BigQuery pipeline."""
+  if argv:
```
arostamianfar: nit: I think you can combine this as `logging.info('Command ...', ' '.join(argv or sys.argv))`

bashir2: Done.

Issue #99, opened 2018-01-26
Create large integration tests.


bashir2:
The current integration tests we have are small and do not cover enough variations in VCF inputs. To automate our release process, we need to add larger cases and do both deeper validation and performance analysis to make sure we are not introducing any regression. One potential dataset to use is 1000 Genomes (see Issue #61 too).

arostamianfar:
A few additional notes:
- I think we should have a 'large' test based on Platinum Genomes (`gs://genomics-public-data/platinum-genomes/vcf/*.vcf`) that runs on every submit. It takes about 15mins (with merging) on 64x8-core machines.
- 1000 genome data would be 'huge' and we'd only run it on a release candidate. By default, the integration test framework should ignore 'huge' tests.
- We should also test with a large number of small files (e.g. copy valid-4.2.vcf many times). This is to test the 3 main 'modes' of the pipeline: SMALL (already tested), MEDIUM (need to copy the file 100+ times), and LARGE (50k+ files). These should also run pretty quickly (<20mins) with a reasonable number of machines.

arostamianfar:
FYI, bumping this to P1 together with Issue #106 (I'd say Issue #106 is P1.5 though :p ). I'm still a bit concerned that we don't have enough integration testing coverage as we build new features.

Issue #110, opened 2018-02-13
Ensure the Beam SDK is using cython modules


arostamianfar:
Apparently, Cython is not used by default when just installing through pip. See this thread: https://lists.apache.org/thread.html/90eb2605b42f5f8ce9d729fdac43af7a914a6e70f4f1cf8ae16e8ce4@%3Cdev.beam.apache.org%3E

I verified that cython is not installed in our docker image (through pip freeze). We should install it prior to installing Beam and determine if there is a performance boost as a result.

PR #132, opened 2018-03-08
Adds large tests to integration tests

bashir2:
Tested:
Manually ran deploy_and_run_tests.sh --include_large_tests

Issue: #99

### DIFF
--- a/gcp_variant_transforms/testing/integration/large_tests/platinum_no_merge.json
+++ b/gcp_variant_transforms/testing/integration/large_tests/platinum_no_merge.json
@@ -5,6 +5,7 @@
   "runner": "DataflowRunner",
   "worker_machine_type": "n1-standard-16",
   "max_num_workers": "20",
+  "num_workers": "20",
   "assertion_configs": [
     {
       "query": ["NUM_ROWS_QUERY"],
--- a/gcp_variant_transforms/testing/integration/large_tests/platinum_no_merge.json
+++ b/gcp_variant_transforms/testing/integration/large_tests/platinum_no_merge.json
@@ -4,7 +4,7 @@
   "input_pattern": "gs://genomics-public-data/platinum-genomes/vcf/*.vcf",
   "runner": "DataflowRunner",
   "worker_machine_type": "n1-standard-16",
-  "max_num_workers": "5",
+  "max_num_workers": "20",
   "assertion_configs": [
     {
       "query": ["NUM_ROWS_QUERY"],
--- a/gcp_variant_transforms/testing/integration/large_tests/platinum_no_merge.json
+++ b/gcp_variant_transforms/testing/integration/large_tests/platinum_no_merge.json
@@ -1,0 +1,22 @@
+{
+  "test_name": "platinum-no-merge",
+  "table_name": "platinum_no_merge",
+  "input_pattern": "gs://genomics-public-data/platinum-genomes/vcf/*.vcf",
+  "runner": "DataflowRunner",
+  "worker_machine_type": "n1-standard-16",
+  "max_num_workers": "5",
+  "assertion_configs": [
+    {
+      "query": ["NUM_ROWS_QUERY"],
+      "expected_result": {"num_rows": 309551691}
+    },
+    {
+      "query": ["SUM_START_QUERY"],
+      "expected_result": {"sum_start": 24107891293483855}
+    },
+    {
+      "query": ["SUM_END_QUERY"],
+      "expected_result": {"sum_end": 24107908471506508}
+    }
+  ]
+}
--- a/gcp_variant_transforms/testing/integration/large_tests/platinum_with_merge.json
+++ b/gcp_variant_transforms/testing/integration/large_tests/platinum_with_merge.json
@@ -6,6 +6,7 @@
   "variant_merge_strategy": "MOVE_TO_CALLS",
   "worker_machine_type": "n1-standard-16",
   "max_num_workers": "20",
+  "num_workers": "20",
   "assertion_configs": [
     {
       "query": ["NUM_ROWS_QUERY"],
--- a/gcp_variant_transforms/testing/integration/large_tests/platinum_with_merge.json
+++ b/gcp_variant_transforms/testing/integration/large_tests/platinum_with_merge.json
@@ -1,0 +1,23 @@
+{
+  "test_name": "platinum-with-merge",
+  "table_name": "platinum_with_merge",
+  "input_pattern": "gs://genomics-public-data/platinum-genomes/vcf/*.vcf",
+  "runner": "DataflowRunner",
+  "variant_merge_strategy": "MOVE_TO_CALLS",
+  "worker_machine_type": "n1-standard-16",
+  "max_num_workers": "10",
+  "assertion_configs": [
+    {
+      "query": ["NUM_ROWS_QUERY"],
+      "expected_result": {"num_rows": 309551691}
+    },
+    {
+      "query": ["SUM_START_QUERY"],
+      "expected_result": {"sum_start": 24107891293483855}
+    },
+    {
+      "query": ["SUM_END_QUERY"],
+      "expected_result": {"sum_end": 24107908471506508}
+    }
+  ]
+}
--- a/gcp_variant_transforms/testing/integration/large_tests/platinum_with_merge.json
+++ b/gcp_variant_transforms/testing/integration/large_tests/platinum_with_merge.json
@@ -5,19 +5,19 @@
   "runner": "DataflowRunner",
   "variant_merge_strategy": "MOVE_TO_CALLS",
   "worker_machine_type": "n1-standard-16",
-  "max_num_workers": "10",
+  "max_num_workers": "20",
   "assertion_configs": [
     {
       "query": ["NUM_ROWS_QUERY"],
-      "expected_result": {"num_rows": 309551691}
+      "expected_result": {"num_rows": 261285806}
     },
     {
       "query": ["SUM_START_QUERY"],
-      "expected_result": {"sum_start": 24107891293483855}
+      "expected_result": {"sum_start": 20441297971132809}
     },
     {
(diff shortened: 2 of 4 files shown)


gcp_variant_transforms/testing/integration/large_tests/platinum_no_merge.json
```diff
@@ -0,0 +1,22 @@
...
+  "worker_machine_type": "n1-standard-16",
+  "max_num_workers": "5",
+  "assertion_configs": [
+    {
+      "query": ["NUM_ROWS_QUERY"],
+      "expected_result": {"num_rows": 309551691}
+    },
+    {
+      "query": ["SUM_START_QUERY"],
+      "expected_result": {"sum_start": 24107891293483855}
```
arostamianfar: huh, I did not know BQ can actually handle such large sums! cool!

gcp_variant_transforms/testing/integration/large_tests/platinum_with_merge.json
```diff
@@ -0,0 +1,23 @@
...
+  "table_name": "platinum_with_merge",
+  "input_pattern": "gs://genomics-public-data/platinum-genomes/vcf/*.vcf",
+  "runner": "DataflowRunner",
+  "variant_merge_strategy": "MOVE_TO_CALLS",
+  "worker_machine_type": "n1-standard-16",
+  "max_num_workers": "10",
+  "assertion_configs": [
+    {
+      "query": ["NUM_ROWS_QUERY"],
+      "expected_result": {"num_rows": 309551691}
```
arostamianfar: Actually, just noticed that this should be updated as it's from no-merge option.

gcp_variant_transforms/testing/integration/large_tests/platinum_no_merge.json
bashir2: Indeed :-)

gcp_variant_transforms/testing/integration/large_tests/platinum_with_merge.json
bashir2: Right, forgot to update it and for this case it never reached table validation before. Thanks for pointing it out.

bashir2:
Update: This actually succeeded with 20*16 workers in ~26 minutes (this is without image building as I use --skip_build but even with that it should be sub 0.5 hour which I think is reasonable to be done once for every commit). I am rerunning to check if there is a flaky issue but it is quite possible that in the early morning hours I was too tired and I misinterpreted the error. If I cannot reproduce the Quota related issue in a few runs, I would like to commit this as is.

PTAL.

gcp_variant_transforms/testing/integration/large_tests/platinum_no_merge.json
```diff
@@ -0,0 +1,22 @@
+{
+  "test_name": "platinum-no-merge",
+  "table_name": "platinum_no_merge",
+  "input_pattern": "gs://genomics-public-data/platinum-genomes/vcf/*.vcf",
+  "runner": "DataflowRunner",
+  "worker_machine_type": "n1-standard-16",
+  "max_num_workers": "20",
```
arostamianfar: Actually, I wonder if we can get faster performance with `num_workers` instead of `max_num_workers` (autoscaling takes adds several minutes).

bashir2: The reason I chose `max_num_workers` is that using `num_workers` alone tends to produce more variability in terms of both run-time and resource usage. I guess part of the reason is that the actual number of workers can go higher (for example check [this run](https://pantheon.corp.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2018-03-08_16_10_46-8867730247156911233?project=gcp-variant-transforms-test) with `num_workers=20` but no `max_num_workers`).

I want this to be fairly consistent in terms of run-time such that performance metric graphs/alerts become more meaningful. So I ended up experimenting with setting both `max_num_workers` and `num_workers` to get both extra speed-up and fair consistency. The new config uses more resources (e.g., CPU) but that is expected because auto-scaling does a better job.

[A platinum merge run with the new config](https://pantheon.corp.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2018-03-08_17_02_28-6541197997600173644?project=gcp-variant-transforms-test) taking ~19 mins.
[A platinum merge run only with `max_num_workers`](https://pantheon.corp.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2018-03-08_11_40_55-10364243419571597795?project=gcp-variant-transforms-test) taking ~24 mins.

PTAL.

pgrosu: Hi Bashir (@bashir2) and Asha (@arostamianfar),

Though I'm sure you already know this, but even if you don't specify either `num_workers` or `max_num_workers` - but not both being specified - then the Dataflow service will actually set the unspecified one, as noted here (0 is the equivalent of unspecified for `num_workers`):

workers

The link to this part of the documentation is the following:

https://cloud.google.com/dataflow/pipelines/specifying-exec-params#setting-other-cloud-pipeline-options

You will need to inspect what are the `currentNumWorkers` for the job is with the `getCurrentNumWorkers()`:

https://developers.google.com/resources/api-libraries/documentation/dataflow/v1b3/java/latest/com/google/api/services/dataflow/model/AutoscalingEvent.html

Hope it helps,
Paul

bashir2: Thanks Paul (@pgrosu) for your comment. Yes, I think setting one of these and leaving the other unset will leave the choice to the autoscalling algorithm. One of the use cases for these large tests is to monitor the performance of the pipeline and catch regressions quickly. For that to work reliably, I am thinking that maybe we should even turn autoscalling off. But I think setting both of these to the same value has a similar effect.

arostamianfar: Interesting! I did not realize that just setting num_workers can actually use autoscale as well. Should probably update our [docs](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/docs/large_inputs.md) too.

bashir2: Yes we should.

pgrosu: Hi Bashir and Asha,

I totally agree auto-scaling must be off as that is zone-specific overhead that can be very unpredictable, which pre-allocation solves.  Regarding the Dataflow performance metric, why is it taking 19 minutes, which seems kind of long for 30 GB of data (compressed).  What does the [Dataflow Monitor analysis](https://cloud.google.com/dataflow/pipelines/dataflow-monitoring-intf) indicate as a bottleneck?  Are there [shuffling opportunities](https://cloud.google.com/blog/big-data/2017/06/introducing-cloud-dataflow-shuffle-for-up-to-5x-performance-improvement-in-data-analytic-pipelines) or [fusion possibilities](https://cloud.google.com/dataflow/service/dataflow-service-desc#fusion-optimization) for your execution graph?  Are the cores maximally utilized?  What I/O opportunities are overlooked?  With 320 cores (20\*16) and 1,200 GB RAM (20\*60) this should go much faster than 26 minutes.  Take a look at my analysis of 1000 Genomes with Dion here that could be done in seconds, especially since your transfer-rate is quite fast within a datacenter (zone):

https://github.com/googlegenomics/utils-java/issues/62#issuecomment-220444203

Regarding checking any regression cases, maybe utilizing the GATK test cases from here might help:

https://github.com/broadinstitute/gatk/tree/master/src/test

Others which I'm sure you know can be:

NCBI:  
* ftp://ftp.ncbi.nih.gov/snp/organisms/
* ftp://ftp.ncbi.nih.gov/snp/organisms/human_9606/VCF/

EBI: 
* ftp://ftp.ensembl.org/pub/release-91/variation/vcf/

Not sure if combining large-scale performance and regression is an optimal combination, since both can be significant rabbit-holes, and limiting the analysis to one variable while building up the pipeline might save you unexpected headaches.   Also probably some scale-up benchmarks for different zones during different times of the day might help users, but that could be quite a lot of work to perform.

Hope it helps,
Paul

arostamianfar: Thanks for mentioning the thread, Paul (@pgrosu). If I understand the reference pipeline correctly, I believe the main difference between the VT pipeline and the one you referenced is that we are not just reading data from VCF and are also performing several transformations like parsing all fields in each VCF record, merging variants across files, and writing to BQ. Just reading records from VCF (i.e. treating them as a 'text' file) is much faster and can be comparable to the pipeline you referenced (the Reads/Variants APIs already contained data in a parsed/merged form).
Having said that, we haven't yet enabled Cython (#110) and haven't really done any optimizations, so our code can definitely run faster.

pgrosu: Hi Asha (@arostamianfar), you might be surprised, but your pipeline can actually be fast even without Cython with branching/loop optimizations.  Since you have so much memory, most of the data can just be mapped there even without an LRU optimization.  Dataflow will do a lot for you, but sometimes it needs guidance.  Parsing has equivalency to the reading text files, and merging(sort) is O(n*lg(n)) as a worst-case scenario.  So you still can come close to my scenario :)

arostamianfar: Interesting! We'll definitely consider this when considering optimizing the pipeline! For now, we've been mostly occupied on feature development and building foundational tasks like integration testing as majority of users are ok with the current performance, but want a robust pipeline.
Thanks so much for the feedback and comments though! Please do continue to engage with us as we're building VT. We really appreciate it!

pgrosu: Glad to help out :)

Issue #101, opened 2018-01-31
Add an option to 'detect and correct' VCF header mismatch errors


arostamianfar:
It is surprisingly common for VCF files to have missing header fields (see Issue #61 for a 1000 genomes file). The pipeline currently fails as it is unable to map the missing field(s) to the BigQuery column(s). Note that we need to generate the schema _prior_ to running the pipeline, which is why we need to rely on complete and valid headers.

The current workaround is to manually specify the missing header fields through `--representative_header_file`, but this is painful and not scalable.

An alternative approach is to do two passes on the data: the first one parses all of the fields in all records just to find the missing headers and the 2nd one actually processes the data. We'd use the data from the first pass to generate the BigQuery schema. Of course, this adds additional computation (roughly 30% more), so we should provide this as an optional feature for "robust" imports. As an optimization, we could provide a sampling logic if the data is expected to be uniform (e.g. sample 30% of the data and assume that everything else follows that schema).

nmousavi:
This is also done by --infer-undefined-header-fields flag set, right? The optimization part can be moved to a new issue if at all (should be lower priority though).

Reopen if there is anything else left.

Issue #79, opened 2018-01-23
Catch header schema mismatch errors earlier


arostamianfar:
Currently, if there is a header schema mismatch (e.g. a FORMAT field is missing from the header), the pipeline will proceed all the way until the very end and only fails when it tries to write to BigQuery. It also does not give a representative error message (complains that a field is missing in BQ JSON schema, see Issue #61). We should catch such errors sooner and give proper error. Depends on Issue #59 as we need to first refactor that part of the code.

Ideally we may be able to 'detect and correct' such errors, but this is more difficult currently as the BQ schema cannot be dynamically created/changed.

PR #136, opened 2018-03-08
Refactor conflict resolver class into its own file.

nmousavi:
Refactor _HeaderConflictResolver to a separate file and renamed it as
FieldConflictResolver. Why?
  - is going to grow soon (in next PRs)
  - deserve its own testing

Related to issue  #79 , #101 

Design doc: Design Doc: https://goo.gl/NXe27p

Tested:
  unit test
  presubmit

### DIFF
--- a/gcp_variant_transforms/transforms/merge_headers.py
+++ b/gcp_variant_transforms/transforms/merge_headers.py
@@ -15,81 +15,11 @@
 """beam combiner function for merging VCF file headers."""
 from collections import OrderedDict
 import apache_beam as beam
-import vcf
 
 from gcp_variant_transforms.beam_io import vcf_header_io
+from gcp_variant_transforms.transforms import vcf_field_conflict_resolver
 
 __all__ = ['MergeHeaders']
-
-class _HeaderConflictResolver(object):
-  """A class for resolving mistmatch between header fields (e.g. type, num).
-  """
-
-  def __init__(self, split_alternate_allele_info_fields=True):
-    """Initialize the class.
-
-    Args:
-     split_alternate_allele_info_fields (bool): Whether INFO fields with
-       `Number=A` are store under the alternate_bases record.
-     """
-    self._split_alternate_allele_info_fields = (
-        split_alternate_allele_info_fields)
-
-  def resolve(self, vcf_field_key, first_vcf_field_value,
-              second_vcf_field_value):
-    """Returns resolution for the conflicting field values.
-
-    Args:
-      vcf_field_key (str): field key in VCF header.
-      first_vcf_field_value (int or str): first field value.
-      second_vcf_field_value (int or str): second field value.
-    Raises:
-      ValueError: if the conflict cannot be resolved.
-    """
-    if vcf_field_key == 'type':
-      return self._resolve_type(first_vcf_field_value, second_vcf_field_value)
-    elif vcf_field_key == 'num':
-      return self._resolve_number(first_vcf_field_value, second_vcf_field_value)
-    else:
-      # We only care about conflicts in 'num' and 'type' fields.
-      return first_vcf_field_value
-
-  def _resolve_type(self, first, second):
-    if first == second:
-      return first
-    elif (first in ('Integer', 'Float') and second in ('Integer', 'Float')):
-      return 'Float'
-    else:
-      raise ValueError('Incompatible values cannot be resolved: '
-                       '{}, {}'.format(first, second))
-
-  def _resolve_number(self, first, second):
-    if first == second:
-      return first
-    elif (self._is_bigquery_field_repeated(first) and
-          self._is_bigquery_field_repeated(second)):
-      # None implies arbitrary number of values.
-      return None
-    else:
-      raise ValueError('Incompatible numbers cannot be resolved: '
-                       '{}, {}'.format(first, second))
-
-  def _is_bigquery_field_repeated(self, vcf_num):
-    """Returns true if the corresponding field in bigquery schema is repeated.
-
-    Args:
-      vcf_num (int): value of field `Number` in VCF header.
-    """
-    if vcf_num in (0, 1):
-      return False
-    elif (vcf_num == vcf.parser.field_counts['A'] and
-          self._split_alternate_allele_info_fields):
-      # info field with `Number=A` does not become a repeated field if flag
-      # `split_alternate_allele_info_fields` is on.
-      # See `variant_transform_options.py` for more details.
-      return False
-    else:
-      return True
 
 
 class _HeaderMerger(object):
@@ -99,8 +29,8 @@ class _HeaderMerger(object):
     """Initialize :class:`VcfHeader` object.
 
     Args:
-      resolver (:class:`_HeaderConflictResolver`): for resolving possible
-        header value mistmatches.
+      resolver (:class:`vcf_field_cnflict_resolver.FieldConflictResolver``):
+        for resolving possible header value mistmatches.
     """
     self._resolver = resolver
 
@@ -197,7 +127,8 @@ class MergeHeaders(beam.PTransform):
     """
     super(MergeHeaders, self).__init__()
     self._header_merger = _HeaderMerger(
-        _HeaderConflictResolver(split_alternate_allele_info_fields))
+        vcf_field_conflict_resolver.FieldConflictResolver(
+            split_alternate_allele_info_fields))
 
   def expand(self, pcoll):
     return pcoll | 'MergeHeaders' >> beam.CombineGlobally(_MergeHeadersFn(
--- a/gcp_variant_transforms/transforms/merge_headers.py
+++ b/gcp_variant_transforms/transforms/merge_headers.py
@@ -57,8 +57,8 @@ class _HeaderMerger(object):
     self._merge_header_fields(first.contigs, second.contigs)
 
   def _merge_header_fields(self, first, second):
-    #type: (Dict[str, OrderedDict[str, Any]],
-    #       Dict[str, OrderedDict[str, Any]]) -> None
+    #type: (Dict[str, OrderedDict[str, Union[str, int]]],
+    #       Dict[str, OrderedDict[str, Union[str, int]]]) -> None
     """Modifies ``first`` to add any keys from ``second`` not in ``first``.
 
     Args:
--- a/gcp_variant_transforms/transforms/merge_headers.py
+++ b/gcp_variant_transforms/transforms/merge_headers.py
@@ -18,7 +18,7 @@ import apache_beam as beam
 from typing import Dict, Any  #pylint: disable=unused-import
 
 from gcp_variant_transforms.beam_io import vcf_header_io
-from gcp_variant_transforms.transforms import vcf_field_conflict_resolver
+from gcp_variant_transforms.libs import vcf_field_conflict_resolver
 
 # TODO(nmousavi): Consider moving this into a separate file.
 class _HeaderMerger(object):
--- a/gcp_variant_transforms/transforms/merge_headers.py
+++ b/gcp_variant_transforms/transforms/merge_headers.py
@@ -15,22 +15,21 @@
 """beam combiner function for merging VCF file headers."""
 from collections import OrderedDict
 import apache_beam as beam
+from typing import Dict, Any  #pylint: disable=unused-import
 
 from gcp_variant_transforms.beam_io import vcf_header_io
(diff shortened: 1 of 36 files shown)


gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
```diff
@@ -0,0 +1,90 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import vcf
+
+__all__ = ['FieldConflictResolver']
+
+class FieldConflictResolver(object):
+  """A class for resolving all VCF field related mistmatches."""
```
bashir2: Is this functionality only going to be used for header merging? If yes, dose it help to reference "header" somewhere in the comment or even more preferably in the name of the class? Or maybe even the name of the module, should have "header" in it?

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
```diff
@@ -0,0 +1,90 @@
...
+
+  def __init__(self, split_alternate_allele_info_fields=True):
+    # type: (bool) -> None
+    """Initialize the class.
+
+    Args:
+     split_alternate_allele_info_fields (bool): Whether INFO fields with
+       `Number=A` are store under the alternate_bases record.
```
bashir2: nit: store -> stored

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
```
bashir2: I think it makes more sense to move this out of the transforms directory to "libs". There is nothing Beam specific in this module (which is a good thing), is it?

gcp_variant_transforms/transforms/merge_headers_test.py
```diff
@@ -305,8 +306,9 @@ def test_combine_pipeline(self):
     headers_1 = self._get_header_from_reader(vcf_reader_1)
     headers_2 = self._get_header_from_reader(vcf_reader_2)
 
-    header_merger = HeaderMerger(HeaderConflictResolver(
-        split_alternate_allele_info_fields=True))
```
bashir2: Out of curiosity, do you know why all the tests in this file except this one, directly use the the methods of the `_MergeHeadersFn` instead of using TestPipeline? It seems to me that tests for a PTransform should preferably test its public interface not how it is implemented. I have not checked the details and I think you are not the original author of this but I thought maybe you know.

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
```diff
@@ -0,0 +1,90 @@
...
+  def resolve(self, vcf_field_key, first_vcf_field_value,
+              second_vcf_field_value):
+    # type: (str, Union[str, int], Union[str, int]) -> Union[str, int]
+
+    """Returns resolution for the conflicting field values.
+
+    Args:
+      vcf_field_key: field key in VCF header.
+      first_vcf_field_value: first field value.
+      second_vcf_field_value: second field value.
```
bashir2: nit: Returns?

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
```diff
@@ -0,0 +1,90 @@
...
+    elif vcf_field_key == 'num':
+      return self._resolve_number(first_vcf_field_value, second_vcf_field_value)
+    else:
+      # We only care about conflicts in 'num' and 'type' fields.
+      return first_vcf_field_value
+
+  def _resolve_type(self, first, second):
+    if first == second:
+      return first
+    elif (first in ('Integer', 'Float') and second in ('Integer', 'Float')):
```
bashir2: Consider making 'Integer' and 'Float' constants.

gcp_variant_transforms/transforms/vcf_field_conflict_resolver_test.py
```diff
@@ -0,0 +1,68 @@
...
+  def test_resolving_field_definition_conflict_in_type(self):
+    self.assertEqual(
+        self._resolver.resolve('type', 'Integer', 'Float'), 'Float')
+    with self.assertRaises(ValueError):
+      self._resolver.resolve('type', 'Integer', 'String')
+      self.fail('Should raise exception for unresolvable types')
+
+  def test_resolving_field_definition_conflict_in_number(self):
+    self.assertEqual(
+        self._resolver.resolve('num', 2, 3), None)
```
bashir2: Maybe 'type' and 'num' should be constants too (and used in test)?

gcp_variant_transforms/transforms/merge_headers.py
```diff
@@ -99,8 +29,8 @@ def __init__(self, resolver):
     """Initialize :class:`VcfHeader` object.
 
     Args:
-      resolver (:class:`_HeaderConflictResolver`): for resolving possible
-        header value mistmatches.
+      resolver (:class:`vcf_field_cnflict_resolver.FieldConflictResolver``):
```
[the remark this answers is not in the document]
nmousavi: Done!

gcp_variant_transforms/transforms/merge_headers.py
```diff
@@ -99,8 +29,8 @@ def __init__(self, resolver):
...
+        for resolving possible header value mistmatches.
```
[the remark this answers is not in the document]
nmousavi: Done!

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
```diff
@@ -0,0 +1,90 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import vcf
+
+__all__ = ['FieldConflictResolver']
```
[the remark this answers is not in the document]
nmousavi: Done!

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
nmousavi: It's not specific to header, and is going to be used for resolving conflicts between schema and field_data as well (in my next PR).

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
```diff
@@ -0,0 +1,90 @@
...
+
+class FieldConflictResolver(object):
+  """A class for resolving all VCF field related mistmatches."""
+
+  def __init__(self, split_alternate_allele_info_fields=True):
+    # type: (bool) -> None
+    """Initialize the class.
+
+    Args:
+     split_alternate_allele_info_fields (bool): Whether INFO fields with
```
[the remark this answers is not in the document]
nmousavi: Done!

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
nmousavi: Done!

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
```diff
@@ -0,0 +1,90 @@
...
+       `Number=A` are store under the alternate_bases record.
+     """
+    self._split_alternate_allele_info_fields = (
+        split_alternate_allele_info_fields)
+
+  def resolve(self, vcf_field_key, first_vcf_field_value,
+              second_vcf_field_value):
+    # type: (str, Union[str, int], Union[str, int]) -> Union[str, int]
+
```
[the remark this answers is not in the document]
nmousavi: Removed!

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
nmousavi: Return is mentioned in one-line desc. Do you want it to be repeated?

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
nmousavi: Done!

gcp_variant_transforms/transforms/vcf_field_conflict_resolver_test.py
nmousavi: Done!

gcp_variant_transforms/transforms/merge_headers_test.py
nmousavi: I don't know :-).

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
nmousavi: Done!

gcp_variant_transforms/transforms/merge_headers.py
```diff
@@ -120,14 +49,15 @@ def merge(self, first, second):
...
-
     self._merge_header_fields(first.infos, second.infos)
     self._merge_header_fields(first.filters, second.filters)
     self._merge_header_fields(first.alts, second.alts)
     self._merge_header_fields(first.formats, second.formats)
     self._merge_header_fields(first.contigs, second.contigs)
 
   def _merge_header_fields(self, first, second):
+    #type: (Dict[str, OrderedDict[str, Any]],
+    #       Dict[str, OrderedDict[str, Any]]) -> None
```
bashir2: This is a great example of how adding types actually makes reading the code much easier; I just experienced the difference looking at the old and the new versions; thanks for gradually adding types to the code.

On that note: Is the type of the `OrderedDict` values `Any` or `Union[str, int]`? Because it seems to me that `resolve()` expects `Union[str, int]`.

gcp_variant_transforms/transforms/vcf_field_conflict_resolver.py
bashir2: Hmm, okay then I think this warrants adding more documentation either at module level or here, maybe with some examples of how you expect this to be used in the Header and non-Header cases. It was not obvious to me that this is not header specific because for example how you compare keys, types, etc. which seemed header specific.

I am sure in your next PRs this become more clear, so feel free to add the documentation there (but please add a TOOD now not to forget).

gcp_variant_transforms/transforms/merge_headers.py
nmousavi: I think so, but not sure as it depends on what parser sends to us. But I can confirm for the attributes we care/resolve here (i.e type and num) it's either str or int. Do you want me to change it to Union[str, int]?

gcp_variant_transforms/transforms/merge_headers_test.py
nmousavi: Added a TODO to unify tests. 

@arostamianfar : let us know if there is any reason, they should stay as they are.

gcp_variant_transforms/transforms/merge_headers.py
bashir2: If we were sure they are always `str` or `int` then yes I would have preferred that. If we are not sure, then I think we should fix the type for `FieldConflictResolver.resolve` function to be `Any` (instead of `Union[str, int]`) because you are sending these values to that function. Either choice is fine with me.

So this is an issue which actually should be picked by tools. For example in IntelliJ, it will actually give us warning if you submit the code in its current form. It is on me to add a tool for doing type checking as part of presubmit which I haven't found cycles to work on yet, sorry. We should not have this discussion in a review thread.

gcp_variant_transforms/transforms/merge_headers_test.py
arostamianfar: I agree with Bashir. Thanks for adding the TODO!

gcp_variant_transforms/transforms/merge_headers.py
nmousavi: Added Union[str, int] as we discussed offline. Confirmed PyVcf always return str or int. If this changes later on (using another parser), we don't care (doesn't hurt anyway).

PR #141, opened 2018-03-16
Adds the option for processing ALLELE_NUM.

bashir2:
Check the documentation of [--allele_number](https://useast.ensembl.org/info/docs/tools/vep/script/vep_options.html#opt_allele_number) option of VEP.

Tested:
Unit tests added; Also ran the pipeline on a sample of gnomAD with --use_allele_num

Issue #81


bashir2:
BTW, stating the obvious: This PR is on top of PR #131 so the first commit is actually that PR. To see changes for this PR only, please check the [second commit](https://github.com/googlegenomics/gcp-variant-transforms/pull/141/commits/a0a66c863ece44ba08107f71741d1dace27ccafa).

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -599,6 +624,27 @@ def _alt_matches_annotation_alt_minimal_mode(
       return True
     return False
 
+  def _add_annotations_by_allele_num(
+      self, proc_var, annotation_dict, annotation_field_name):
+    # type: (ProcessedVariant, Dict[str, str], str) -> None
+    if _ALLELE_NUM_ANNOTATION not in annotation_dict:
```
arostamianfar: can we always use this check so that we no longer need the use_allele_num flag? (i.e. the logic would be to first look for allele_num, if it doesn't exist, then use minimal or exact match depending on the --minimal flag). Or are you concerned about cases where allele_num is incorrect?

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -599,6 +624,27 @@ def _alt_matches_annotation_alt_minimal_mode(
...
+      self._allele_num_missing_counter.inc()
+      return
+    index_str = annotation_dict[_ALLELE_NUM_ANNOTATION]
+    try:
+      alt_index = int(index_str) - 1
+      alt_list = proc_var._alternate_datas
```
arostamianfar: consider using the public getter?

bashir2:
BTW, thanks for the quick review.

gcp_variant_transforms/libs/processed_variant.py
arostamianfar: I see. Sounds good. Let's keep the flag then.

PR #149, opened 2018-03-22
Add logic in resolving incompatible conflicts in the resolver

allieychen:
- Add a flag `resolve_all` for the `FieldConflictResolver`, which resolves all incompatible conflicts when setting to true.
- Add two unit tests.
- Correct some warnings.

Tested: unit test


gcp_variant_transforms/libs/vcf_field_conflict_resolver.py
```diff
@@ -14,27 +14,38 @@
...
 class FieldConflictResolver(object):
-  """A class for resolving all VCF field related mistmatches."""
+  """A class for resolving all VCF field related mismatches."""
   # TODO(nmousavi): explain this class more.
 
-  def __init__(self, split_alternate_allele_info_fields=True):
-    # type: (bool) -> None
+  def __init__(self,
+               split_alternate_allele_info_fields=True,
+               resolve_all=False):
```
nmousavi: I would suggest 'resolve_always', WDYT?

gcp_variant_transforms/libs/vcf_field_conflict_resolver.py
```diff
@@ -14,27 +14,38 @@
...
     """Initialize the class.
 
     Args:
-     split_alternate_allele_info_fields: Whether INFO fields with
-       `Number=A` are stored under the alternate_bases record.
-     """
+      split_alternate_allele_info_fields: Whether INFO fields with `Number=A`
+        are stored under the alternate_bases record.
+      resolve_all: Resolves all incompatible conflicts:
+        any incompatible type conflict is resolved as type = `String`
```
nmousavi: This is not true, if the conflict is between Integer and Float, we resolve it to Float. 
Maybe say that if this flag set, we always find a resolution, i.e., as a last resort we convert all type conflicts to String and Number to '.'?

gcp_variant_transforms/libs/vcf_field_conflict_resolver_test.py
```diff
@@ -28,6 +28,9 @@ def setUp(self):
     self._resolver = vcf_field_conflict_resolver.FieldConflictResolver()
     self._resolver_allele = vcf_field_conflict_resolver.FieldConflictResolver(
         split_alternate_allele_info_fields=True)
+    self._resolver_all = vcf_field_conflict_resolver.FieldConflictResolver(
```
nmousavi: maybe s/resolver_all/resolver_always/?

gcp_variant_transforms/libs/vcf_field_conflict_resolver_test.py
```diff
@@ -68,6 +71,31 @@ def test_resolving_field_definition_conflict_in_number_allele(self):
     for i in [self._field_count('A')]:
       for j in [self._field_count('R'), self._field_count('G'), 0, 1, 2, None]:
         with self.assertRaises(ValueError):
-          self._resolver.resolve(VcfParserConstants.NUM, i, j)
+          self._resolver_allele.resolve(VcfParserConstants.NUM, i, j)
```
nmousavi: Nice catch! Thanks!

gcp_variant_transforms/libs/vcf_field_conflict_resolver_test.py
```diff
@@ -28,6 +28,9 @@ def setUp(self):
     self._resolver = vcf_field_conflict_resolver.FieldConflictResolver()
     self._resolver_allele = vcf_field_conflict_resolver.FieldConflictResolver(
         split_alternate_allele_info_fields=True)
+    self._resolver_all = vcf_field_conflict_resolver.FieldConflictResolver(
+        resolve_all=True
```
arostamianfar: nit: please put closing bracket on the same line

gcp_variant_transforms/libs/vcf_field_conflict_resolver.py
allieychen: Thanks! Here I mean all incompatible conflicts while Integer vs. Float are compatible. I realized it was not clear what are incompatible conflicts. I will change based on your comments :)

gcp_variant_transforms/libs/vcf_field_conflict_resolver_test.py
allieychen: Right! Thanks.

gcp_variant_transforms/libs/vcf_field_conflict_resolver.py
allieychen: Good advice! Thanks.

gcp_variant_transforms/libs/vcf_field_conflict_resolver_test.py
allieychen: Done.

allieychen:
PTAL

Issue #151, opened 2018-03-23
Fix integration testing failures


arostamianfar:
Our integration testing started failing since PR #148 with error [1]. I believe it may be b/c the test pushes the runtime above 30 minutes, which causes the oauth token to get expired.

Solutions:
- (short term): Reduce the # of files in this test to be 10k (i.e. just under "valid-10000-copies-1/"). This should push the test to run under 20mins, and we can verify whether it fixes the issue.
- Investigate how to either increase the token expiration time and/or catch the exception.

[1]
```
google.auth.exceptions.RefreshError: ('Failed to retrieve http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/542668135037@cloudbuild.gserviceaccount.com/?recursive=true from the Google Compute Enginemetadata service. Status: 404 Response:\n404 page not found\n', <google.auth.transport.requests._Response object at 0x7f7504072c90>)
```

pgrosu:
Hi Asha,

Why not try to integrate with the internal Jenkins (Kororo) as the [protobuf transitioned to it](https://github.com/google/protobuf/tree/master/kokoro) in order to get around this and other resource problems?  It has github integration to trigger automatic testing for new PRs or updated commits.

~p

arostamianfar:
Hi Paul,

I think Kokoro is great for cases where is an equivalent source under google internal repo. Given that we build everything in open source, then that is not an option for us (unless we add a sync from github to internal repo). Other than this strange issue, I actually think our existing integration testing framework is working very well.

We are actually discussing if we want to make the continuous runs smaller (<20mins) to avoid this bug and essentially migrate some tests to "huge" so that they are only run for release candidates.

pgrosu:
Hi Asha,

I didn't realize you needed to duplicate the code with Kororo, though that could be automated.  In any case, if it's just an issue with refreshing the token, that is pretty easy.  I'm not sure what the programming language you use, but see the following previous posts/docs for how to use `setAccessToken()` and `setRefreshToken()`:

https://github.com/googlegenomics/utils-java/issues/8

https://developers.google.com/identity/protocols/OAuth2WebServer#offline

Regarding the timeout I thought you could specify a [`travis_wait` amount (in minutes)](https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out).  When you mean "huge" are you saying that you have a customized build environment, that is different that the following standard ones:

[...]

PR #148, opened 2018-03-20
Add integration test for option optimize for large inputs #106

allieychen:
- Create a folder in `gcp-variant-transforms-testfiles/large_tests` with 70000 copies of file `valid_4_2` to test `LARGE_DATA_THRESHOLD = 50000` pipeline.
- Option `--optimize_for_large_inputs` is included.

Tested: Manually ran deploy_and_run_tests.sh --run_presubmit_tests
Issue: #106

Issue #106, opened 2018-02-05
Add at least one integration test for each option


arostamianfar:
Now that we have the continuous integration testing working, we should add at least one integration test to cover each option: https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/options/variant_transform_options.py

PR #152, opened 2018-03-23
Revert "Add integration test for option optimize for large inputs #106" #151

allieychen:
Reverts googlegenomics/gcp-variant-transforms#148
- This pull request cause issue #151. 
- The reason is not clear yet, but a quick test shows reverting the code will solve issue #151. Further investigation will be performed.


allieychen:
I created a new trigger and tried in my only branch. I only ran for one time and it works.

PR #153, opened 2018-03-26
Add docs for loading multiple files and troubleshooting.

arostamianfar:
Preview: https://github.com/arostamianfar/gcp-variant-transforms/tree/docs2

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -36,7 +36,12 @@ See [handling large inputs](./large_inputs.md) for more details.
 Try increasing the disk size allocated to each worker by specifying
 `--disk_size_gb <disk_size>` and/or increasing the number of workers by
 specifying `--max_num_workers <num_workers>`. By default, each worker gets 250GB
-of disk, which may not be enough to read the entire dataset.
+of disk, and the aggregate disk size available to all workers should be at least
+as large as the uncompressed size of the VCF files being loaded. However, to
+accomoddate for intermediate stages of the pipeline and also to account for
+the additional overhead introduced by the transforms, the aggregate disk size
+among all workers should be at least 3 to 4 times the total size of raw VCF
+files.
 
 See [handling large inputs](./large_inputs.md) for more details.
 
@@ -75,7 +80,7 @@ You can fix this by:
   header file. However, running with this option adds ~30% more compute to the
   pipeline.
 
-# Error: "BigQuery schema has no such field"
+## Error: "BigQuery schema has no such field"
 
 Same as [above](#error-json-parsing-error--no-such-field-field_name).
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,0 +1,81 @@
+# Troubleshooting
+
+This page summarizes common error scenarios when running the pipeline and
+provide recommanded workarounds. If you are still unable to successfully load
+your VCF files, please post on the
+[google-genomics-discuss](https://groups.google.com/forum/#!forum/google-genomics-discuss)
+group or file a GitHub issue if you believe that there is a bug in the pipeline.
+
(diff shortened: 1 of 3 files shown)


docs/troubleshooting.md
```diff
@@ -0,0 +1,81 @@
...
+  a separate process).
+
+See [handling large inputs](./large_inputs.md) for more details.
+
+## Pipeline crashes due to out of disk error
+
+Try increasing the disk size allocated to each worker by specifying
+`--disk_size_gb <disk_size>` and/or increasing the number of workers by
+specifying `--max_num_workers <num_workers>`. By default, each worker gets 250GB
+of disk, which may not be enough to read the entire dataset.
```
nmousavi: I can read it such that it implies each worker reads the entire dataset.

docs/troubleshooting.md
```diff
@@ -0,0 +1,81 @@
...
+
+See [handling large inputs](./large_inputs.md) for more details.
+
+## Error: "JSON parsing error ... No such field: <field_name>"
```
nmousavi: This error will never happen after PR#146.

docs/troubleshooting.md
```diff
@@ -0,0 +1,81 @@
...
+  of strings), which will match any field. Please also check whether the
+  missing field is for `##INFO` or `##FORMAT`.
+
+* If changing the file(s) is not an option, you can also run the pipeline with
+  `--representative_header_file <file_path>`, where you provide a merged view
+  of all headers in all files. You can add any missing fields to that file.
+  We are working on a tool to make this process easier and provide
+  recommandations for missing fields.
+
+* Coming soon: Run the pipeline with `--infer_undefined_headers`. This will do
```
nmousavi: Why coming soon? This has been added since long time ago. :-)

docs/troubleshooting.md
```diff
@@ -0,0 +1,81 @@
...
+  two passes on the data and will infer definition for undefined headers. You
+  do not need to make any changes to the VCF files or provide a representative
+  header file. However, running with this option adds ~30% more compute to the
+  pipeline.
+
+# Error: "BigQuery schema has no such field"
```
nmousavi: PR#146 will introduce another error if you want to add it now

"Value and schema do not match for field"

docs/troubleshooting.md
arostamianfar: Good point. Updated.

docs/troubleshooting.md
arostamianfar: Yeah, I think we can update the docs after the new docker image is released (it could still happen if somehow there is a bug, so I prefer to leave it, but swap the order with the one below)

docs/troubleshooting.md
arostamianfar: Right, but the docker image hasn't been updated and it's not yet tested well (there are no integration tests for it). Also, I thought that it wouldn't work without your recent PR to covert list and single-value numbers?

docs/troubleshooting.md
arostamianfar: I was hoping you could add these and resolve all the "coming soon"s once the new docker image is released with your changes :)

arostamianfar:
PTAL

docs/troubleshooting.md
nmousavi: Docker image contains this flag (built on Feb 27, and PR#114 submitted on Feb 23).
Infer header logic works, but it wont be enough if the value provided for the field is empty (and no = sign is present) because PyVcf extracts the data as Flag which cannot be casted to string(this case is handled in PR#146).

arostamianfar: yeap, so I think it's best if we wait for the new release and update this doc once we have added some integration testing.

PR #157, opened 2018-03-27
Use basestring instead of str for bigquery str.

nmousavi:
VT converts strings into unicode when inserting to Bigquery for field of
type string.

Testing:
  directrunner


nmousavi:
PTAL!

PR #161, opened 2018-03-28
Make pyvcf to use cython

samanvp:
Tested: I made sure the tests were failing before removing those 5 lines from vcfio_test.py


docker/Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN pip install --upgrade pip && pip install --upgrade virtualenv && \
     virtualenv /opt/gcp_variant_transforms/venv && \
     . /opt/gcp_variant_transforms/venv/bin/activate && \
     cd /opt/gcp_variant_transforms/src && \
-    pip install --upgrade .
+    pip install --upgrade --no-binary pyvcf,apache-beam[gcp] .
```
nmousavi: why apache-beam[gcp] is required? This was deleted recently.

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -199,11 +199,6 @@ def _get_invalid_file_contents(self):
             '#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	Sample\n',
             '1    1  '
         ],
-        # Missing "GT:GQ" format, but GQ is provided.
```
nmousavi: Why is it removed? You mentioned this makes test fail. But do you know why? How is it related to cython?

docker/Dockerfile
samanvp: This is for making sure these two lib are not installed from binary (so if it's possible use cython).
I saw Asha had a change today related to the new version of apache-beam, but I am not really sure if this change still is needed or not.
@arostamianfar what do your think?

gcp_variant_transforms/beam_io/vcfio_test.py
samanvp: This test is failing when pyvcf uses cython, this is a known issue:
https://github.com/apache/beam/pull/4221

We have to remove it otherwise test does not pass.

samanvp:
On a local run I was getting down from ~4 min to ~3 minutes, but I couldn't get any consistent result when I set the DataflowRunner flag.

docker/Dockerfile
nmousavi: I don't see how `--no-binary` is relevant to cython. If you need to specify this, I think there is a bug in how these packages are distributed.

Their configs look alright after all:
https://github.com/apache/beam/blob/master/sdks/python/MANIFEST.in
https://github.com/jamescasbon/PyVCF/blob/master/MANIFEST.in

gcp_variant_transforms/beam_io/vcfio_test.py
nmousavi: Ah, this is really bad. 
On the plus side, C PyVcf is more robust to malformed input. :-)

nmousavi:
I see. Do you know whether the improvement was due to C PyVcf or beam (or both)?

docker/Dockerfile
samanvp: --no-binary causes pip not to use binary packages and instead compile them from their source. 

In the case of pyvcf we definitely know --no-binary makes a difference; it causes the test case fails which indicates pyvcf is using cython.

arostamianfar: You no longer need to specify apache-beam[gcp] as pre-compiled wheel packages are available as of 2.4.0 (e.g. if you look at what is installed in docker, it is now "apache_beam-2.4.0-cp27-cp27mu-manylinux1_x86_64.whl"). You should keep the pyvcf one though :)

nmousavi: Pyc files are included in both cases (building either from source or bin). You can see them in `lib/python2.7/site-packages/vcf/`. And setuptools should install cython version in anyway as you have installed cython already. Therefore, It's not clear why installing pyvcf from bin doesn't work.

nmousavi: If you could figure out why it's not working, great! if not, I think we can keep this as is. LGTM!

PR #169, opened 2018-04-04
Add release documentation.

### LINKED ISSUE #77: Add proper release process and versioning
Ensure both github and docker images have proper release process (includes running integration tests) and are versioned appropriately.

arostamianfar:
Initial instructions for doing releases. Notes:
* This can be simplified a lot and many things can be automated.
* This takes a very paranoid view! We can relax some of the extra testing mentioned here later.

Preview: https://github.com/arostamianfar/gcp-variant-transforms/blob/rdoc/docs/release.md

Fixes #77


docs/release.md
```diff
@@ -0,0 +1,98 @@
...
+   the following command to make the next steps easier:
+   `export COMMIT_SHA=<commit sha>`.
+1. Wait until the
+   [Cloud Build](https://console.cloud.google.com/gcr/builds?project=gcp-variant-transforms-test)
+   process builds a new image that contains the change submitted in the previous
+   step. The image has the form
+   `gcr.io/gcp-variant-transforms-test/gcp-variant-transforms:${COMMIT_SHA}`.
+1. The tests in the previous step only cover presubmit tests. We require the
+   release image to pass all tests including `huge_tests`. Run the following on
+   your workstation in a **clean** branch synced to the version change commit
```
nmousavi: How do you do that? Could you add cmd for this too. :-)

docs/release.md
```diff
@@ -0,0 +1,98 @@
...
+1. Navigate to the
+   [releases tab](https://github.com/googlegenomics/gcp-variant-transforms/releases)
+   and click on **Draft a new release**.
+   1. Set the **Tag version** to `vX.X.X`
+      where `X.X.X` is the release version specified previously.
+   1. Set the **Target** to the commit that changed the version number.
+   1. Set the **Release title** to "Release vX.X.X".
+   1. Enter a description of the main features and/or bug fixes going into the
+      release.
+   1. Click on **Publish release**.
```
nmousavi: Those asset files are generated automatically, nothing to do manually, right? .

docs/release.md
arostamianfar: Done :)

docs/release.md
arostamianfar: Correct.

docs/release.md
```diff
@@ -0,0 +1,105 @@
...
+   a long string like `4ff1fdfb3957a115a80fe9343306952cd98b6e5c`). You may run
+   the following command to make the next steps easier:
+   `export COMMIT_SHA=<commit sha>`.
+1. Wait until the
+   [Cloud Build](https://console.cloud.google.com/gcr/builds?project=gcp-variant-transforms-test)
+   process builds a new image that contains the change submitted in the previous
+   step. The image has the form
+   `gcr.io/gcp-variant-transforms-test/gcp-variant-transforms:${COMMIT_SHA}`.
+1. The tests in the previous step only cover presubmit tests. We require the
+   release image to pass all tests including `huge_tests`. Run the following on
```
bashir2: I think you can automate this step by creating a new build trigger that is sensitive to tags (instead of a branch) and uses a cloudbuild config that includes `huge_tests` too. So basically after the setup.py change, we create a release tag on the git repo which triggers the above new build trigger (and minimizes possibility of mistakes in manual steps like the one below). For now please add a TODO to replace this step.

docs/release.md
```diff
@@ -0,0 +1,105 @@
...
+     ```bash
+     gcloud auth configure-docker
+     SOURCE_IMAGE=gcr.io/gcp-variant-transforms-test/gcp-variant-transforms:"${COMMIT_SHA}"
+     TARGET_IMAGE=gcr.io/gcp-variant-transforms/gcp-variant-transforms:"${COMMIT_SHA}"
+     docker pull "${SOURCE_IMAGE}"
+     docker tag "${SOURCE_IMAGE}" "${TARGET_IMAGE}"
+     docker push "${TARGET_IMAGE}"
+     ```
+
+   * Push a new image from source by running:
```
bashir2: Why do you even document this option? Is it because the previous step relies on `docker` being installed? IMO only an image that has been fully tested should be released (I understand the next step reproduces the same image but it is technically a new image).

docs/release.md
```diff
@@ -0,0 +1,105 @@
...
+         --substitutions _CUSTOM_TAG_NAME="${COMMIT_SHA}" .
+     ```
+1. Once the new image is published, navigate to the
+   [Container registry](https://console.cloud.google.com/gcr/images/gcp-variant-transforms/GLOBAL/gcp-variant-transforms?project=gcp-variant-transforms)
+   page and update the published image to have the labels `latest` and `X.X.X`
+   for the version number. You may optionally run a "sanity" test on the newly
+   released image by running:
+
+   ```bash
+   # TODO(arostami): modify deploy_and_run_tests.sh to use custom image.
```
bashir2: Hmm, it already does; basically you can use `--skip_build --image_tag TAG_NAME --keep_image`. Or did I misunderstood the goal of this TODO?

docs/release.md
```diff
@@ -0,0 +1,105 @@
...
+       --image gcr.io/gcp-variant-transforms/gcp-variant-transforms
+   ```
+1. Navigate to the
+   [releases tab](https://github.com/googlegenomics/gcp-variant-transforms/releases)
+   and click on **Draft a new release**.
+   1. Set the **Tag version** to `vX.X.X`
+      where `X.X.X` is the release version specified previously.
+   1. Set the **Target** to the commit that changed the version number.
+   1. Set the **Release title** to "Release vX.X.X".
+   1. Enter a description of the main features and/or bug fixes going into the
```
bashir2: Out of curiosity: Is there any version control for this description too? And is it only possible to make changes to it through the GitHub UI, or is possible to create/modify it with a PR too?

docs/release.md
bashir2: BTW, it probably makes more sense to run the integration test on the actual released image under the main project (i.e., `gcp-variant-transforms`) and not the test project. You can use the `--project` option of `deploy_and_run_tests.sh` but make sure you add [`--gs_dir [BUCKET]`](https://github.com/googlegenomics/gcp-variant-transforms/blob/55094c830fb66964885cf32340fb46fad6f55b16/deploy_and_run_tests.sh#L63) as well.

docs/release.md
arostamianfar: That's an interesting suggestion! I had not considered the build trigger on tags :) I have actually been thinking that we can automated most of these steps by creating a new `release.sh` script. For now, I added a TODO, but we can do it later. Filed #170 for this.

docs/release.md
arostamianfar: Good point. Removed :) And yes, I wanted to provide an option for a user who does't have docker installed, but I think it's ok to require it.

docs/release.md
arostamianfar: Correct, but I specifically do not want to run any jobs in the gcp-variant-transforms project. I think we should keep it as clean as possible and it should only store production docker images (i'm worried that once we allow jobs to be run there, people will start experimenting with it).

docs/release.md
arostamianfar: I'm not aware of any version control for release descriptions.

docs/release.md
```diff
@@ -40,38 +40,22 @@ The process is as follows:
        --keep_image --skip_build --run_all_tests \
        --project gcp-variant-transforms-test
    ```
+   TODO(arostami): Automate this step through cloudbuild!
```
bashir2: nit: arostamianfar

docs/release.md
bashir2: Then I think this TODO should not reference `deploy_and_run_tests.sh` (and BTW I agree that people should not use that script on `gcp-variant-transforms` project, I was just saying what this TODO says is already done). All you need is a small script that does the following steps. Maybe it can simply go to the `release.sh` script you mentioned earlier and simply drop this TODO.

arostamianfar: I see. Sounds good. I decided to remove the TODOs altogether (felt strange to have them in this doc) and instead just have Issue #170 as the "TODO".

PR #175, opened 2018-04-06
Add report module to generate the conflicts report. #178

allieychen:
- Add a module which combines the representative headers and the merged definitions to generate a report in .csv format.
- When no headers provided (for instance, the merge headers module fails), the report can still generate without providing a solution.
- The user can specify the directory.
- Unit tests are included.

Tested: unit tests. 
issue: #178


gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,116 @@
...
+
+  def __init__(self, directory=''):
+    # type: (str) -> None
+    self.directory = directory
+
+  def report_conflicts(self, header_definitions, representative_header=None):
+    # type: (VcfHeaderDefinitions, VcfHeader) -> None
+    """Generates a report.
+
+    Combines the conflicts extracted from ``header_definitions`` and the
```
nmousavi: nit: use single `

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,116 @@
...
+    for field_id, definitions_to_files_map in conflicts.iteritems():
+      row = [
+          field_id,
+          self._generate_conflicts(definitions_to_files_map),
+          self._generate_solution(headers, field_id)
+      ]
+      contents.append(row)
+    return contents
+
+  def _generate_conflicts(self, definition_to_files_map):
```
nmousavi: Since it just extracts definitions, I like `_extract_definitions` or `_extract_defintions_and_file_name` better

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,116 @@
...
+  _HEADER_LINE = ['ID', 'Conflicts', 'Proposed Solution']
+  _REPORT_NAME = 'header_conflicts_report.csv'
+  _NO_CONFLICTS_MESSAGE = ['No conflicts found.']
+  _NO_SOLUTION_MESSAGE = 'Not resolved'
+
+  def __init__(self, directory=''):
+    # type: (str) -> None
+    self.directory = directory
+
+  def report_conflicts(self, header_definitions, representative_header=None):
```
nmousavi: I like `resolved_header` better than `representative_header` because one with no context is more likely to guess what it is. WDYT?

gcp_variant_transforms/libs/generate_report_test.py
```diff
@@ -0,0 +1,236 @@
...
+    lines_2 = [
+        '##FORMAT=<ID=NS,Number=1,Type=Float,Description="Number samples">\n',
+        '#CHROM  POS ID  REF ALT QUAL  FILTER  INFO  FORMAT  Sample3\n'
+    ]
+    vcf_reader_1 = vcf.Reader(fsock=iter(lines_1))
+    vcf_reader_2 = vcf.Reader(fsock=iter(lines_2))
+    headers_1 = self._get_vcf_header_from_reader(vcf_reader_1, 'file1')
+    headers_2 = self._get_vcf_header_from_reader(vcf_reader_2, 'file2')
+
+    pipeline = TestPipeline()
```
nmousavi: This class should be tested in isolation. Please do not use pipeline for it. It adds dependency to other part of the pipeline which is unnecessary and can be headache to debug (e.g., imagine a scenario where your test fails because some other module used in the pipeline is buggy).

allieychen: Thanks for pointing this out! It is really helpful to test the individual functionality.

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,127 @@
...
+  def _write_to_report(self, contents):
+    # type: (List[str]) -> None
+    if self._directory and not os.path.exists(self._directory):
+      os.makedirs(self._directory)
+    file_path = os.path.join(self._directory, self._REPORT_NAME)
+    with open(file_path, 'w') as file_to_write:
+      writer = csv.writer(file_to_write)
+      if not contents:
+        writer.writerow(self._NO_CONFLICTS_MESSAGE)
+        return
```
arostamianfar: nit: I personally find using `else` to be more readable in this case as you only have one special case and the `else` block is short.

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,127 @@
...
+    conflict_definitions = []
+    for definition, file_names in definition_to_files_map.iteritems():
+      definition = self._format_definition(definition.num, definition.type)
+      conflict_definitions.append(definition + ' in ' + str(file_names))
+    return ' '.join(conflict_definitions)
+
+  def _extract_resolution(self, header, filed_id):
+    # type: (Dict[str, Dict[str, Union[str, int]]], str) -> str
+    if filed_id not in header:
+      return self._NO_SOLUTION_MESSAGE
```
arostamianfar: Please comment which scenarios this can happen (I suspect we'll always have a resolution unless no `resolved_header` is provided?).

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Combines the conflicts extracted from ``header_definitions`` and their
+    resolutions from ``resolved_header`` to generate a conflicts report.
+    """
+    resolved_header = resolved_header or VcfHeader()
+    format_conflicts = self._extract_conflicts(header_definitions.formats)
+    info_conflicts = self._extract_conflicts(header_definitions.infos)
+    format_headers = resolved_header.formats
+    info_headers = resolved_header.infos
+    contents = list()
```
arostamianfar: Prefer using `contents = []` as I think that's much more common in Google code.

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,127 @@
...
+      ):
+    # type: (...) -> Dict[str, Dict[Definition, List[str]]]
+    """Extracts the fields that have conflicted definitions."""
+    # len(v) > 1 means there are conflicted definitions for this field.
+    return dict([(k, v) for k, v in definitions.items() if len(v) > 1])
+
+  def _generate_contents(
+      self,
+      conflicts,  # type: Dict[str, Dict[Definition, List[str]]]
+      headers  # type: Dict[str, Dict[str, Union[str, int]]
```
arostamianfar: s/headers/resolved_headers

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,127 @@
...
+      row = [
+          field_id,
+          self._extract_definitions_and_file_names(definitions_to_files_map),
+          self._extract_resolution(headers, field_id)
+      ]
+      contents.append(row)
+    return contents
+
+  def _extract_definitions_and_file_names(self, definition_to_files_map):
+    # type: (Dict[Definition, List[str]]) -> str
```
arostamianfar: Please give an example of what a typical output looks like.

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,127 @@
...
+    conflict_definitions = []
+    for definition, file_names in definition_to_files_map.iteritems():
+      definition = self._format_definition(definition.num, definition.type)
+      conflict_definitions.append(definition + ' in ' + str(file_names))
+    return ' '.join(conflict_definitions)
+
+  def _extract_resolution(self, header, filed_id):
+    # type: (Dict[str, Dict[str, Union[str, int]]], str) -> str
```
arostamianfar: Please give an example of what a typical output looks like.

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,127 @@
...
+    # len(v) > 1 means there are conflicted definitions for this field.
+    return dict([(k, v) for k, v in definitions.items() if len(v) > 1])
+
+  def _generate_contents(
+      self,
+      conflicts,  # type: Dict[str, Dict[Definition, List[str]]]
+      headers  # type: Dict[str, Dict[str, Union[str, int]]
+      ):
+    # type: (...) -> List[str]
+    """Generates the report contents.
```
arostamianfar: Please give an example of what a typical output looks like (it can be a merge of the two examples below :) ).

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,127 @@
...
+
+class ConflictsReporter(object):
+  """Generates the header conflicts report."""
+
+  _HEADER_LINE = ['ID', 'Conflicts', 'Proposed Solution']
+  _REPORT_NAME = 'header_conflicts_report.csv'
+  _NO_CONFLICTS_MESSAGE = ['No conflicts found.']
+  _NO_SOLUTION_MESSAGE = 'Not resolved'
+
+  def __init__(self, directory=''):
```
arostamianfar: If we decide to remove `directory` from this class, then I think this can just be a module instead of a class (i.e. you'll have one public `generate_conflicts_report(header_defs, resolved_header, path)` method)

gcp_variant_transforms/libs/generate_report.py
```diff
@@ -0,0 +1,127 @@
...
+    self._directory = directory
+
+  def report_conflicts(self, header_definitions, resolved_header=None):
+    # type: (VcfHeaderDefinitions, VcfHeader) -> None
+    """Generates a report.
+
+    Combines the conflicts extracted from ``header_definitions`` and their
+    resolutions from ``resolved_header`` to generate a conflicts report.
+    """
+    resolved_header = resolved_header or VcfHeader()
```
arostamianfar: Why not make resolved_header required?

gcp_variant_transforms/libs/generate_report_test.py
```diff
@@ -0,0 +1,164 @@
...
+    header_definitions._formats = {'NS': {Definition(1, 'Float'): ['file2']}}
+
+    infos = OrderedDict([
+        ('NS', Info('NS', 1, 'Integer', 'Number samples', None, None))])
+    formats = OrderedDict([('NS', Format('NS', 1, 'Float', 'Number samples'))])
+    resolved_header = vcf_header_io.VcfHeader(infos=infos, formats=formats)
+
+    ConflictsReporter().report_conflicts(header_definitions, resolved_header)
+
+    report = ConflictsReporter._REPORT_NAME
```
arostamianfar: Please use testing.TempDir for these tests and please see my earlier comment about asking the user for the full path instead of hard coding it.

gcp_variant_transforms/libs/generate_report_test.py
```diff
@@ -0,0 +1,164 @@
...
+    with open(report) as f:
+      s = []
+      reader = csv.reader(f)
+      for row in reader:
+        s.append(row)
+      self.assertEqual(s, [ConflictsReporter._NO_CONFLICTS_MESSAGE])
```
arostamianfar: Consider making a `self._assert_contents_equal(file_path, expected_content)` method that performs the verification for all tests.

gcp_variant_transforms/libs/generate_report_test.py
allieychen: Done.

gcp_variant_transforms/libs/conflicts_reporter.py
```diff
@@ -0,0 +1,139 @@
...
+
+from gcp_variant_transforms.beam_io.vcf_header_io import VcfHeader  # pylint: disable=unused-import
+from gcp_variant_transforms.beam_io.vcf_header_io import VcfParserHeaderKeyConstants
+from gcp_variant_transforms.transforms.merge_header_definitions import Definition  # pylint: disable=unused-import
+from gcp_variant_transforms.transforms.merge_header_definitions import VcfHeaderDefinitions  # pylint: disable=unused-import
+
+
+_HEADER_LINE = 'ID;Conflicts;Proposed Resolution\n'
+_NO_CONFLICTS_MESSAGE = 'No conflicts found.'
+_NO_SOLUTION_MESSAGE = 'Not resolved'
```
nmousavi: nit: add a dot here too?

gcp_variant_transforms/libs/conflicts_reporter.py
```diff
@@ -0,0 +1,139 @@
...
+  content_lines.extend(_generate_contents(
+      _extract_conflicts(header_definitions.formats), resolved_headers.formats))
+  content_lines.extend(_generate_contents(
+      _extract_conflicts(header_definitions.infos), resolved_headers.infos))
+  _write_to_report(content_lines, file_path)
+
+
+def _extract_conflicts(
+    definitions  # type: Dict[str, Dict[Definition, List[str]]]
+):
```
nmousavi: nit: should `):` be moved forward by 4space? can you check with style guide? Thanks!

gcp_variant_transforms/libs/conflicts_reporter.py
```diff
@@ -0,0 +1,139 @@
...
+    which maps ``Definition`` to a list of file names.
+  """
+  # len(v) > 1 means there are conflicted definitions for this field.
+  return dict([(k, v) for k, v in definitions.items() if len(v) > 1])
+
+
+def _generate_contents(
+    conflicts,  # type: Dict[str, Dict[Definition, List[str]]]
+    resolved_headers  # type: Dict[str, Dict[str, Union[str, int]]
+):
```
nmousavi: ditto?

gcp_variant_transforms/libs/conflicts_reporter.py
```diff
@@ -0,0 +1,139 @@
...
+  # type: (...) -> List[str]
+  """Generates the report contents.
+
+  The conflicted definitions, the file names and the resolutions are included
+  in the contents.
+  Output example:
+  (NS; num=1 type=Float in ['file1','file2'] num=1 type=Integer in ['file3'];
+  num=1 type=Float)
```
nmousavi: nit: WDYT about clarifying what last num, type pair are. Something like 

` ...; candidate resolution num=1 type=Float`

or proposed resolution

gcp_variant_transforms/libs/conflicts_reporter.py
```diff
@@ -0,0 +1,139 @@
...
+    contents.append(';'.join(row))
+  return contents
+
+
+def _extract_definitions_and_file_names(definition_to_files_map):
+  # type: (Dict[Definition, List[str]]) -> str
+  """Extracts the definitions and related file names.
+
+  Output example:
+  num=1 type=Float in ['file1','file2'] num=1 type=Integer in ['file3']
```
nmousavi: nit: add a comma in between?

`num=1 type=Float in ['file1','file2'], num=1 type=Integer in ['file3']`

gcp_variant_transforms/libs/conflicts_reporter.py
allieychen: Done.

gcp_variant_transforms/libs/conflicts_reporter.py
allieychen: The style guide didn't say anything for this case. But I think leaving 4 spaces are more readable. It is also adopted in multiple modules in this project. Thanks for pointing out. Done.

gcp_variant_transforms/libs/conflicts_reporter.py
allieychen: Done.

gcp_variant_transforms/libs/conflicts_reporter.py
allieychen: This is one row in the report. In the beginning of the report, it shows a header, and ``Proposed Resolution`` is the header for this column (saved as csv file). FYI, I have added one report example.

gcp_variant_transforms/libs/conflicts_reporter.py
allieychen: Thanks, better!

PR #180, opened 2018-04-11
Add doc on how to run VT in Europe.

### LINKED ISSUE #115: Add documentation on how to run the pipeline in a particular zone/region
There is both region [1] and zone arguments (and in two places! one in the pipelines API and one in the Dataflow API). We should document how to use these features.

[1] https://cloud.google.com/dataflow/docs/concepts/regional-endpoints#supported_regional_endpoints

nmousavi:
Fixes #115 
Tested:
 no

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ status of your operation by running:
 gcloud alpha genomics operations describe <operation-id>
 ```
 
-The returned data will have `done: true` when the operation is done. 
+The returned data will have `done: true` when the operation is done.
 A detailed description of the Operation resource can be found in the
 [API documentation](https://cloud.google.com/genomics/reference/rest/v1/operations)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -154,6 +154,64 @@ python -m gcp_variant_transforms.vcf_to_bq \
   --runner DataflowRunner
 ```
 
+### Security and Compliance
```
arostamianfar: I suggest renaming this as "Running jobs in a particular region/zone" (or something less verbose). I like us to avoid mentioning compliance and we're not 100% sure about how the data is handled across regions.

README.md
```diff
@@ -154,6 +154,64 @@ python -m gcp_variant_transforms.vcf_to_bq \
...
+
+You may need to constrain Cloud Dataflow job processing to a specific
+geographic region in support of yourproject’s security and compliance needs.
+For example, the [European Union Data Protection Directive](https://cloud.google.com/security/compliance/eu-data-protection-directive/).
```
arostamianfar: ditto here. I'd remove reference to the compliance page.

README.md
```diff
@@ -154,6 +154,64 @@ python -m gcp_variant_transforms.vcf_to_bq \
...
+  cmd: |
+    ./opt/gcp_variant_transforms/bin/vcf_to_bq \
+      --project my_project \
+      --input_pattern gs://my_bucket/vcffiles/*.vcf \
+      --output_table my_project:my_bigquery_dataset.my_bigquery_table \
+      --staging_location gs://my_bucket/staging \
+      --temp_location gs://my_bucket/temp \
+      --job_name vcf-to-bigquery \
+      --runner DataflowRunner
+      --region europe-west1 \
```
arostamianfar: Can we make the new fields bolded inside the doc? Otherwise, I suggest doing something like [dsub](https://github.com/googlegenomics/dsub/blob/master/docs/code.md#--command-shell-string) where they use `...` and only emphasize the new args that need to be added.  ditto below.

README.md
```diff
@@ -154,6 +154,64 @@ python -m gcp_variant_transforms.vcf_to_bq \
...
 ```
 
+### Security and Compliance
+
+You may need to constrain Cloud Dataflow job processing to a specific
+geographic region in support of yourproject’s security and compliance needs.
+For example, the [European Union Data Protection Directive](https://cloud.google.com/security/compliance/eu-data-protection-directive/).
+
+For this, you need to specify zone and region in Dataflow API (as well as
+Pipeline API if you are using Docker). For example, in order to restrict
```
arostamianfar: nit: s/Pipeline API/the Pipelines API

README.md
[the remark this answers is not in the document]
nmousavi: Done!

nmousavi: Done!

README.md
nmousavi: Done!

README.md
nmousavi: Done!

README.md
nmousavi: Done!

PR #181, opened 2018-04-11
Resolve always if allow_incompatible_records=True.

nmousavi:
If allow_incomptible_records flag is set, we resolve any conflict
in 1) headers, 2) headers and records, and 3) records.

### DIFF
--- a/gcp_variant_transforms/transforms/merge_headers.py
+++ b/gcp_variant_transforms/transforms/merge_headers.py
@@ -118,7 +118,9 @@ class _MergeHeadersFn(beam.CombineFn):
 class MergeHeaders(beam.PTransform):
   """A PTransform to merge VCF file headers."""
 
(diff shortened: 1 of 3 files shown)


gcp_variant_transforms/transforms/merge_headers.py
```diff
@@ -118,7 +118,9 @@ def extract_output(self, merged_headers):
 class MergeHeaders(beam.PTransform):
   """A PTransform to merge VCF file headers."""
 
-  def __init__(self, split_alternate_allele_info_fields=True):
+  def __init__(self,
+               split_alternate_allele_info_fields=True,
+               allow_incompatible_records=False):
     # type: (bool) -> None
```
arostamianfar: Please update type and add to Args.

nmousavi: Done!

PR #182, opened 2018-04-12
Extend type inference for undefined header fields.

nmousavi:
Tested:
  unit test

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ status of your operation by running:
 gcloud alpha genomics operations describe <operation-id>
 ```
 
-The returned data will have `done: true` when the operation is done. 
+The returned data will have `done: true` when the operation is done.
 A detailed description of the Operation resource can be found in the
 [API documentation](https://cloud.google.com/genomics/reference/rest/v1/operations)
 
@@ -154,6 +154,13 @@ python -m gcp_variant_transforms.vcf_to_bq \
   --runner DataflowRunner
 ```
 
+### Running jobs in a particular region/zone
+
+You may need to constrain Cloud Dataflow job processing to a specific
+geographic region in support of yourproject’s security and compliance needs.
+See [Running in particular zone/region doc](docs/running_in_particular_zone_region.md)
+
+
 ## Additional topics
 
 * [Understanding the BigQuery Variants Table Schema](docs/bigquery_schema.md)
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/transforms/infer_undefined_headers.py
```diff
@@ -64,9 +62,21 @@ def _get_field_type(self, field_value):
...
+      return vcf_field_conflict_resolver.VcfParserConstants.FLOAT
     else:
       return vcf_field_conflict_resolver.VcfParserConstants.STRING
 
+  def _can_cast_to(self, value, cast_type):
+    """Returns true if `value` can be casted to type `type`"""
+    try:
+      _ = cast_type(value)
+      return True
+    except Exception:
```
arostamianfar: Please catch `TypeError` instead. Catching generic exceptions are not allowed per styleguide.

gcp_variant_transforms/transforms/infer_undefined_headers.py
```diff
@@ -64,9 +62,21 @@ def _get_field_type(self, field_value):
       return vcf_field_conflict_resolver.VcfParserConstants.INTEGER
     elif isinstance(field_value, float):
       return vcf_field_conflict_resolver.VcfParserConstants.FLOAT
+    elif self._can_cast_to(field_value, int):
```
arostamianfar: Consider doing `elif isinstance(field_value, int) or self._can_cast_to(int)` above.

gcp_variant_transforms/transforms/infer_undefined_headers.py
nmousavi: Catched `ValueError`. Done!

gcp_variant_transforms/transforms/infer_undefined_headers.py
nmousavi: It would be wrong to do so, and you will end up inferring int type for float values (and vice versa) because you can always cast float to int.

arostamianfar: ah! nice catch! Just making sure: is this already covered by unit tests? I just want to avoid someone else making the same mistake if they end up refactoring this code in the future. Perhaps you can also add a comment about it.

nmousavi: Good point. We have a test for int -> float. Added one for making sure float -> int doesn't happen.

arostamianfar: Perfect. Thanks!

PR #193, opened 2018-04-15
First version of VEP runner.

bashir2:
Please note that there are many TODOs because the scope of this PR is already too big. Some of the TODOs should be tracked under separate issues, some of them will be addressed in following PRs of Issue #81 and there are two test related ones that will be addressed during the review process of this PR.

Issue #81.

Tested:
* unit-tests added
* `./deploy_and_run_tests.sh`
* several runs on real data including the following platinum set annotation:
```
python -m gcp_variant_transforms.vcf_to_bq \
  --input_pattern gs://genomics-public-data/platinum-genomes/vcf/** \
  --output_table bashir-variant:vcf_test_imports.platinum_vep_processed_check_ref \
  --annotation_fields CSQ_VT --use_allele_num --project bashir-variant \
  --runner DataflowRunner \
  --staging_location gs://test-bashir-variant/staging \
  --temp_location gs://test-bashir-variant/temp \
  --max_num_workers 10 \
  --worker_machine_type n1-standard-8 \
  --job_name bashir-test-vep --setup_file ./setup.py \
  --run_vep \
  --vep_image_uri gcr.io/bashir-variant/vep_91 \
  --vep_cache_path gs://bashir-variant-vep_cache/vep_cache/vep_cache_homo_sapiens_GRCh38_91.tar.gz \
  --vep_num_fork 8 \
  --vep_output_dir gs://vep_test_files/output_from_vt/vep_platinum_check_ref
```

### DIFF
--- a/gcp_variant_transforms/libs/annotation/vep/test_util.py
+++ b/gcp_variant_transforms/libs/annotation/vep/test_util.py
@@ -1,0 +1,39 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+
+# TODO(bashir2): Create tests with non local inputs as well.
+INPUT_PATTERN = 'some/input/pattern*'
+INPUT_FILES_WITH_SIZE = [
+    ('some/input/pattern/a', 100),
+    ('some/input/pattern/b', 100),
+    ('some/input/pattern/c', 100),
+    ('some/input/pattern/dir1/a', 100),
+    ('some/input/pattern/dir1/dir2/b', 100),
+    ('some/input/pattern/dir2/b', 100),
+    ('some/input/pattern/dir2/c', 100),
+]
+
+
+class FileMetadataStub(object):
+  """This is an object to imitate apache_beam.io.filesystem.FileMetadata."""
+
+  def __init__(self, path, size_in_bytes):
+    self.path = path
+    self.size_in_bytes = size_in_bytes
+
+  def __repr__(self):
+    return 'path= {} size_in_bytes={}'.format(self.path, self.size_in_bytes)
--- a/gcp_variant_transforms/libs/annotation/vep/vep_runner.py
+++ b/gcp_variant_transforms/libs/annotation/vep/vep_runner.py
@@ -14,17 +14,18 @@
 
 from __future__ import absolute_import
 
-from typing import Dict, List, Any  # pylint: disable=unused-import
-
 import argparse  # pylint: disable=unused-import
 import logging
 import time
 
-from apache_beam.io import filesystems
+from typing import Dict, List, Any  # pylint: disable=unused-import
+
 from apache_beam.io import filesystem  # pylint: disable=unused-import
+from apache_beam.io import filesystems
 from apache_beam.options import pipeline_options
 from googleapiclient import discovery
 from oauth2client import client
+from gcp_variant_transforms.libs.annotation.vep import vep_runner_util
 
 
 # Minimum size of virtual machine disks to account for cache files.
@@ -48,14 +49,6 @@ _API_ACTIONS = 'actions'
 # The expected path of the run_vep.sh script in the docker container.
 _VEP_RUN_SCRIPT = '/opt/variant_effect_predictor/run_vep.sh'
 
-# The siffux that is added to output files (the output is always a VCF).
-_VEP_OUTPUT_SUFFIX = '_vep_output.vcf'
-
-# This is used for matching all files under the output directory structure.
-# It is okay for it to be Google Cloud Storage (GCS) specific because the
-# assumption of using GCS as the remote file systems is baked into VepRunner.
-_GCS_RECURSIVE_WILDCARD = '**'
-
 # The local name of the output file and directory for VEP runs.
 _LOCAL_OUTPUT_DIR = '/mnt/vep/output_files'
 _LOCAL_OUTPUT_FILE = _LOCAL_OUTPUT_DIR + '/output.vcf'
@@ -66,7 +59,7 @@ _POLLING_INTERVAL_SECONDS = 30
 
 def create_runner_and_update_args(known_args, pipeline_args):
   # type: (argparse.Namespace, List[str]) -> VepRunner
-  """Creates an instance of VepRunner using the provided args and upate them.
+  """Creates an instance of VepRunner using the provided args and updates them.
 
   In particular, the two arguments that are updated are `input_pattern` and
   `annotation_fields`.
@@ -80,9 +73,10 @@ def create_runner_and_update_args(known_args, pipeline_args):
   pipeline_service = discovery.build(
       'genomics', 'v2alpha1', credentials=credentials)
   runner = VepRunner(
-      pipeline_service, known_args.input_pattern, known_args.vep_output_dir,
-      known_args.vep_info_field, known_args.vep_image_uri,
-      known_args.vep_cache_path, known_args.vep_num_fork, pipeline_args)
+      pipeline_service, known_args.input_pattern,
+      known_args.annotation_output_dir, known_args.vep_info_field,
+      known_args.vep_image_uri, known_args.vep_cache_path,
+      known_args.vep_num_fork, pipeline_args)
   known_args.input_pattern = runner.get_output_pattern()
   if known_args.annotation_fields:
     known_args.annotation_fields.append(known_args.vep_info_field)
@@ -130,13 +124,11 @@ class VepRunner(object):
     self._output_dir = output_dir
     self._vep_info_field = vep_info_field
     self._process_pipeline_args(pipeline_args)
-    # _current_operations holds the list of operation names currently running.
     self._running_operation_ids = []  # type: List[str]
 
   def get_output_pattern(self):
     # type: () -> str
-    return (self._format_dir_path(self._output_dir) +
-            _GCS_RECURSIVE_WILDCARD + _VEP_OUTPUT_SUFFIX)
+    return vep_runner_util.get_output_pattern(self._output_dir)
 
   def _get_api_request_fixed_parts(self):
     # type: () -> Dict
@@ -155,7 +147,7 @@ class VepRunner(object):
             ],
             'environment': {
                 'VEP_CACHE': '/mnt/vep/vep_cache/{}'.format(
-                    _get_base_name(self._vep_cache_path)),
+                    vep_runner_util.get_base_name(self._vep_cache_path)),
                 'NUM_FORKS': str(self._vep_num_fork),
                 'VCF_INFO_FILED': self._vep_info_field,
                 # TODO(bashir2): Decide how to do proper reference validation,
@@ -243,8 +235,9 @@ class VepRunner(object):
 
   def _is_done(self, operation):
     # type: (str) -> bool
-    # TODO(bashir2): Silence the log messages of googleapiclient.discovery
-    # module for the next call of the API since they flood the log file.
+    # TODO(bashir2): Catch exceptions in the following call and also silence the
+    # log messages of googleapiclient.discovery module for the next call of the
+    # API since they flood the log file.
     # pylint: disable=no-member
     request = self._pipeline_service.projects().operations().get(name=operation)
     is_done = request.execute()['done']
@@ -265,16 +258,16 @@ class VepRunner(object):
       raise AssertionError('There are already {} operations running.'.format(
           len(self._running_operation_ids)))
     logging.info('Finding all files that match %s', self._input_pattern)
-    match_results = filesystems.FileSystems.match([
-        self._input_pattern])  # type: List[filesystem.MatchResult]
+    match_results = filesystems.FileSystems.match(
+        [self._input_pattern])  # type: List[filesystem.MatchResult]
     if not match_results:
       raise ValueError('No files matched input_pattern: {}'.format(
           self._input_pattern))
     logging.info('Number of files: %d', len(match_results[0].metadata_list))
     self._check_and_write_to_output_dir(self._output_dir)
-    pipelines_data = _PipelinesExecutionInfo(
+    single_vm_actions_list = vep_runner_util.disribute_files_on_workers(
         match_results[0].metadata_list, self._output_dir, self._max_num_workers)
-    for vm_index, actions in enumerate(pipelines_data.single_vm_actions_list):
+    for vm_index, actions in enumerate(single_vm_actions_list):
       operation_name = self._call_pipelines_api(
           actions, self._get_output_log_path(self._output_dir, vm_index))
       logging.info('Started operation %s on VM %d processing %d input files',
@@ -282,7 +275,7 @@ class VepRunner(object):
       self._running_operation_ids.append(operation_name)
 
   def _call_pipelines_api(self, single_vm_actions, output_log_path):
-    # type: (_SingleWorkerActions, str) -> str
+    # type: (vep_runner_util.SingleWorkerActions, str) -> str
     api_request = self._get_api_request_fixed_parts()
     size_gb = single_vm_actions.disk_size_bytes / (1 << 30)
     api_request[_API_PIPELINE]['resources'][
@@ -303,7 +296,7 @@ class VepRunner(object):
 
   def _check_and_write_to_output_dir(self, output_dir):
     # type: (str) -> None
-    real_dir = self._format_dir_path(output_dir)
+    real_dir = vep_runner_util.format_dir_path(output_dir)
     # NOTE(bashir2): We cannot use exists() because for example on GCS, the
     # directory names are only symbolic and are not physical files.
     match_results = filesystems.FileSystems.match(['{}*'.format(real_dir)])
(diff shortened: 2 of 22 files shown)


gcp_variant_transforms/libs/annotation/vep_runner_test.py
```diff
@@ -0,0 +1,118 @@
...
+    self._mock_request = mock.Mock()  # pylint: disable=attribute-defined-outside-init
+    self._mock_service.pipelines = mock.Mock(return_value=self._mock_pipelines)
+    self._mock_pipelines.run = mock.Mock(return_value=self._mock_request)
+    self._mock_request.execute = mock.Mock(return_value={'name': 'operation'})
+    mock_service_builder = mock.Mock(return_value=self._mock_service)
+    mock_app_default = mock.Mock(return_value=default_cred)
+    # TODO: It is better to pass a `service` object to VepRunner instead of
+    # this patch gymnastic to create a `mock_service`. A factor method can be
+    # provided to create instances of VepRunner. This needs to be decided in
+    # the review process and updated before submit.
```
bashir2: Asha, before adding more unit-tests I wanted to make sure the general test related design is agreed upon because last time you had some concerns with dependency injection patterns and suggested `patch`. This is one example that I think the right approach is to inject the service object into VepRunner but it is also possible to replace real objects with mocks using `patch`. Please let me know what you think.

gcp_variant_transforms/libs/annotation/vep_runner.py
```diff
@@ -0,0 +1,393 @@
...
+    # _current_operations holds the list of operation names currently running.
+    self._current_operations = []  # type: List[str]
+
+  def get_output_pattern(self):
+    # type: () -> str
+    return (self._make_sure_dir(self._output_dir) +
+            _GCS_RECURSIVE_WILDCARD + _VEP_OUTPUT_SUFFIX)
+
+  def _get_api_request_fixed_parts(self):
+    # tyep: () -> Dict
```
allieychen: s/tyep/type. Happened to catch the typo. :)

gcp_variant_transforms/libs/annotation/vep_runner.py
```diff
@@ -0,0 +1,393 @@
...
+    # type: (str) -> str
+    """Returns `dir_name` possibly with an added '/' if not already included."""
+    return filesystems.FileSystems.join(dir_name, '')
+
+  def _create_output_log_path(self, output_dir, vm_index):
+    # type: (str, int) -> str
+    return '{}/logs/output_VM_{}'.format(output_dir, vm_index)
+
+  def _create_actions(self, input_file, output_file):
+    # type: (str) -> List
```
allieychen: s/(str)/(str, str)

gcp_variant_transforms/libs/annotation/vep_runner.py
```diff
@@ -0,0 +1,393 @@
...
+        raise ValueError('Expected {}{} at the beginning of input {}'.format(
+            scheme, self._FILE_SYSTEM_SCHEME_SEPARATOR, input_path))
+      output_file = input_path[len(scheme) +
+                               len(self._FILE_SYSTEM_SCHEME_SEPARATOR):]
+    output_file += _VEP_OUTPUT_SUFFIX
+    return filesystems.FileSystems.join(self._output_dir, output_file)
+
+
+def _get_base_name(file_path):
+  # type: (str) -> str
```
arostamianfar: Please add a comment about why this is needed and give an example.

gcp_variant_transforms/libs/annotation/vep_runner.py
```diff
@@ -0,0 +1,393 @@
...
+#
+# API constants:
+_API_PIPELINE = 'pipeline'
+_API_ACTIONS = 'actions'
+
+# The expected path of the run_vep.sh script in the docker container.
+_VEP_RUN_SCRIPT = '/opt/variant_effect_predictor/run_vep.sh'
+
+# The siffux that is added to output files (the output is always a VCF).
+_VEP_OUTPUT_SUFFIX = '_vep_output.vcf'
```
arostamianfar: To clarify: does VEP always produce .vcf files even if the input is gzipped?

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+    if len(current_worker.io_map) >= num_jobs:
+      return True
+    return False
+
+  def _map_to_output_dir(self, input_path):
+    # type: (str) -> (str)
+    """Maps an input path to its corresponding output path.
+
+    For example, for `input_path` being 'gs://my_bucket/input.vcf', it returns
+    'gs://output_bucket/out_dir/my_bucket/input.vcf' where `self._output_dir`
```
arostamianfar: Please update comment as it also includes the VEP suffix.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+_LOCAL_OUTPUT_DIR = '/mnt/vep/output_files'
+_LOCAL_OUTPUT_FILE = _LOCAL_OUTPUT_DIR + '/output.vcf'
+
+# The time between operation polling rounds.
+_POLLING_INTERVAL_SECONDS = 30
+
+
+def create_runner_and_update_args(known_args, pipeline_args):
+  # type: (argparse.Namespace, List[str]) -> VepRunner
+  """Creates an instance of VepRunner using the provided args and upate them.
```
arostamianfar: s/upate/updates

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+      num_workers: Maximum number of workers to use.
+    """
+    self._single_vm_actions_list = []  # type: List[_SingleWorkerActions]
+    self._is_zip_input = False
+    self._output_dir = output_dir
+    self._num_workers = num_workers
+    self._input_size = len(file_metadata_list)
+
+    current_worker = _SingleWorkerActions()
+    for metadata in file_metadata_list:
```
arostamianfar: I think this logic can be simplified a lot. Essentially, you are dividing n tasks among m workers. I believe something like this should work:
```
file_metadata_chunks = [file_metadata_list[i::num_workers] for i in range(num_workers)]
for file_metadata_chunk in file_metadata_chunks:
  if not file_metadata_chunk: continue  # in case num files < num workers
  worker = _SingleWorkerActions()
  for file_metadata in file_metadata_chunk:
    worker.io_map[file_metadata.path] = ...
    worker.disk_size += ...
  self._single_vm_actions_list.append(worker)
```

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+#
+# API constants:
+_API_PIPELINE = 'pipeline'
+_API_ACTIONS = 'actions'
+
+# The expected path of the run_vep.sh script in the docker container.
+_VEP_RUN_SCRIPT = '/opt/variant_effect_predictor/run_vep.sh'
+
+# The siffux that is added to output files (the output is always a VCF).
+_VEP_OUTPUT_SUFFIX = '_vep_output.vcf'
```
arostamianfar: ah, I see. `my.vcf` to `my.vcf_vep_output.vcf` sounds good!

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+                    _get_base_name(self._vep_cache_path)),
+                'NUM_FORKS': str(self._vep_num_fork),
+                'VCF_INFO_FILED': self._vep_info_field,
+                # TODO(bashir2): Decide how to do proper reference validation,
+                # the following --check_ref just drops variants that have
+                # wrong REF. If there are too many of them, it indicates that
+                # VEP database for a wrong reference sequence is being used
+                # and this has to caught and communicated to the user.
+                'OTHER_VEP_OPTS':
+                    '--everything --check_ref --allow_non_variant',
```
arostamianfar: Let's leave as is for this PR. I agree with you that limiting the # of flags is a good idea. In general, users get confused and will use whatever is the default values anyway.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+    if flag not in pipeline_flags or not pipeline_flags[flag]:
+      raise ValueError('Could not find {} among pipeline flags {}'.format(
+          flag, pipeline_flags))
+    logging.info('Using %s flag: %s.', flag, pipeline_flags[flag])
+    return pipeline_flags[flag]
+
+  def wait_until_done(self):
+    """Polls currently running operations and waits until all are done."""
+    while self._running_operation_ids:
+      self._running_operation_ids = [op for op in self._running_operation_ids
```
arostamianfar: ah, I thought you were checking for `done` operations over and over again and did not realize that `_running_operation_ids` is self-assigned. LGTM! :)

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+    operation_name = request.execute()['name']
+    return operation_name
+
+  def _check_and_write_to_output_dir(self, output_dir):
+    # type: (str) -> None
+    real_dir = self._format_dir_path(output_dir)
+    # NOTE(bashir2): We cannot use exists() because for example on GCS, the
+    # directory names are only symbolic and are not physical files.
+    match_results = filesystems.FileSystems.match(['{}*'.format(real_dir)])
+    if match_results and match_results[0].metadata_list:
```
arostamianfar: I see your point. However, from my DeepVariant experience, users really like just specifying a single output path for all of their runs and don't care about the details. I can bet that users will keep hitting this exception when running the pipeline as they'd forget to assign a unique folder (a common case: user runs program, it fails b/c they don't have the correct VEP cache, they correct it, rerun, the program now fails saying that the log file exists and directory is not empty). Let's discuss to see if there is a way that we can achieve both goals (i.e. provide the flexibility for the user to specify their exact output folder, but also not fail if they dont).

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+    self.io_map = {}  # type: Dict[str, str]
+
+  def __repr__(self):
+    return 'disk_size_bytes= {} , io_map= {}'.format(
+        self.disk_size_bytes, str(self.io_map))
+
+
+# TODO decide whether we want to unit-test this class in isolation and if so
+# move this and _SingleWorkerActions to a separate module.
+class _PipelinesExecutionInfo(object):
```
arostamianfar: My main concern is that making this a class implies that it's doing some 'work' and actually has some state, whereas the 'state' can just be achieved as a return value. I don't feel strongly about it though, so feel free to leave as is.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+
+  # The part of a path separating the file system scheme from the actual path.
+  _FILE_SYSTEM_SCHEME_SEPARATOR = '://'
+
+  # The expected suffix for .gz and .bgz files.
+  _GZ_SUFFIX = 'gz'
+
+  # This is used as a heuristic to account for the size of the unzipped file
+  # based on some anecdotal samples.
+  # TODO(bashir2): Revisit the file size calculation logic.
```
arostamianfar: Having the TODO is ok for now :)

gcp_variant_transforms/libs/annotation/vep/vep_runner_test.py
```diff
@@ -93,13 +82,59 @@ def test_get_output_pattern(self):
...
+  We need this matcher to avoid duplicating the whole JSON object created in
+  the production code. Instead we just do a simple heuristic match by going
+  through all `commands` and check if at least in one of them one of the
+  input files appear.
+  """
+
+  def __init__(self, input_file_list):
+    self.input_file_set = set(input_file_list)
+
+  def __eq__(self, other):
```
bashir2: Please note that this pattern can be replaced with a Spy pattern similar to what I have done in [_CounterSpy](https://github.com/googlegenomics/gcp-variant-transforms/blob/b5ae3e74699e5231699d724483f0d0cc2e49f248/gcp_variant_transforms/libs/processed_variant_test.py#L30).

Here I was mostly trying to see how easy/readable it is to achieve my testing goals with mock only. So if it is hard to understand, there are less pythonic mock alternatives too.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+
+# Minimum size of virtual machine disks to account for cache files.
+# NOTE(bashir2): The proper way of setting this is to measure the input cache
+# file size and use some heuristics to calculate the size of the decompressed
+# version. The current value is more than double what is needed for 'human'.
+_MINIMUM_DISK_SIZE_GB = 200
+
+# The name of the file placed at the root of output_dir that includes
+# information on how the pipelines were run, input files, etc.
+_GLOBAL_LOG_FILE = 'VEP_run_info.txt'
```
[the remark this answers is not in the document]
bashir2: Done.
But please note that this file is currently empty and solely used as a lock to prevent overwrites (there is a TODO below to put some the run information in it). The actual log files are `output_VM*` (check `_get_output_log_path` below). I added `.log` extension there as well.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -0,0 +1,464 @@
...
+                    # TODO(bashir2): Add the option of using preemptible
+                    # machines and the retry functionality.
+                    'preemptible': False,
+                    'serviceAccount': {
+                        'scopes': [
+                            'https://www.googleapis.com/auth/'
+                            'devstorage.read_write']
+                    }
+                },
+                'regions': [
```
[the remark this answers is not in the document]
bashir2: I made this particular suggestion and one more in `_make_action` below. Not sure about the dsub link you shared because it is in `is_pipelines_operation` function (maybe you meant to share a permanent link?) but in any case, there are probably some more ways to make these more compact that may or may not hinder readability. I am fine the way it is now.

gcp_variant_transforms/libs/annotation/vep/vep_runner_util.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
```
arostamianfar: In general, please avoid creating generic util classes as much as possible since they become a bag of unrelated stuff over time (i'd argue that it already is). I think a better approach is to make a module that just contains the file distribution logic and then move the rest into vep_runner module (or create another module). I'm ok to leave as is for now though as I don't want to block this PR much longer :)

gcp_variant_transforms/libs/annotation/vep/vep_runner_test.py
arostamianfar: This is ok. But, could you please make the `__eq__` method idempotent? It's strange how it's changing the state of the class. I think it's more readable to to create a set of `command_part`  elements and then just check whether the two sets intersect.

gcp_variant_transforms/libs/annotation/vep/test_util.py
```diff
@@ -0,0 +1,39 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+
+# TODO(bashir2): Create tests with non local inputs as well.
+INPUT_PATTERN = 'some/input/pattern*'
```
arostamianfar: I think it's odd to include sample patterns here unless it's being used very often. I personally think it's ok to repeat them since it's only two tests for now. If you decide to keep them though, please change to a method like `get_sample_input_data()` that returns both pattern + files and adds some connection between the two. We did a similar thing in [testdata_util.py](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/testing/testdata_util.py#L39).

gcp_variant_transforms/libs/annotation/vep/vep_runner_test.py
```diff
@@ -0,0 +1,140 @@
...
+  @staticmethod
+  def create(path, mime_type=None, compression_type=None):
+    """Overriding `create` to remove any interaction with real file systems."""
+    return mock.Mock()
+
+
+class VepRunnerTest(unittest.TestCase):
+
+  def _create_test_instance(self, pipeline_args=None):
+    self._mock_service = mock.Mock()  # pylint: disable=attribute-defined-outside-init
```
arostamianfar: Can these be defined in `setup()`?

gcp_variant_transforms/libs/annotation/vep/vep_runner_test.py
```diff
@@ -0,0 +1,140 @@
...
+
+  def _get_pipeline_args(self, num_workers=1):
+    return ['--project', _PROJECT,
+            '--region', _REGION,
+            '--max_num_workers', str(num_workers),
+            '--worker_machine_type', 'n1-standard-8',
+           ]
+
+  def test_instantiation(self):
+    """This is just to test object construction."""
```
arostamianfar: I'm curious if you expect this to actually catch something that's not covered by other tests? If not, I suggest removing it as it's trivial.

gcp_variant_transforms/libs/annotation/vep/vep_runner_test.py
bashir2: Hmm, not being idempotent is actually the point. So if you do equality comparison on two sets of actions that have the same file (from the input set) the first one should succeed but the second should fail because we don't want to run VEP twice on the same file; note the "or it was repeated" in the error log message below.

Your comment actually makes my concerns stronger about this code not being easy to understand. As I said, I can change it to the Spy pattern which is probably easier to read/understand (For now, I extended class comment a little to make this more clear but good code should be self documenting as much as possible).

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
bashir2: BTW, to make sure all comments are replied: As discussed off-PR, while I agree with your point about nuisance of forcing this constraint, I think there is a fundamental difference between this example and DeepVariant. In the DV case I think users almost never care about intermediate files content but for annotations and VEP, those output files are actually valuable independent of BigQuery, specially give the current slow way that we have for running VEP.

Obviously we can/should revisit this and several other decisions we have made about this first version after we get some user feedback.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -177,6 +177,40 @@ def add_arguments(self, parser):
...
+              'VCFs before loading to BigQuery.'))
+    parser.add_argument(
+        '--vep_image_uri',
+        default="",
+        help=('The URI of the docker image for VEP.'))
+    parser.add_argument(
+        '--vep_cache_path',
+        default="",
+        help=('The path for VEP cache on Google Cloud Storage.'))
+    parser.add_argument(
```
arostamianfar: nit: please move this arg to the top to make all VEP-specific args grouped together.

gcp_variant_transforms/libs/annotation/vep/vep_runner_test.py
bashir2: Done.

gcp_variant_transforms/options/variant_transform_options.py
bashir2: Done.

Issue #189, opened 2018-04-12
Add python type checking to all classes


arostamianfar:
We have adopted PEP-484 for our type checking, but many classes are still missing them. We should add them to all existing classes.

allieychen:
In Q2, the type checking will be added to public functions and methods. If there were types listed in `Args` for private methods, they will also be converted as type hint following PEP-484.

Issue #178, opened 2018-04-10
Implement a pre-processing tool


arostamianfar:
Now that we have 'robust' VCF import (see #101, #79, and [design doc](https://docs.google.com/document/d/1JusaPllbpXynNziNzdyOdSRTM8i7UXbNvtkbrWNMq3Q/edit)), we should implement a pre-processing tool to aid users when importing malformed and/or incompatible VCF files.
This tool will essentially extend the existing proposal for a "resource estimator tool" (see #67) and generate a report of incompatible VCF fields as well as resolution suggestions (e.g. resolving "float" and "string" into "string").

PR #194, opened 2018-04-16
refator vcf to bq script #178

allieychen:
Before writing the generating conflicts report script for [Issue 178](https://github.com/googlegenomics/gcp-variant-transforms/issues/178)(Basically creating another script similar to `vcf_to_bq`: parsing the arguments and define the pipelines), it is observed that there are some functions can be reused by `vcf_to_bq` and the `preprocessor`. 

Therefore, refactored the `vcf_to_bq` and moved some general functions that can be shared by `vcf_to_bq` and the `preprocessor` to a new module.

Type hints ([Issue 189](https://github.com/googlegenomics/gcp-variant-transforms/issues/189)) are also added.

Tested: unit tests

### DIFF
--- a/gcp_variant_transforms/general_process.py
+++ b/gcp_variant_transforms/general_process.py
@@ -1,0 +1,151 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""General functions that are used by both vcf_to_bq and preprocessor.
+
+It includes parsing the command line arguments, reading the input, applying the
+PTransforms and writing the output.
+"""
+
+from argparse import ArgumentParser
+from argparse import Namespace  # pylint: disable=unused-import
+from typing import List  # pylint: disable=unused-import
+import enum
+
+import apache_beam as beam
+from apache_beam.io.filesystems import FileSystems
+from apache_beam import Pipeline  # pylint: disable=unused-import
+from apache_beam.pvalue import PCollection  # pylint: disable=unused-import
+
+from gcp_variant_transforms.beam_io import vcfio
+from gcp_variant_transforms.beam_io import vcf_header_io
+from gcp_variant_transforms.options.variant_transform_options import VariantTransformsOptions  # pylint: disable=unused-import
+from gcp_variant_transforms.transforms import filter_variants
+from gcp_variant_transforms.transforms import infer_undefined_headers
+from gcp_variant_transforms.transforms.merge_headers import MergeHeaders  # pylint: disable=unused-import
+
+# If the # of files matching the input file_pattern exceeds this value, then
+# headers will be merged in beam.
+_SMALL_DATA_THRESHOLD = 100
+_LARGE_DATA_THRESHOLD = 50000
+
+
+class PipelineModes(enum.Enum):
+  """An Enum specifying the mode of the pipeline based on the data size."""
+  SMALL = 0
+  MEDIUM = 1
+  LARGE = 2
+
+
+def parse_args(argv, command_line_options):
+  # type: (List[str], List[type]) -> (Namespace, List[str])
+  """Parses the arguments.
+
+  Args:
+    argv: A list of string representing the pipeline arguments.
+    command_line_options: A list of type ``VariantTransformsOptions`` that
+      specifies the options that will be added to parser.
+  """
+  parser = ArgumentParser()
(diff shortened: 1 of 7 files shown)


gcp_variant_transforms/general_process.py
```diff
@@ -0,0 +1,151 @@
...
+  return known_args, pipeline_args
+
+
+def get_pipeline_mode(known_args):
+  # type: (Namespace) -> int
+  """Returns the mode the pipeline should operate in based on input size."""
+  if known_args.optimize_for_large_inputs:
+    return PipelineModes.LARGE
+
+  match_results = FileSystems.match([known_args.input_pattern])
```
allieychen: @bashir2 For the type hints `Namespace`, I got warnings as unresolved attribute reference for class `Namespace`, I didn't find a good way to resolve it. Any suggestions?

bashir2: Right, the type checking system cannot understand these kind of dynamic objects (thanks to Python). With a quick look at argparse, I think these attributes (e.g., input_pattern) are added dynamically.

BTW, you can just use argparse.Namespace if you want to avoid importing the Namespace class (since the style guide recommends against it).

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -33,32 +33,33 @@
...
-from gcp_variant_transforms.beam_io import vcf_header_io
-from gcp_variant_transforms.beam_io import vcfio
+from gcp_variant_transforms import general_process
 from gcp_variant_transforms.libs import metrics_util
 from gcp_variant_transforms.libs import vcf_header_parser
 from gcp_variant_transforms.libs.variant_merge import merge_with_non_variants_strategy
 from gcp_variant_transforms.libs.variant_merge import move_to_calls_strategy
 from gcp_variant_transforms.libs import processed_variant
 from gcp_variant_transforms.options import variant_transform_options
+from gcp_variant_transforms.options.variant_transform_options import MergeOptions
```
arostamianfar: The styleguide actually recommends not importing classes and just importing modules...we've not been using that in tests and the unsuced imports though... but, this is an interesting case; could you use the full path in `type`? Given that it's only once place, then it shouldn't be that bad i guess.

gcp_variant_transforms/general_process.py
```diff
@@ -0,0 +1,151 @@
...
+          infer_undefined_headers.InferUndefinedHeaderFields(
+              beam.pvalue.AsSingleton(merged_header)))
+
+
+def read_headers(pipeline, pipeline_mode, known_args):
+  # type: (Pipeline, int, Namespace) -> PCollection
+  """Creates an initial PCollection by reading the VCF file headers."""
+  if pipeline_mode == PipelineModes.LARGE:
+    headers = pipeline | (beam.Create([known_args.input_pattern])
+                          | vcf_header_io.ReadAllVcfHeaders())
```
arostamianfar: nit: this indentation pattern is a bit awkward; I think this is better:
```
headers |= (pipeline
    | beam.Create([known_args.input_pattern])
    | vcf_header_io.ReadAllVcfHeaders())
```

gcp_variant_transforms/general_process.py
```diff
@@ -0,0 +1,151 @@
...
+  return (read_variants(pipeline, known_args)
+          | 'FilterVariants' >> filter_variants.FilterVariants(
+              reference_names=known_args.reference_names)
+          | ' InferUndefinedHeaderFields' >>
+          infer_undefined_headers.InferUndefinedHeaderFields(
+              beam.pvalue.AsSingleton(merged_header)))
+
+
+def read_headers(pipeline, pipeline_mode, known_args):
+  # type: (Pipeline, int, Namespace) -> PCollection
```
arostamianfar: nit: pipeline_mode is actually an Enum. Does InteliJ support it?

gcp_variant_transforms/general_process.py
```diff
@@ -0,0 +1,151 @@
...
+  if not match_results:
+    raise ValueError('No files matched input_pattern: {}'.format(
+        known_args.input_pattern))
+
+  total_files = len(match_results[0].metadata_list)
+  if total_files > _LARGE_DATA_THRESHOLD:
+    return PipelineModes.LARGE
+  elif total_files > _SMALL_DATA_THRESHOLD:
+    return PipelineModes.MEDIUM
+
```
arostamianfar: nit: please remove extra blank line :)

gcp_variant_transforms/general_process.py
```diff
@@ -0,0 +1,151 @@
...
+  return PipelineModes.SMALL
+
+
+def _add_parser_arguments(options, parser):
+  # type: (List[VariantTransformsOptions], ArgumentParser) -> None
+  for transform_options in options:
```
arostamianfar: Now that we have these in a separate module, I think it's ok to just inline them in the `parse_args` method (it becomes more readable that way).

gcp_variant_transforms/general_process.py
```diff
@@ -0,0 +1,151 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
```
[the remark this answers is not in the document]
arostamianfar: Perhaps `vcf_to_bq_common.py`. We can name the preprocessor `vcf_to_bq_preprocessor.py`.

allieychen:
Yes, I think so. I will check the unit tests to find out why it is decreased.

Issue #78, opened 2018-01-23
Optimize merging logic to handle massive number of records (300B+)


arostamianfar:
The pipeline is currently very slow when merging massive number of records (300B+).
Optimization strategies:
- Experiment with [Cloud Dataflow Shuffle](https://cloud.google.com/blog/big-data/2017/06/introducing-cloud-dataflow-shuffle-for-up-to-5x-performance-improvement-in-data-analytic-pipelines): it allegedly gives 5x performance improvement over the standard method.
- Experiment with sharding records by chromosome: We currently throw everything in one giant PCollection and let Dataflow sort them out. We may be able optimize the pipeline by first sharding by chromosome.

Optimizing this step will be more important especially when we add more advanced merging strategies (e.g. merge with non-variants).

arostamianfar:
Thanks for investigating the Cloud Dataflow Shuffle, Nima! 
Update: using Cloud Dataflow Shuffle actually solves the issue of handling massive number of records when merging. However, it can still be expensive since it's attempting to merge data from all chromosomes. We can optimize this further by partitioning the data by chromosome first. Saman is going to work on partitioning data for sharded BQ export, and this also fits nicely there (we can "kill two birds with one stone" :p ). Reassigning to Saman.

arostamianfar:
Saman, now that Cloud Dataflow Shuffle pricing is updated, could you please try with `--experiments=shuffle_mode=service` with the platinum1000 test to see what cost/performance you get? I think we can close this issue once that experiment is done.

arostamianfar:
Closing this bug. Summary:
- Using the shuffle service is actually more expensive than our native partition-based merge (turned on with the `--optimize_for_large_inputs` flag).
- Using a large number of small workers is much more efficient than using a small number of large workers.
- Using SSDs can further reduce cost if a large enough worker pool is not available.

More details are provided in the [handling large inputs](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/docs/large_inputs.md) doc.

PR #195, opened 2018-04-17
Partitioning before Merging

samanvp:
In order to be able to process large datasets (#78), we need to partition
input into smaller patches so that Merging would be possible to run on a
ordinary VM. Here we add the ability to partition variants based on
their reference_name (CHROM).

This change enables VT to process Platinum_100 (with a caviate of some
changes needed for writing to BigQuery table).


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -248,8 +252,15 @@ def run(argv=None):
...
   if variant_merger:
-    variants |= (
-        'MergeVariants' >> merge_variants.MergeVariants(variant_merger))
+    if known_args.optimize_for_large_inputs:
+      partitions = variants | 'PartitionVariants' >> beam.Partition(
+          partition_variants.PartitionVariants(), _DEFAULT_PARTITION_NO)
+      merged = [(partitions[i] | 'MergeVariants'+str(i) >>
+                 merge_variants.MergeVariants(variant_merger))
+                for i in xrange(_DEFAULT_PARTITION_NO)]
+      variants = merged | 'FlattenPartitions' >> beam.Flatten()
```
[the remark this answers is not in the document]
samanvp: Note this implementation does not make any guarantee that each chromosome will be assigned to a separate partition. If reference names do not follow the standard format ('^(c|ch|chr)?([0-9][0-9]?)$') we hash them and randomly assign them to a bucket (without checking for collisions). Even when the reference names have standard format we assign chromosome 1-22 to partitions [0, 21] and the remaining ones (chrX, chrY, chrM, or any variants with unknown_... reference_name) are assigned randomly to other partition indices > 21.

For the use case you mentioned, we will be using the partition_config.yaml file (which is being implemented right now). We will offer a default config file so that users can partition their output table by just setting a flag to the default config file.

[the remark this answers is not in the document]
bashir2: Thanks Saman for the clarifications. What I am trying to get is to avoid parsing a config file for chromosome based partitioning. Have you thought about generating list of chromosome strings (possibly with count) and feeding it back to the pipeline using side-inputs? I have added both topics related to partitioning to the weekly meeting agenda, we can discuss further there.

gcp_variant_transforms/transforms/partition_variants.py
```diff
@@ -0,0 +1,73 @@
...
+# Partition 0 to 21 is reserved for chromosomes with a common name, ie "ch1".
+_RESERVED_PARTITIONS = 22
+
+class PartitionVariants(beam.PartitionFn):
+  """Partitions variants based on thier chromosome."""
+
+  def __init__(self):
+    """Initializes the transform, mainly a dic of chrom->index for quick look up."""
+
+    self._by_chr = collections.defaultdict(int)
```
arostamianfar: It doesnt' look like you're actually using the defaultdict functionality, so why not just use a simple dict?

gcp_variant_transforms/transforms/partition_variants_test.py
```diff
@@ -0,0 +1,114 @@
...
+    #   re.compile(r'^(c|ch|chr)?([0-9][0-9]?)$')
+    variants = []
+    # Partition 0
+    variants.append(vcfio.Variant(reference_name='chr1'))
+    variants.append(vcfio.Variant(reference_name='CHR1'))
+    # Partition 1
+    variants.append(vcfio.Variant(reference_name='ch2'))
+    variants.append(vcfio.Variant(reference_name='CH2'))
+    # Partition 2
+    variants.append(vcfio.Variant(reference_name='c3'))
```
arostamianfar: It may actually be better to use some more realistic examples here as well.

gcp_variant_transforms/transforms/partition_variants_test.py
```diff
@@ -0,0 +1,114 @@
...
+    variants.append(vcfio.Variant(reference_name='chr06'))
+    # Partition 9
+    variants.append(vcfio.Variant(reference_name='chr10'))
+    variants.append(vcfio.Variant(reference_name='  chr10   '))
+    # Partition 21
+    variants.append(vcfio.Variant(reference_name='chr22'))
+    variants.append(vcfio.Variant(reference_name='chr22    '))
+    return variants, variants[0:2], variants[2:4], variants[4:6], variants[6:9], variants[9:11], variants[11:13]
+
+  def _get_invalid_variants(self):
```
arostamianfar: consider rewording 'valid' and 'invalid'  (perhaps 'nonstandard' and 'standard'?)

gcp_variant_transforms/transforms/partition_variants_test.py
```diff
@@ -0,0 +1,114 @@
...
+    variants.append(vcfio.Variant(reference_name='chr23'))
+    variants.append(vcfio.Variant(reference_name='chr99'))
+    variants.append(vcfio.Variant(reference_name='chrX'))
+    variants.append(vcfio.Variant(reference_name='chrY'))
+    variants.append(vcfio.Variant(reference_name='chrM'))
+    # All these variants will not match to standard reference_name Reg Exp, thus
+    # they will all end up in the default partition.
+    return variants, variants
+
+  def test_partition_variants(self):
```
arostamianfar: I think the test can be written in a more readable manner. Let's chat offline.

gcp_variant_transforms/vcf_to_bq.py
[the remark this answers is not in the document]
arostamianfar: Bashir, what is the drawback of using a config file? We'd need to implement that logic anyway (and it's not that complicated), but does provide a robust way of partitioning. Note that even the 'hg19 reference' genome has a log of random chromosome names: http://hgdownload.cse.ucsc.edu/goldenpath/hg19/chromosomes/

setup.py
```diff
@@ -25,6 +25,7 @@
     'google-api-python-client>=1.6',
     'intervaltree>=2.1.0,<2.2.0',
     'pyvcf<0.7.0',
+    'mmh3>=2.5.1',
```
arostamianfar: Given the random breakages that we've had in the past due to pip packages getting updated, could we pin this particular version instead? Or have a lower limit of versions? I suspect it's not used by any of our dependencies.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -74,7 +75,10 @@
 _LARGE_DATA_THRESHOLD = 50000
 _MERGE_HEADERS_FILE_NAME = 'merged_headers.vcf'
 _MERGE_HEADERS_JOB_NAME = 'merge-vcf-headers'
-
+# The first 22 partitions [0, 21] are reserved for standard reference_name
+# formated as RegExp '^(c|ch|chr)?([0-9][0-9]?)$'.
+# Every other identifers will be matched to next available partitions [22, 26].
+_DEFAULT_PARTITION_NO = 22 + 5
```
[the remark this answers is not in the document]
arostamianfar: I see. Thanks for the extra context. However, I have been thinking about this and I'm a bit uneasy about having this random number in our main module, and later the PartitionConfig class added here as well. I think a better approach could be:
- Create a new libs.variant_partition.VariantPartition class that inherits from object (i.e. simple lib class).
- Define this constant in that module.
- Implement a `get_num_partitions()` method that, for now, just returns that constant. Later on, it can parse the config file and return the appropriate value.
- Implement a `partition(variant)` method inside that lib and move all of the partitioning logic there.
- Change PartitionVariants to get the VariantPartition lib as an arg. The partition_for method simply does `return variant_partition.partition(variant)`. You may actually choose to remove the ParitionVariants class altogether since it's just a dummy class now and can call VariantPartition.partition directly (you'd have a dummy num_partitions arg, but that's ok).

Finally, the code in main becomes something like:
```
variant_partition_lib = VariantPartition() # this can later take in config_file_path
partitions = variants | 'PartitionVariants' >> beam.Partition(variant_parition_lib.partition, variant_parition_lib.get_num_partitions())
merged = [... for i in xrange(variant_parition_lib.get_num_partitions())]
...
```

The benefit of this approach is that all of the partitioning logic lives in one place and we can add/remove logic about handling config files without affecting the main module.
WDYT? I'm very open to other ideas as well as long as it removes the constant from main :)

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -248,8 +252,15 @@ def run(argv=None):
   variants |= 'FilterVariants' >> filter_variants.FilterVariants(
       reference_names=known_args.reference_names)
   if variant_merger:
-    variants |= (
-        'MergeVariants' >> merge_variants.MergeVariants(variant_merger))
+    if known_args.optimize_for_large_inputs:
+      partitions = variants | 'PartitionVariants' >> beam.Partition(
```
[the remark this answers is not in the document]
arostamianfar: The main point is to increase readability as it gives a "name" and "context" for what this block of code is doing, but feel free to leave as is for now since it's pretty small :)

setup.py
samanvp: Thanks for clarifying.

gcp_variant_transforms/transforms/partition_variants.py
```diff
@@ -0,0 +1,82 @@
...
+    Args:
+      variant - the variant to be partitioned
+      partition_no - the total number of partitions (> _RESERVED_PARTITIONS)
+    Yields:
+      An integer in the range of [0, partition_no)
+    Raises:
+      ValueError: If partition_no <= _RESERVED_PARTITIONS
+    """
+
+    chrom = (variant.reference_name).strip().lower()
```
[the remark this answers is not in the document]
samanvp: I don't think so, I thought it will make it more readable, now that I look at it I don't see how this is more readable :)

samanvp:
Following up on your comment of refactoring the partitioning logic from the actual Beam.partition_for() I added a new class to this PR. I was planning to add this new class in a follow up PR which implements reading the partition_config.yaml file as well as applying to to the pipeline.
To avoid redundant work, I added that class to this PR along with all of its unit tests. However, the logic of actually utilizing a config_file (through a flag) and exporting the final output to multiple BQ table is not added to this PR and will be send out in a future one.

gcp_variant_transforms/transforms/partition_variants_test.py
```diff
@@ -0,0 +1,121 @@
...
+    variants.append(vcfio.Variant(reference_name='6', start=0))
+    variants.append(vcfio.Variant(reference_name='06', start=0))
+    variants.append(vcfio.Variant(reference_name='chr06', start=0))
+    # Partition 9
+    variants.append(vcfio.Variant(reference_name='chr10', start=0))
+    variants.append(vcfio.Variant(reference_name='  chr10   ', start=0))
+    # Partition 21
+    variants.append(vcfio.Variant(reference_name='chr22', start=0))
+    variants.append(vcfio.Variant(reference_name='chr22    ', start=0))
+    return variants, variants[0:2], variants[2:4], variants[4:6],\
```
arostamianfar: Instead of this, I think it's much more readable to return a dict where the key is the partition number (int) and the value is a list of variants that should map to that partition. You can just return that dict both here and in non-standard case.
In your test, you would essentially pass the dict values as inputs to the pipeline, and then loop through [0, get_num_partitions) and if the key is in the dict, make sure that it matches the value, if not make sure it's empty. WDYT?
P.S. Is checking the 'label' important? If so, I suppose you can have a tuple of partition number and label as the key to your dict.

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -0,0 +1,91 @@
...
+
+VariantPartition class basically returns an index for a given (
+reference_name, pos) pair. The main utilization of this class is in
+partition_for() function used by DataFlow pipeline.
+"""
+
+from __future__ import absolute_import
+
+import re
+from mmh3 import hash  # pylint: disable=no-name-in-module,redefined-builtin
```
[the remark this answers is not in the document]
arostamianfar: huh intersting!

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -0,0 +1,92 @@
...
+          '_DEFAULT_NUM_PARTITIONS must be > _RESERVED_AUTO_PARTITIONS')
+    self._reference_name_to_partition_map = {}
+
+  def get_num_partitions(self):
+    """Returns the number of partitions."""
+    return _DEFAULT_NUM_PARTITIONS
+
+  def get_partition(self, reference_name, pos=0):
+    reference_name = reference_name.strip().lower()
+    if ((not reference_name) or (pos < 0)):
```
arostamianfar: nit: no need for surrounding brackets (styleguide is picky about avoiding unnecessary surrounding brackets :) )
`if not reference or pos < 0:` should work.

gcp_variant_transforms/transforms/partition_variants.py
```diff
@@ -0,0 +1,33 @@
...
+
+class PartitionVariants(beam.PartitionFn):
+  """Partitions variants based on their reference_name."""
+
+  def __init__(self, partition):
+    # type: (variant_partition.VariantPartition) -> None
+    self._partition = partition
+
+  def partition_for(self, variant, num_partitions):
+    # type: (Variant, int) -> int
```
arostamianfar: nit: s/Variant/vcfio.Variant

gcp_variant_transforms/vcf_to_bq.py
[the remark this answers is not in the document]
arostamianfar: Sounds good!

gcp_variant_transforms/transforms/partition_variants_test.py
```diff
@@ -0,0 +1,130 @@
...
+    # Partition 26
+    variants.append(vcfio.Variant(reference_name='NOTc4', start=0))
+    variants.append(vcfio.Variant(reference_name='chr0', start=0))
+    variants.append(vcfio.Variant(reference_name='chrY', start=0))
+    expected_partitions[26] = variants[9:12]
+    # The rest of partitions will be empty
+    expected_partitions[23] = []
+    return variants, expected_partitions
+
+  def test_partition_variants(self):
```
arostamianfar: This is still a bit difficult to read; consider the following approach using some python magic :D
```
expected_partitions = {}
expected_partitions[0] = [vcfio.Variant(reference_name='chr1', start=0),
                          vcfio.Variant(reference_name='CHR1', start=0)]
...etc
# do not add empty partitions as it can be error prone (e.g. what if you change the default #?)
return expected_partitions
```
In your test then:
```
variant_partitions = _get_standard_variant_partitions()
variant_partitions.update(_get_non_standard_variant_partitions())
variants = [variant for variant_list in variant_partitions for variant in variant_list]  # essentially flattens the lists
...
pipeline | Create(variants) 
...
for i in xrange(partitioner.get_num_partitions()):
  assert_that(partitions[i], equal_to(variant_partitions.get(i, [])), label=str(i))

```

samanvp: done.

gcp_variant_transforms/transforms/partition_variants_test.py
```diff
@@ -0,0 +1,146 @@
...
+                               vcfio.Variant(reference_name='chrX', start=0),
+                               vcfio.Variant(reference_name='chrM', start=0)]
+    # Partition 26
+    expected_partitions[26] = [vcfio.Variant(reference_name='NOTc4', start=0),
+                               vcfio.Variant(reference_name='chr0', start=0),
+                               vcfio.Variant(reference_name='chrY', start=0)]
+    return expected_partitions
+
+  def test_partition_variants(self):
+    standard_variants = self._get_standard_variants()
```
arostamianfar: you dont actually need these extra variants; can just get them from `expected_partitions.values()`
you just need to 'flatten' them, which you can do with that for loop i had mentioned previously :)

PR #204, opened 2018-04-23
Add a link to Color blog post to README.

arostamianfar:
Per PM's request :) linking to the blog post directly in README. This will give a better idea to users about what VT can do.

Longer term, we can have a "news" or "blogs" section when we have multiple blog posts :)


arostamianfar:
Done. PTAL

PR #207, opened 2018-04-24
Replace HeaderFields with VcfHeader.

### LINKED ISSUE #111: Remove HeaderFields in favor of VcfHeader
We currently have both [HeaderFields](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/libs/vcf_header_parser.py#L32)  and [VcfHeader](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/beam_io/vcf_header_io.py#L34).

HeaderFields is an older object that was written prior to having VcfHeader. We should remove HeaderFields in favor of VcfHeader.

arostamianfar:
Note that VcfHeader itself would be replaced with Nucleus' proto later on (which will make referencing fields like `num`, `desc`, and `type` much more native rather than a dict lookup). This PR makes that transition easier so that at least we have a unified representation of headers throughout the code.

Fixes #111.

Tested:
unit + integration tests.

### DIFF
--- a/gcp_variant_transforms/beam_io/vcf_header_io.py
+++ b/gcp_variant_transforms/beam_io/vcf_header_io.py
@@ -65,18 +65,17 @@ class VcfHeader(object):
               ):
     """Initializes a VcfHeader object.
 
+    It keeps the order of values in the input dictionaries. Order is important
+    in some fields like `contigs` and for ensuring order is unchanged in
+    VCF->VCF pipelines.
+
     Args:
-      infos: A dictionary mapping info keys (str) to vcf info metadata
-        values (:class:`~vcf.parser._Info`).
-      filters: A dictionary mapping filter keys (str) to vcf filter
-        metadata values (:class:`~vcf.parser._Filter`).
-      alts: A dictionary mapping alt keys (str) to vcf alt metadata
-        values (:class:`~vcf.parser._Alt`).
-      formats: A dictionary mapping format keys (str) to vcf format
-        metadata values (:class:`~vcf.parser._Format`).
-      contigs: A dictionary mapping contig keys (str) to vcf contig
-        metadata values (:class:`~vcf.parser._Contig`).
-      file_name: A str representing the file name of the vcf file.
+      infos: A dictionary mapping info keys to vcf info metadata values.
+      filters: A dictionary mapping filter keys to vcf filter metadata values.
+      alts: A dictionary mapping alt keys to vcf alt metadata values.
+      formats: A dictionary mapping format keys to vcf format metadata values.
+      contigs: A dictionary mapping contig keys to vcf contig metadata values.
+      file_name: The file name of the vcf file.
     """
     self.infos = self._values_asdict(infos or {})
     self.filters = self._values_asdict(filters or {})
@@ -100,11 +99,11 @@ class VcfHeader(object):
                                                  self.contigs]])
 
   def _values_asdict(self, header):
-    """Converts PyVCF header values to dictionaries."""
+    """Converts PyVCF header values to ordered dictionaries."""
     ordered_dict = OrderedDict()
     for key in header:
       # These methods were not designed to be protected. They start with an
-      # underscore to avoid confilcts with field names. For more info, see
+      # underscore to avoid conflicts with field names. For more info, see
       # https://docs.python.org/2/library/collections.html#collections.namedtuple
       ordered_dict[key] = header[key]._asdict()  # pylint: disable=W0212
     return ordered_dict
--- a/gcp_variant_transforms/beam_io/vcf_header_io.py
+++ b/gcp_variant_transforms/beam_io/vcf_header_io.py
@@ -16,7 +16,9 @@
 
 from __future__ import absolute_import
 
+from collections import OrderedDict
 from functools import partial
+from typing import Dict  # pylint: disable=unused-import
 import vcf
 
 import apache_beam as beam
@@ -54,26 +56,27 @@ class VcfHeader(object):
   """Container for header data."""
 
   def __init__(self,
-               infos=None,
-               filters=None,
-               alts=None,
-               formats=None,
-               contigs=None,
-               file_name=None):
+               infos=None,  # type: Dict[str, OrderedDict[vcf.parser._Info]]
+               filters=None,  # type: Dict[str, OrderedDict[vcf.parser._Filter]]
+               alts=None,  # type: Dict[str, OrderedDict[vcf.parser._Alt]]
+               formats=None,  # type: Dict[str, OrderedDict[vcf.parser._Format]]
+               contigs=None,  # type: Dict[str, OrderedDict[vcf.parser._Contig]]
+               file_name=None  # type: str
+              ):
     """Initializes a VcfHeader object.
 
     Args:
-      infos (dict): A dictionary mapping info keys (str) to vcf info metadata
+      infos: A dictionary mapping info keys (str) to vcf info metadata
         values (:class:`~vcf.parser._Info`).
-      filters (dict): A dictionary mapping filter keys (str) to vcf filter
+      filters: A dictionary mapping filter keys (str) to vcf filter
         metadata values (:class:`~vcf.parser._Filter`).
-      alts (dict): A dictionary mapping alt keys (str) to vcf alt metadata
+      alts: A dictionary mapping alt keys (str) to vcf alt metadata
         values (:class:`~vcf.parser._Alt`).
-      formats (dict): A dictionary mapping format keys (str) to vcf format
+      formats: A dictionary mapping format keys (str) to vcf format
         metadata values (:class:`~vcf.parser._Format`).
-      contigs (dict): A dictionary mapping contig keys (str) to vcf contig
+      contigs: A dictionary mapping contig keys (str) to vcf contig
         metadata values (:class:`~vcf.parser._Contig`).
-      file_name (str): A str representing the file name of the vcf file.
+      file_name: A str representing the file name of the vcf file.
     """
     self.infos = self._values_asdict(infos or {})
     self.filters = self._values_asdict(filters or {})
@@ -98,11 +101,14 @@ class VcfHeader(object):
 
   def _values_asdict(self, header):
     """Converts PyVCF header values to dictionaries."""
+    ordered_dict = OrderedDict()
+    for key in header:
+      # These methods were not designed to be protected. They start with an
+      # underscore to avoid confilcts with field names. For more info, see
+      # https://docs.python.org/2/library/collections.html#collections.namedtuple
+      ordered_dict[key] = header[key]._asdict()  # pylint: disable=W0212
+    return ordered_dict
 
-    # These methods were not designed to be protected. They start with an
-    # underscore to avoid confilcts with field names. For more info, see below:
-    # https://docs.python.org/2/library/collections.html#collections.namedtuple
-    return {key: header[key]._asdict() for key in header}  # pylint: disable=W0212
 
 class VcfHeaderSource(filebasedsource.FileBasedSource):
   """A source for reading VCF file headers.
--- a/gcp_variant_transforms/libs/bigquery_util.py
+++ b/gcp_variant_transforms/libs/bigquery_util.py
@@ -73,7 +73,7 @@ _FALLBACK_FIELD_NAME_PREFIX = 'field_'
 
 
 def get_bigquery_sanitized_field_name(field_name):
-  # (str) -> str
+  # type: (str) -> str
   """Returns the sanitized field name according to BigQuery restrictions.
 
   BigQuery field names must follow `[a-zA-Z][a-zA-Z0-9_]*`. This method converts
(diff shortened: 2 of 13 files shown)


gcp_variant_transforms/libs/bigquery_vcf_schema.py
```diff
@@ -120,9 +122,11 @@ def generate_schema_from_header_fields(
       continue
     calls_record.fields.append(bigquery.TableFieldSchema(
         name=bigquery_util.get_bigquery_sanitized_field_name(key),
-        type=bigquery_util.get_bigquery_type_from_vcf_type(field.type),
-        mode=_get_bigquery_mode_from_vcf_num(field.num),
-        description=bigquery_util.get_bigquery_sanitized_field(field.desc)))
+        type=bigquery_util.get_bigquery_type_from_vcf_type(
+            field[_HeaderKeyConstants.TYPE]),
+        mode=_get_bigquery_mode_from_vcf_num(field[_HeaderKeyConstants.NUM]),
```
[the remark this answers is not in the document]
arostamianfar: Excellent suggestion! Done!

gcp_variant_transforms/libs/bigquery_vcf_schema_test.py
```diff
@@ -195,7 +195,7 @@ def test_variant_merger_modify_schema(self):
         ('I1', Info('I1', 1, 'String', 'desc', 'src', 'v')),
         ('IA', Info('IA', field_counts['A'], 'Integer', 'desc', 'src', 'v'))])
     formats = OrderedDict([('F1', Format('F1', 1, 'String', 'desc'))])
-    header_fields = vcf_header_parser.HeaderFields(infos, formats)
+    header_fields = vcf_header_io.VcfHeader(infos=infos, formats=formats)
```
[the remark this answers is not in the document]
arostamianfar: Good question. For the schema (and our code in general), we just care about INFO and FORMAT fields and ignore the rest of the fields. The rest are included as part of the 'representative header' for completeness (in case we end up using them at some point in the future and/or if they're used by the parser).

gcp_variant_transforms/libs/vcf_header_parser.py
```diff
@@ -16,32 +16,23 @@
...
     input_file (str): A string specifying the path to the representative VCF
-    file, i.e., the VCF file that contains a header representative of all VCF
-    files matching the input_pattern of the job. It can be local or remote (e.g.
-    on GCS).
+      file, i.e., the VCF file that contains a header representative of all VCF
+      files matching the input_pattern of the job. It can be local or
+      remote (e.g. on GCS).
   Returns:
-    ``HeaderFields`` specifying header info.
+    :class:`vcf_header_io.VcfHeader` specifying header info.
```
bashir2: Please remove the type.

gcp_variant_transforms/libs/vcf_header_parser.py
```diff
@@ -16,32 +16,23 @@
...
 
-__all__ = ['HeaderFields', 'get_vcf_headers']
-
-
-# Stores parsed header information.
-#   - infos: Stores all `##INFO` header lines. Type is a ``dict`` with values
-#     equal to ``namedtuple`` defined in ``vcf.parser._Info``.
-#   - formats: Stores all `##FORMAT` header lines. Type is a ``dict`` with
-#     values equal to ``namedtuple`` defined in ``vcf.parser._Formats``.
-HeaderFields = namedtuple('HeaderFields', ['infos', 'formats'])
```
bashir2: I am thinking that VcfHeader actually belongs to this file because (i) all that this module currently has, in some sense, is a factory for VcfHeader objects (i.e., get_vcf_headers); and (ii) the dependency of this module on something from beam_io seems a little odd and I think a more logical dependency order is to have VcfHeader defined here and beam_io.vcf_header_io depend on this module instead (so the opposite direction).

If you make that change, many of the other dependencies in libs/* modules on beam_io will also be gone (specially those that this PR has introduced). So in general the dependency chain is beam_io modules depending on libs, not the opposite. WDYT?

gcp_variant_transforms/libs/bigquery_vcf_schema.py
```diff
@@ -120,9 +122,12 @@ def generate_schema_from_header_fields(
       continue
     calls_record.fields.append(bigquery.TableFieldSchema(
         name=bigquery_util.get_bigquery_sanitized_field_name(key),
-        type=bigquery_util.get_bigquery_type_from_vcf_type(field.type),
-        mode=_get_bigquery_mode_from_vcf_num(field.num),
-        description=bigquery_util.get_bigquery_sanitized_field(field.desc)))
+        type=bigquery_util.get_bigquery_type_from_vcf_type(
+            field[_HeaderKeyConstants.TYPE]),
```
bashir2: Why is this change needed? Can't you continue to use field.type, field.num, etc.?

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -366,7 +372,7 @@ def __init__(self,
     for field in annotation_fields or []:
       if field not in header_fields.infos:
         raise ValueError('{} INFO not found in the header'.format(field))
-      header_desc = header_fields.infos[field].desc
+      header_desc = header_fields.infos[field][_HeaderKeyConstants.DESC]
```
bashir2: Also here and elsewhere (why the change)?

gcp_variant_transforms/libs/bigquery_vcf_schema.py
```diff
@@ -134,16 +139,13 @@ def generate_schema_from_header_fields(
...
+        mode=bigquery_util.get_bigquery_mode_from_vcf_num(
+            field[_HeaderKeyConstants.NUM]),
+        description=bigquery_util.get_bigquery_sanitized_field(
+            field[_HeaderKeyConstants.DESC])))
     info_keys.add(key)
   if variant_merger:
     variant_merger.modify_bigquery_schema(schema, info_keys)
   return schema
-
-def _get_bigquery_mode_from_vcf_num(vcf_num):
```
bashir2: Moving this function to another module seems to be independent of the scope of this change; is there any particular reason to do it in this very PR?

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -98,11 +101,14 @@ def __repr__(self):
 
   def _values_asdict(self, header):
     """Converts PyVCF header values to dictionaries."""
+    ordered_dict = OrderedDict()
```
bashir2: Can you please elaborate why the order is important? It seems non-trivial to me and if there is a specific reason probably it should be documented either in this function's docstring or (better) in __init__.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -54,26 +56,27 @@ class VcfHeader(object):
...
+               alts=None,  # type: Dict[str, OrderedDict[vcf.parser._Alt]]
+               formats=None,  # type: Dict[str, OrderedDict[vcf.parser._Format]]
+               contigs=None,  # type: Dict[str, OrderedDict[vcf.parser._Contig]]
+               file_name=None  # type: str
+              ):
     """Initializes a VcfHeader object.
 
     Args:
-      infos (dict): A dictionary mapping info keys (str) to vcf info metadata
+      infos: A dictionary mapping info keys (str) to vcf info metadata
```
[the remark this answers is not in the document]
arostamianfar: Done.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -98,11 +101,14 @@ def __repr__(self):
 
   def _values_asdict(self, header):
     """Converts PyVCF header values to dictionaries."""
```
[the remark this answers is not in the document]
arostamianfar: Added the ordered note for now. The type thing becomes ugly as it has arbitrary values.

gcp_variant_transforms/beam_io/vcf_header_io.py
arostamianfar: Sure, added a comment in `__init__`. Essentially, we get ordered dicts from pyvcf and we should continue to keep the order. It's important for fields like contigs and for cases where we want to go from VCF->VCF. Nuclues' [proto](https://github.com/google/nucleus/blob/master/nucleus/protos/variants.proto#L161) is also ordered for the same reason.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -98,11 +101,14 @@ def __repr__(self):
...
+    ordered_dict = OrderedDict()
+    for key in header:
+      # These methods were not designed to be protected. They start with an
+      # underscore to avoid confilcts with field names. For more info, see
```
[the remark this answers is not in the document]
arostamianfar: Done. Thanks!

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -73,6 +73,7 @@ class TableFieldConstants(object):
 
 
 def get_bigquery_sanitized_field_name(field_name):
+  # (str) -> str
```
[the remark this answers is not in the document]
arostamianfar: oops...Done! Thanks! I guess InteliJ just treated this as a random comment...lol

gcp_variant_transforms/libs/bigquery_vcf_schema.py
arostamianfar: Unlike HeaderFields that used the namedtuples from pyvcf directly, we convert them to dicts in VcfHeader (there was an issue with picking namedtuples, so we had to convert them to dicts). So, the field no longer has a `num` attribute, but just a key. This will be nicer when we switch to using Nucleus protos as those have properly named fields.

gcp_variant_transforms/libs/bigquery_vcf_schema.py
arostamianfar: Allie actually suggested it in this PR; it was a small enough change (and refactoring related), so I decided to do it in this PR. I can change it in another PR if you feel strongly about keeping the scope smaller.

gcp_variant_transforms/libs/vcf_header_parser.py
arostamianfar: I actually think I/O modules should have the type in themselves (e.g. VCF I/O has Variant records). In any case, we're actually planning to remove VcfHeader altogether in favor of Nucleus [protos](https://github.com/google/nucleus/blob/master/nucleus/protos/variants.proto#L161), so it doesn't really matter at this point :)

gcp_variant_transforms/libs/vcf_header_parser.py
arostamianfar: Done.

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -18,13 +18,14 @@
...
+from gcp_variant_transforms.beam_io import vcf_header_io  # pylint: disable=unused-import
 from gcp_variant_transforms.libs import bigquery_row_generator
-from gcp_variant_transforms.libs import bigquery_schema_descriptor  #pylint: disable=unused-import
+from gcp_variant_transforms.libs import bigquery_schema_descriptor  # pylint: disable=unused-import
 from gcp_variant_transforms.libs import bigquery_vcf_schema
 from gcp_variant_transforms.libs import processed_variant
 from gcp_variant_transforms.libs import vcf_field_conflict_resolver
-from gcp_variant_transforms.libs import vcf_header_parser  #pylint: disable=unused-import
-from gcp_variant_transforms.libs.variant_merge import variant_merge_strategy  #pylint: disable=unused-import
+from gcp_variant_transforms.libs import vcf_header_parser  # pylint: disable=unused-import
```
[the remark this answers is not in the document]
arostamianfar: Good catch! Removed. That's one challenge with using unused-import :/ Hopefully pylint will fix this soon.

gcp_variant_transforms/libs/vcf_header_parser.py
bashir2: Well, you can argue that this module (`vcf_header_parser`) is actually an I/O module as well. In fact, both this module and the beam module use `vcf` which is the real I/O module IMO.

As a general guideline, it is important to pay attention to the "flow" of dependencies between packages and I have found that having Beam/Flume related functionality depend on more pure libraries is the right direction. Anyways, up to you, specially if you feel that the Nucleus change will happen soon, I am okay either way.

gcp_variant_transforms/beam_io/vcf_header_io.py
arostamianfar: Good point about merging: that's a different beast that we need to tackle when we have VCF->VCF. For now, we have a reasonable logic to continue to keep the order if no new fields are added. However, the semantics of VcfHeader should have order and it would pave the way towards Nucleus support.
RE asserts: that is a particular test where we're inferring headers, hence no predefined order. We ensure order is kept in all other tests.

Issue #183, opened 2018-04-12
Pass representative header directly to the parser


arostamianfar:
Currently, we use each file's VCF header to parse its contents, but we can also use the representative header to parse the INFO/FORMAT fields. Note that we should still use each file's headers for some of the other fields (especially the header that contains the same names; others like 'contigs' may also be important to keep). This allows us to handle any VCF file in a robust manner.

PR #208, opened 2018-04-24
Make VCF parser reads representative header.

nmousavi:
Fixes issue #183 

Tested:
  unit test
  presubmit


gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -720,7 +728,9 @@ def _store_header_lines(self, header_lines):
     def _create_generator(self):
       header_processed = False
       for text_line in self._text_lines:
-        if not header_processed and self._header_lines:
+        if (not self._representative_headers_given and
```
arostamianfar: You'd need to parse the last line of the header from each file separately (it contains the sample IDs, which is unique per file). I think the easiest approach is to modify `store_header_lines` to modify self._header_lines accordingly using the `representative_header_lines` variable. You'd then leave this logic unchanged.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -891,6 +902,8 @@ def __init__(
     Args:
       file_pattern (str): The file path to read from either as a single file or
         a glob pattern.
+      representative_headers(): Header definitions to be used for parsing VCF
```
arostamianfar: s/`()`/ `list(str)` (or just add type annotations to each line and remove the inline types).

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -932,13 +948,16 @@ class ReadAllFromVcf(PTransform):
...
       self,
+      representative_headers=None,
       desired_bundle_size=DEFAULT_DESIRED_BUNDLE_SIZE,
       compression_type=CompressionTypes.AUTO,
       allow_malformed_records=False,
       **kwargs):
     """Initialize the :class:`ReadAllFromVcf` transform.
 
     Args:
+      representative_headers(): Header definitions to be used for parsing VCF
```
arostamianfar: ditto here.

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -445,6 +448,35 @@ def test_info_numbers_and_types(self):
     self.assertEqual(2, len(read_data))
     self._assert_variants_equal([variant_1, variant_2], read_data)
 
+  def test_use_of_representative_header(self):
```
arostamianfar: Could you please add a test containing multiple files (sample1 defined in one file, sample2 in another). This tests that we don't end up with the same sample in all files.

gcp_variant_transforms/vcf_to_bq_common.py
```diff
@@ -83,19 +83,41 @@ def get_pipeline_mode(known_args):
     return PipelineModes.MEDIUM
   return PipelineModes.SMALL
 
+def _get_file_content(input_file):
```
arostamianfar: We should actually only read the lines from the header (it's possible for someone to just use a regular VCF file as 'representative'). This logic is actually in [vcf_header_parser](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/libs/vcf_header_parser.py), so we can reuse that code (with some refactoring) after #207 is merged?

gcp_variant_transforms/libs/vcf_header_parser.py
```diff
@@ -55,6 +55,22 @@ def get_vcf_headers(input_file):
   return HeaderFields(vcf_reader.infos, vcf_reader.formats)
 
 
+def get_header_lines(input_file):
```
arostamianfar: nit: s/get_header_lines/get_metadata_header_lines
and clarify that it only returns lines starting with `##` and not `#`

PR #210, opened 2018-04-25
Resolved some of the robustness related TODOs in vep_runner, namely:

bashir2:
- Added retry for server connection failures
- Added argument validation for annotation arguments
- Replaced a confusing test pattern and added more unit-tests
- Added support for --num_workers

Tested: unit-tests and manual run for annotating a set of files

Issue: #81


gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -203,20 +206,22 @@ def _process_pipeline_args(self, pipeline_args):
...
+        flags_dict, 'max_num_workers', 'num_workers')
     if self._max_num_workers <= 0:
       raise ValueError(
-          '--max_num_workers should be a positive number, got: {}'.format(
-              self._max_num_workers))
+          '--max_num_workers and --num_workers should be positive numbers, '
+          'got: {}'.format(self._max_num_workers))
 
-  def _get_flag(self, pipeline_flags, flag):
+  def _get_flag(self, pipeline_flags, *expected_flags):
```
arostamianfar: nit: any reason for not using a list of strings instead of arbitrary num args?

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -228,12 +233,12 @@ def wait_until_done(self):
...
-    # log messages of googleapiclient.discovery module for the next call of the
-    # API since they flood the log file.
+    # TODO(bashir2): Silence the log messages of googleapiclient.discovery
+    # module for the next call of the API since they flood the log file.
     # pylint: disable=no-member
-    request = self._pipeline_service.projects().operations().get(name=operation)
-    is_done = request.execute()['done']
+    request = self._pipeline_service.projects().operations().get(
+        name=operation)
+    is_done = request.execute(num_retries=_NUMBER_OF_API_CALL_RETRIES)['done']
```
arostamianfar: Curious: were you actually seeing errors in this case that you needed retry? Asking b/c we don't do this for our integration tests and it seems like it has been fine so far, but I wonder whether we should add it if you are seeing errors in v2.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -208,9 +212,31 @@ def add_arguments(self, parser):
               'core machine using two processes should help interleaving I/O '
               'vs CPU bound work.'))
 
-    # TODO(bashir2): Add validate() to check --vep_* arguments are sound when
-    # --run_annotation_pipeline is set (for example --annotation_output_dir
-    # should start with gs://).
+  def validate(self, parsed_args):
+    # type: (argparse.Namespace) -> None
+    args_dict = vars(parsed_args)
+    if args_dict[AnnotationOptions._RUN_FLAG]:
```
arostamianfar: You should be able to just reference `parsed_args.run_annotation_pipeline` natively here without creating the constants. See `BigQueryWriteOptions` validate method for an example.

gcp_variant_transforms/options/variant_transform_options_test.py
```diff
@@ -62,3 +66,51 @@ def test_dataset_does_not_exists(self):
...
         response={'status': '404'}, url='', content='')
     self.assertRaises(ValueError, self.options.validate, args, client)
+
+
+class AnnotationOptionsTest(unittest.TestCase):
+
+  def setUp(self):
+    self._options = variant_transform_options.AnnotationOptions()
+
+  def make_args(self, args):
```
arostamianfar: nit: s/make_args/_make_args

gcp_variant_transforms/vcf_to_bq_common.py
```diff
@@ -55,6 +55,8 @@ def parse_args(argv, command_line_options):
       specifies the options that will be added to parser.
   """
   parser = argparse.ArgumentParser()
+  # TODO(bashir2): The next line is not needed; instead of type='bool' use
```
arostamianfar: We actually defined this as it's more robust than the default one (i.e. works with both `True` and `true`; common in python as people are used to "True"). Please continue to use it unless there is another issue about it that I'm unaware of.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
bashir2: Slightly nicer format when calling this with just one flag (which is in most cases). I changed it to a list per your comment.

As a side: I am not sure how much all these list construction/slicing that we do impact performance of a Python code. I have been doing it fairly liberally but not sure if that is good or not. In other languages like C++ or Java, I would have thought more about it specially because the language's syntax does not make these as easy as Python usually does and performance in general is more centeric than Python. Thoughts?

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
bashir2: I am not sure if this is a V2 specific issue (vs a general possibility when talking to external services) but I saw it exactly here when we are polling whether the operations are done. That experience was particularly painful because I was running it on some big VCFs and it happened after a few hours (crashing VT). The exception message was something indicating a server connection error.

Re. integration tests: Given the high availability of our services, I don't expect that we see issues like this very often. In case of here and the polling because we keep doing this every 30 seconds for a few hours, it is more likely to happen.

gcp_variant_transforms/options/variant_transform_options.py
bashir2: Yes I know but I actually preferred this because I already have to declare these constants. The reason is that the static typing system has no idea about these flags (unlike say a named tuple) so for example if I have a typo in the name of the flag it is not different from the correct spelling (both are errors from the type checking perspective).

If you feel strongly about this, I can change it.

gcp_variant_transforms/options/variant_transform_options_test.py
bashir2: Done.

gcp_variant_transforms/vcf_to_bq_common.py
bashir2: Hmm okay but is it actually common that people provide value for boolean options? I thought the pattern is usually --boolean_flag with no value.

Anyway I removed the TODO per your comment. The reason for not liking it was that I should do the same wherever I want to parse arguments (e.g., in the new AnnotationOptionsTest).

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
bashir2: Okay, changed it back then :-)
And re. performance, yes definitely here is not a concern, I started thinking more about it when you told me about the '::' operator, but even there the performance concern was not significant compared to the much nicer code. I thought I bring it up as a general note to think about in coding/reviews.

gcp_variant_transforms/options/variant_transform_options.py
bashir2: I know the common pattern for argparse and have used it in many places, including access to these very flags _outside_ this module (check vep_runner for example). Here though, as I said, I have to define the constants anyways. So forget about the type checking; are you suggesting that I drop the AnnotationOptions constants (e.g., _RUN_FLAG, _VEP_*, etc.)? Please take a look at this whole function not just this particular line. I think the answer is "no", because I have to repeat those flag texts several times including inside raw strings (typos there are not going to be caught by anything, including unit-tests). Now if we agree to keep the constants, then I think it makes the code more understandable to use those constants everywhere in this module, including flag declarations and access to them.

If on the other hand, you feel we need to drop the constants too, then sure I repeat the strings everywhere (but I don't think that is a good choice).

arostamianfar: ok, sounds good. let's keep it in this case (I just don't want it to be a common pattern for all of our args, but seems like it would be isolated to just annotations for the reasons you mentioned).

PR #214, opened 2018-04-26
Format the report #178

allieychen:
The Report generated in the preprocessor is not easy to read. This pull request add some logic to format the report.

- Tested: Unit tests
- Issue: [178](https://github.com/googlegenomics/gcp-variant-transforms/issues/178)

### DIFF
--- a/gcp_variant_transforms/libs/preprocess_reporter.py
+++ b/gcp_variant_transforms/libs/preprocess_reporter.py
@@ -27,7 +27,6 @@ resource estimation.
 from typing import Dict, List, Union  # pylint: disable=unused-import
 
 from apache_beam.io.filesystems import FileSystems
-from apache_beam import pvalue
 
 from gcp_variant_transforms.beam_io import vcf_header_io
 from gcp_variant_transforms.beam_io.vcf_header_io import VcfHeader  # pylint: disable=unused-import
@@ -69,11 +68,10 @@ def generate_report(header_definitions,
   content_lines.extend(_generate_conflicting_headers_lines(
       _extract_conflicts(header_definitions.formats), resolved_headers.formats,
       vcf_header_io.HeaderTypeConstants.FORMAT))
-  if not isinstance(inferred_headers, pvalue.EmptySideInput):
-    content_lines.extend(_generate_inferred_headers_lines(
-        inferred_headers.infos, vcf_header_io.HeaderTypeConstants.INFO))
-    content_lines.extend(_generate_inferred_headers_lines(
-        inferred_headers.formats, vcf_header_io.HeaderTypeConstants.FORMAT))
+  content_lines.extend(_generate_inferred_headers_lines(
+      inferred_headers.infos, vcf_header_io.HeaderTypeConstants.INFO))
+  content_lines.extend(_generate_inferred_headers_lines(
+      inferred_headers.formats, vcf_header_io.HeaderTypeConstants.FORMAT))
   _write_to_report(content_lines, file_path)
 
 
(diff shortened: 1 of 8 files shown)


gcp_variant_transforms/libs/preprocess_reporter.py
```diff
@@ -59,12 +64,16 @@ def generate_conflicts_report(header_definitions,
...
+      _extract_conflicts(header_definitions.infos), resolved_headers.infos,
+      vcf_header_io.HeaderTypeConstants.INFO))
   content_lines.extend(_generate_conflicting_headers_lines(
-      _extract_conflicts(header_definitions.infos), resolved_headers.infos))
-  content_lines.extend(_generate_inferred_headers_lines(inferred_headers.infos))
-  content_lines.extend(_generate_inferred_headers_lines(
-      inferred_headers.formats))
+      _extract_conflicts(header_definitions.formats), resolved_headers.formats,
+      vcf_header_io.HeaderTypeConstants.FORMAT))
+  if not isinstance(inferred_headers, pvalue.EmptySideInput):
```
nmousavi: This check looks weird to me. 
1) Why do you need this check? What happen if empty obj is passed to `_generate_inferred_headers_lines`? Should it take care of it?

2) This is kind of ugly because we mentioned type of `inferred_headers` as VcfHeader, but we are checking for another type here.

gcp_variant_transforms/libs/preprocess_reporter.py
```diff
@@ -84,57 +93,77 @@ def _extract_conflicts(
...
+    resolved_headers,  # type: Dict[str, Dict[str, Union[str, int]]
+    category  # type: str
     ):
   # type: (...) -> List[str]
   """Returns the conflicting headers lines for the report.
 
-  The conflicting definitions, the file names and the resolutions are included
-  in the contents.
+  To better align contents in the report so that it can be viewed easily, for
+  every unique ID, generate several rows with the following steps.
```
nmousavi: generates?

gcp_variant_transforms/libs/preprocess_reporter.py
```diff
@@ -84,57 +93,77 @@ def _extract_conflicts(
...
+  - Step 1: The first row starts with The ID, the category('FOMRAT' or 'INFO'),
```
nmousavi: s/The/the?

gcp_variant_transforms/libs/preprocess_reporter.py
```diff
@@ -84,57 +93,77 @@ def _extract_conflicts(
...
+    one conflicting definition, one file name and the resolution, separated by
+    the ``_DELIMITER ``.
+  - Step 2: If there there are more files having the same definition as the
```
nmousavi: please remove duplicate there

gcp_variant_transforms/libs/preprocess_reporter.py
```diff
@@ -84,57 +93,77 @@ def _extract_conflicts(
...
-    resolved_headers  # type: Dict[str, Dict[str, Union[str, int]]
+    resolved_headers,  # type: Dict[str, Dict[str, Union[str, int]]
+    category  # type: str
     ):
   # type: (...) -> List[str]
   """Returns the conflicting headers lines for the report.
 
-  The conflicting definitions, the file names and the resolutions are included
-  in the contents.
+  To better align contents in the report so that it can be viewed easily, for
```
nmousavi: I found the comment below hard to read.

gcp_variant_transforms/libs/preprocess_reporter.py
```diff
@@ -84,57 +93,77 @@ def _extract_conflicts(
...
+               file_name,
+               _PADDING_CHARACTER]
+        content_lines.append(_DELIMITER.join(row))
   return content_lines
 
 
-def _generate_inferred_headers_lines(inferred_headers):
-  # type: (Dict[str, Dict[str, Union[str, int]]]) -> List[str]
+def _generate_inferred_headers_lines(inferred_headers, category):
+  # type: (Dict[str, Dict[str, Union[str, int]]], str) -> List[str]
```
nmousavi: should type for `inferred_headers` be `VcfHeader` as in `generate_report`?

gcp_variant_transforms/libs/preprocess_reporter.py
nmousavi: Yes, I think there is an option in beam that you can tell replace empty side input with default value.

PR #216, opened 2018-04-27
Fix Build Fail 490618bd-6fbe-4462-a46d-c80310e15d30

nmousavi:
readvariant() module uses rep header for parsing VCF files. Overwriting
rep header flag before writing value into it, gives a false signal that
this file exists. Pushed flag overwirte to after writing value into the file.

Tested:
   Build 8ffbec47-9664-4d57-9bb9-4010775c3c77


nmousavi:
Done!

PR #218, opened 2018-04-30
Fixing Build Breakage

samanvp:
Failure was caused by the mmh3 install in the docker image which was originally added in the
following commit:
https://github.com/googlegenomics/gcp-variant-transforms/commit/d79f0b798b56a456adc78b33aef604c1ee04aa8b

Search for the error message "x86_64-linux-gnu-gcc: error trying to exec 'cc1plus': execvp: No such file or directory" I found this solution, not 100% clear why it works.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,6 +20,9 @@ FROM google/cloud-sdk
 RUN mkdir -p /opt/gcp_variant_transforms/bin && mkdir -p /opt/gcp_variant_transforms/src
 ADD / /opt/gcp_variant_transforms/src/
 
+# Needed for installing mmh3 (one of the required packages in setup.py).
+RUN apt install -y g++
+
 # Install dependencies.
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -21,6 +21,7 @@ RUN mkdir -p /opt/gcp_variant_transforms/bin && mkdir -p /opt/gcp_variant_transf
 ADD / /opt/gcp_variant_transforms/src/
 
 # Install dependencies.
+RUN apt install -y python3-pip
```
nmousavi: Hmm, I am not sure if we need/should install pip with python3 support. Since it's a c++ issue, could you please try if updating gcc/g++ fixes it (`apt install g++`)?

samanvp: Installing g++ also fixes the problem.
Thanks!

nmousavi: Great, could you also comment this is needed for mmh3 install?

samanvp: done.

Issue #48, opened 2018-01-10
JSON ValueError when importing gnomAD variants


hannes-brt:
I am trying to import the gnomAD variants into BigQuery, but I consistently get an error that makes the pipeline fail on chr1. I was able to successfully import some other chromosomes (chr21 and chr10).

The error I'm getting is `ValueError: Out of range float values are not JSON compliant. NAN, INF and -INF values are not JSON compliant. [while running 'VariantToBigQuery/ConvertToBigQueryTableRow']`.

I tried the `--allow_malformed_records True` flag, but that made no difference.

The input file is from https://storage.googleapis.com/gnomad-public/release/2.0.2/vcf/genomes/gnomad.genomes.r2.0.2.sites.chr1.vcf.bgz and I decompressed it before running the pipeline.

My pipeline configuration is

```
name: gnomad-genomes-to-bigquery-pipeline
docker:
  imageName: gcr.io/gcp-variant-transforms/gcp-variant-transforms
  cmd: |
    ./opt/gcp_variant_transforms/bin/vcf_to_bq \
      --project XXX \
      --input_pattern gs://XXX/gnomad/vcf_decompressed/genomes/gnomad.genomes.r2.0.2.sites.chr1.vcf \
      --allow_malformed_records True \
      --output_table dg-platform:GnomAD.gnomad_genomes_chr1 \
      --staging_location gs://XXX/staging \
      --temp_location gs://XXX/temp \
      --job_name gnomad-genomes-to-bigquery-pipeline-chr1 \
      --runner DataflowRunner
```

The full log is
```
2018/01/10 04:34:27 I: Switching to status: pulling-image
2018/01/10 04:34:27 I: Calling SetOperationStatus(pulling-image)
2018/01/10 04:34:27 I: SetOperationStatus(pulling-image) succeeded
2018/01/10 04:34:27 I: Pulling image "gcr.io/gcp-variant-transforms/gcp-variant-transforms"
2018/01/10 04:35:20 I: Pulled image "gcr.io/gcp-variant-transforms/gcp-variant-transforms" successfully.
2018/01/10 04:35:20 I: Done copying files.
2018/01/10 04:35:20 I: Switching to status: running-docker
2018/01/10 04:35:20 I: Calling SetOperationStatus(running-docker)
2018/01/10 04:35:20 I: SetOperationStatus(running-docker) succeeded
[...]

PR #220, opened 2018-05-01
Fixes 'inf' conversion to integer issue.

### LINKED ISSUE #219: Proper handling of "inf" float values.
Context: [this thread](https://groups.google.com/d/topic/google-genomics-discuss/bdvpLwDzcJ8/discussion).

In [bigquery_util._get_bigquery_sanitized_float](https://github.com/googlegenomics/gcp-variant-transforms/blob/c4659bba2cf577d64f15db5cd9f477d9ea2b51b0/gcp_variant_transforms/libs/bigquery_util.py#L213) when a float value is `inf`, it is being converted to `maxint` which in turn causes [this ValueError exception](https://github.com/googlegenomics/gcp-variant-transforms/blob/c4659bba2cf577d64f15db5cd9f477d9ea2b51b0/gcp_variant_transforms/libs/bigquery_row_generator.py#L202).

This can be avoided by using `--allow_incompatible_records` which causes another cast to float but I think this is a regression that was introduced as an unintentional consequence of [this commit](https://github.com/googlegenomics/gcp-variant-transforms/pull/146/files#diff-674d337e04318e6ee1252a280a22b7f9L256).

In the gnomAD dataset, some of the float fields are "Infinity" and that is what triggered this investigation (note `float('inf') == float('Infinity')`)

bashir2:
Tested: updated unit-tests and ran integration tests

Fixes #219

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_row_generator_test.py
+++ b/gcp_variant_transforms/libs/bigquery_row_generator_test.py
@@ -85,6 +85,16 @@ class BigQueryRowGeneratorTest(unittest.TestCase):
         mode=TableFieldConstants.MODE_REPEATED,
         description='INFO foo desc'))
     schema.fields.append(bigquery.TableFieldSchema(
+        name='IF',
+        type=TableFieldConstants.TYPE_FLOAT,
+        mode=TableFieldConstants.MODE_NULLABLE,
+        description='INFO foo desc'))
+    schema.fields.append(bigquery.TableFieldSchema(
+        name='IF2',
+        type=TableFieldConstants.TYPE_FLOAT,
+        mode=TableFieldConstants.MODE_NULLABLE,
+        description='INFO foo desc'))
+    schema.fields.append(bigquery.TableFieldSchema(
         name='IFR',
         type=TableFieldConstants.TYPE_FLOAT,
         mode=TableFieldConstants.MODE_REPEATED,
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/libs/bigquery_row_generator_test.py
```diff
@@ -295,9 +305,9 @@ def test_nonstandard_float_values(self):
     variant = vcfio.Variant(
         reference_name='chr19', start=11, end=12, reference_bases='CT',
         alternate_bases=[], filters=[],
-        info={'II': vcfio.VariantInfo(float('inf'), '1'),
```
bashir2: @arostamianfar I am not sure what the original reason for these type mismatches were, i.e., II* fields are TYPE_INTEGER while the values here are float. With the changes in this PR, the original version of this test fails as expected and I changed these to TYPE_FLOAT because the case that I think needs to be covered is this (in particular the one based on gnomAD, reported in Issue #48 as well).

Please comment if the original version of this test was also based on some realistic cases that we want to cover.

arostamianfar: ah, thanks for fixing this! This was an oversight from my end not realizing that `float(maxint) != maxint`. I'm a bit uneasy about using float(maxint) as well as it becomes a strange value when loaded to BQ (`9.223372036854776e+18`). What do you think of using a smaller value that still makes `float(val) == val`?

bashir2: Sure, I can change `sys.maxint` to `sys.float_info.max`; that said, I generally avoid assigning the maximum value of a type to a variable of that type (in any programming language) because if it is unintentionally increased, it will cause overflows (and becomes `inf` again in Python); e.g., think about `a`, `b` being max_float and we do `a+b`. It is not a huge concern here but are you okay with say max_float/2 (or even max_float/10), I think all of those are huge enough not to create any confusion that these are arbitrary "INF" values.

arostamianfar: Sure. I'm ok with max_float/10. Thanks!

bashir2: Done.

nmousavi: 1) That's correct, field name should change to IF. 
2) Hmm, not sure why overflow is a problem? If a+b makes an overflow (i.e., `inf`) in python, it is actually not bad as these numbers intended to be 'inf' in python?  It won't be a problem if the result is going to be exported back to BQ as the value converted again to max_float. What is the scenario you think overflow creates an issue?

bashir2: Nima, re. 2: As we discussed off-PR, there is not a very particular scenario, specially in this case, that I am concerned about overflows but a general habit/experience that setting values to the maximum of their type is not great because it increases the chance of downstream overflows/errors for reasons I mentioned (this is more of a concern when in some languages the overflow appears as a negative number).

In any case, if you have strong preference for using max_float directly, I can change the code that way, I don't think this is a big deal either way.

PR #224, opened 2018-05-02
First version of the annotation document.

bashir2:
Issue #188


docs/variant_annotation.md
```diff
@@ -0,0 +1,140 @@
...
+![A sample CSQ schema](images/CSQ_schema.png)
+
+### Alternate Allele Matching
+
+As it can be seen in the above example, the alternate alleles listed in the
+`CSQ` field are different from `ALT` strings of the variant record (for example
+the `AAC->C` variant is represented by a single `-` which means a deletion).
+This is because, the gnomAD set is run through VEP with the
+[`--minimal`](http://useast.ensembl.org/info/docs/tools/vep/script/vep_options.html#opt_minimal)
+option. VT can recognize this, in fact, currently there are three modes of `ALT`
```
arostamianfar: nit: this is a bit hard to read; consider rephrasing with two separate sentences (e.g. something like "VT supports matching alt fields in minimal mode. There are actually three modes of ALT matching supported by VT:").

docs/variant_annotation.md
```diff
@@ -0,0 +1,140 @@
...
+#standardSQL
+SELECT reference_name, start_position, reference_bases, ALT.alt, CSQ.*
+FROM vcf_imports_external.gnomad_genomes_chr_hg19 AS T, T.alternate_bases AS ALT, ALT.CSQ AS CSQ
+WHERE CSQ.SYMBOL = "HBB" AND CSQ.IMPACT = "HIGH"
+ORDER BY start_position
+```
+
+and here is an snapshot from BigQuery UI for this query:
+
+![A sample query on the gnomAD genomes table](images/gnomAD_HBB_query.png)
```
arostamianfar: nit: would be better to have the actual query inlined as text so that it's both easier to read and enables copy/paste.

arostamianfar: ah, nvm :) did not realize that the query is actually just above the image.

docs/variant_annotation.md
bashir2: Done.

PR #225, opened 2018-05-02
Add malformed records to the report

allieychen:
Add the malformed records when parsing the VCF files to the report.

Tested: Unit tests.
Issue: [178](https://github.com/googlegenomics/gcp-variant-transforms/issues/178)

### DIFF
--- a/gcp_variant_transforms/vcf_to_bq_preprocess.py
+++ b/gcp_variant_transforms/vcf_to_bq_preprocess.py
@@ -91,7 +91,7 @@ def run(argv=None):
   known_args, pipeline_args = vcf_to_bq_common.parse_args(argv,
                                                           _COMMAND_LINE_OPTIONS)
   options = pipeline_options.PipelineOptions(pipeline_args)
-  pipeline_mode = vcf_to_bq_common.PipelineModes.MEDIUM
+  pipeline_mode = vcf_to_bq_common.get_pipeline_mode(known_args)
 
   with beam.Pipeline(options=options) as p:
     headers = vcf_to_bq_common.read_headers(p, pipeline_mode, known_args)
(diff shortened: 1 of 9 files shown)


gcp_variant_transforms/vcf_to_bq_preprocess.py
```diff
@@ -95,25 +89,20 @@ def run(argv=None):
...
-      known_args.allow_incompatible_records = True
-      variants = vcf_to_bq_common.read_variants(p, known_args)
-      malformed_records = variants | filter_variants.SiftMalformedVariants(
-          reference_names=known_args.reference_names)
-      inferred_headers, merged_headers = (_get_inferred_headers(
-          variants, known_args, merged_headers))
+    if known_args.report_all_conflicts:
+      variants = p | 'ReadFromVcf' >> vcfio.ReadFromVcf(
+          known_args.input_pattern, allow_malformed_records=True)
+      malformed_records = variants | filter_variants.ExtractMalformedVariants()
```
nmousavi: Should not you apply 'FilterVariants'  before extracting malformed variants?

allieychen: `FilterVariants` basically filter out those malformed records, while `ExtractMaformedVariants` keeps those malformed records. So I think these two steps are separate from each other.

nmousavi: We still want part of `FilterVariants` because it filters out not only  malformed records but also those with reference name we are not interested (using `_should_keep_reference_name`), right? I think we should ignore variants with reference name we are not interested in preprocessing tool too. If agree, maybe refactor fn that reads variant to do filtering unwanted variants as well. Then in preprocessing tool, we extract malformed variants while in VT we filter malformed variants.

allieychen: Yes, that is a good advice. Just discussed with Asha, it is fine we don't consider `reference_name` for now (this tool only takes `input_pattern`, `report_all_conflicts`, `report_path`, `resolved_headers_path` as the options). We may consider some filters after Saman's work.

PR #197, opened 2018-04-18
Fixing BigQuery "Too many sources provided" error

samanvp:
We observed during multiple runs of VT pipeline whenever we set the --num_workers to
high number (for example 256), it fails with the following error
message:

    BigQuery execution failed., Error:
    Message: Too many sources provided: xxxxx. Limit is 10000
In this change we implemented a solution there was proposed here:
https://stackoverflow.com/questions/44255924/error-message-too-many-sources-provided-15285-limit-is-10000

#199 
Tested:2 test run finished successfully, Platinum_10 and Plantinum_1000

PR #226, opened 2018-05-02
Another hack to avoid BigQuery write failure

### LINKED ISSUE #199: Beam sink to BigQuery failure: Too many sources provided
We observed during multiple runs of VT pipeline whenever we set the --num_workers to a high number (256 or higher), it fails with the following error message:

```
BigQuery execution failed., Error:
Message: Too many sources provided: xxxxx. Limit is 10000

```
After communicating with Beam team we were pointed to [this temporary solution](https://stackoverflow.com/questions/44255924/error-message-too-many-sources-provided-15285-limit-is-10000) and we were told:

> BQ sink failure to producing too many files (shards) is a known issue ... hoping to add a BQ custom sink to Python SDK which will fix this. Relevant JIRA is https://issues.apache.org/jira/browse/BEAM-2801.

samanvp:
Following hack proposed in #197 (to fix #199) we still observed BQ write failures.
This is a second attempt to deal with this problem until a principled
solution by DataFlow team will be ready:
https://issues.apache.org/jira/browse/BEAM-2801

### DIFF
--- a/gcp_variant_transforms/options/variant_transform_options.py
+++ b/gcp_variant_transforms/options/variant_transform_options.py
@@ -92,6 +92,8 @@ class VcfReadOptions(VariantTransformsOptions):
 class BigQueryWriteOptions(VariantTransformsOptions):
   """Options for writing Variant records to BigQuery."""
 
+  _APPEND_FLAG = 'append'
+
   def add_arguments(self, parser):
     parser.add_argument('--output_table',
                         required=True,
@@ -106,7 +108,7 @@ class BigQueryWriteOptions(VariantTransformsOptions):
               'easier, because it avoids having to map each field with the '
               'corresponding alternate record while querying.'))
     parser.add_argument(
-        '--append',
+        '--' + BigQueryWriteOptions._APPEND_FLAG,
         type='bool', default=False, nargs='?', const=True,
         help=('If true, existing records in output_table will not be '
               'overwritten. New records will be appended to those that '
@@ -124,11 +126,11 @@ class BigQueryWriteOptions(VariantTransformsOptions):
       raise ValueError(
           'Expected a table reference (PROJECT:DATASET.TABLE) '
           'instead of {}.'.format(parsed_args.output_table))
+    if not client:
+      credentials = GoogleCredentials.get_application_default().create_scoped(
+          ['https://www.googleapis.com/auth/bigquery'])
+      client = bigquery.BigqueryV2(credentials=credentials)
     try:
-      if not client:
-        credentials = GoogleCredentials.get_application_default().create_scoped(
-            ['https://www.googleapis.com/auth/bigquery'])
-        client = bigquery.BigqueryV2(credentials=credentials)
       client.datasets.Get(bigquery.BigqueryDatasetsGetRequest(
           projectId=output_table_re_match.group('project'),
           datasetId=output_table_re_match.group('dataset')))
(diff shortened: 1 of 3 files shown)


gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -113,7 +119,22 @@ def expand(self, pcoll):
             self._allow_incompatible_records,
             self._omit_empty_sample_calls))
     if self._limited_write:
-      bq_rows |= 'LimitWrite' >> limit_write.LimitWrite(_WRITE_SHARDS_LIMIT)
+      bq_row_partitions = bq_rows | beam.Partition(random_partition,
```
arostamianfar: I believe you can actually use a lambda function here:
`beam.Partition(lambda _, n: random.randint(0, n - 1), NUM_PATITIONS)`

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -113,7 +119,22 @@ def expand(self, pcoll):
...
+                                                   _NUM_RANDOM_PARTITIONS)
+      bq_writes = []
+      for i in range(_NUM_RANDOM_PARTITIONS):
+        bq_rows = (bq_row_partitions[i] | 'LimitWrite' + str(i) >>
+                   limit_write.LimitWrite(_WRITE_SHARDS_LIMIT))
+        bq_writes.append(bq_rows | 'WriteToBigQuery' + str(i) >>
```
arostamianfar: nit: I think the indentation would be nicer if you break after `append(` and then on `|`

gcp_variant_transforms/transforms/variant_to_bigquery.py
arostamianfar: Please add a comment about what this code is doing and why it's needed. Also, mention why we always append as well.

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -113,7 +119,22 @@ def expand(self, pcoll):
...
+                         beam.io.Write(beam.io.BigQuerySink(
+                             self._output_table,
+                             schema=self._schema,
+                             create_disposition=(
+                                 beam.io.BigQueryDisposition.CREATE_IF_NEEDED),
+                             write_disposition=(
+                                 beam.io.BigQueryDisposition.WRITE_APPEND))))
+      return bq_writes
+    # If _limit_write == False we don't need any hack to avoid the BQ failure.
```
arostamianfar: nit: I think you can remove this comment once you add the comment above.

arostamianfar: nit: may be better to explicitly have an 'else' statement here so that it's more clear it's one or the other.

samanvp: done.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -124,23 +124,41 @@ def validate(self, parsed_args, client=None):
...
-                         (output_table_re_match.group('project'),
-                          output_table_re_match.group('dataset')))
+                         (project_id, dataset_id))
       else:
         # For the rest of the errors, use BigQuery error message.
         raise
-
+    # Ensuring given output table doesn't already exist to avoid overwriting it.
+    args_dict = vars(parsed_args)
+    if not args_dict['append']:
```
arostamianfar: you can just do `if not parsed_args.append:`

samanvp: done.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -124,23 +124,41 @@ def validate(self, parsed_args, client=None):
...
         raise ValueError('Dataset %s:%s does not exist.' %
-                         (output_table_re_match.group('project'),
-                          output_table_re_match.group('dataset')))
+                         (project_id, dataset_id))
       else:
         # For the rest of the errors, use BigQuery error message.
         raise
-
+    # Ensuring given output table doesn't already exist to avoid overwriting it.
+    args_dict = vars(parsed_args)
```
arostamianfar: remove :)

samanvp: sorry, missed this.

gcp_variant_transforms/options/variant_transform_options_test.py
```diff
@@ -29,43 +28,61 @@
...
-
-
 class BigQueryWriteOptionsTest(unittest.TestCase):
   """Tests cases for the BigQueryWriteOptions class."""
 
   def setUp(self):
-    self.options = variant_transform_options.BigQueryWriteOptions()
+    self._options = variant_transform_options.BigQueryWriteOptions()
+
+  def _make_args(self, args):
```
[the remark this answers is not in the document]
samanvp: done.

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -113,14 +116,34 @@ def expand(self, pcoll):
...
+      # We split data into _NUM_BQ_LIMITED_WRITE_PARTITIONS random partitions
+      # and then write each part to final BQ by appending them together.
+      # Combined with LimitWrite transform, this will avoid the BQ failure.
+      bq_row_partitions = bq_rows | beam.Partition(
+          lambda _, n: random.randint(0, n - 1),
+          _NUM_BQ_LIMITED_WRITE_PARTITIONS)
+      bq_writes = []
+      for i in range(_NUM_BQ_LIMITED_WRITE_PARTITIONS):
+        bq_rows = (bq_row_partitions[i] | 'LimitWrite' + str(i) >>
+                   limit_write.LimitWrite(_WRITE_SHARDS_LIMIT))
```
bashir2: It seems that we are doing both partitioning and GroupByKey (in LimitWrite), is this still needed? I mean do we have any evidence that both of these are required and partitioning/append alone is not enough?

arostamianfar: yes, both are still needed. Otherwise, you can still get more than 10K files per partition. We believe doing partition+limit write is the most effective approach and it speeds up the pipeline a lot (you take advantage of distributing the groupbykey step among more machines and it also runs 20 bq load operations in parallel rather than 1). We were able to successfully load 30B records to BQ in ~3h (all previous attempts failed). I think Saman also ran with 1000g as well.

PR #233, opened 2018-05-08
Remove `optimize_for_large_inputs` for preprocessor tool (Issue 178)

allieychen:
Refactor the get pipeline mode, so that `optimize_for_large_inputs` is not necessary for preprocessor pipeline.

Issue: [178](https://github.com/googlegenomics/gcp-variant-transforms/issues/178)


allieychen:
This in essentially the same piece of code that was removed because of the Travis CI checking [error](https://travis-ci.org/googlegenomics/gcp-variant-transforms/builds/376423126). But now it works fine. It might be some issues related to Travis CI itself.

PR #234, opened 2018-05-09
Integration tests for running VEP and fixing another annotation alt matching edge case.

### LINKED ISSUE #185: Add more thorough integration tests for annotation fields
Essentially extend this one: https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/testing/integration/presubmit_tests/small_tests/valid_4_2_VEP.json

### LINKED ISSUE #223: Integration tests for vep_runner.
We need some integration tests that annotates input VCFs using the `vep_runner` module and compare the generated fields with expected values, preferably already stored in the same VCFs (under a different field). These will indirectly test our [VEP docker release](https://github.com/googlegenomics/variant-annotation/tree/master/batch/vep).

bashir2:
This also adds minor tweaks like choosing default machine_types which were added to force fewer flags set by the user and/or in integration tests.

Tested:
* `./deploy_and_run_tests.sh --run_presubmit_tests`
* New unit-test added for the fix

Fixes #223 
Fixes #185


gcp_variant_transforms/testing/integration/run_tests.py
```diff
@@ -92,7 +92,10 @@ def __init__(self,
             '--temp_location {}'.format(context.temp_location),
             '--job_name {}-{}'.format(test_name, dataset_id.replace('_', '-'))]
     for k, v in kwargs.iteritems():
-      args.append('--{} {}'.format(k, v))
+      if k == 'annotation_output_dir':
+        args.append('--{} {}/{}'.format(k, v, dataset_id))
```
arostamianfar: instead of special-casing this flag, consider adding "/{DATASET_ID}" to the string and always doing `.format(DATASET_ID=dataset_id)`.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -436,12 +435,16 @@ def add_annotation_data(self, proc_var, annotation_field_name, data):
...
           alt._info[annotation_field_name] = annotations_list
 
-  def _find_common_alt_ref_prefix(self, proc_var):
+  def _find_common_alt_ref_prefix_char(self, proc_var):
     # type: (ProcessedVariant) -> str
-    alt_list = [
-        alt.alternate_bases for alt in proc_var._alternate_datas]
-    alt_list.append(proc_var.reference_bases or '')
-    return os.path.commonprefix(alt_list)
+    if proc_var.reference_bases:
```
arostamianfar: I think it's more readable to do `if not proc_var.reference_bases: return ''` and then have the rest of the logic (it's strange how the first if doesn't return and then switches the control to the for loop below it). It also looks like you are either returning empty string or the first character of reference_bases, so, it may be better to special case all cases where you return empty string first, and then `return proc_var.reference_bases[0]` (this latter part is just a suggestion though)

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -85,6 +85,12 @@ def create_runner_and_update_args(known_args, pipeline_args):
     known_args.annotation_fields.append(known_args.vep_info_field)
   else:
     known_args.annotation_fields = [known_args.vep_info_field]
+  # TODO(bashir2): The VEP runner by default runs VEP with --allele_number hence
+  # we turn on this feature here. However, this might be inconsistent with other
+  # annotation fields that are originally present in input files, if they do not
+  # have ALLELE_NUM annotation. The fix is to make annotation ALT matching
+  # smarter to fall back on other matching methods if ALLELE_NUM is not present.
+  known_args.use_allele_num = True
```
arostamianfar: why not set the default value to true in the flag instead? setting this to true here essentially makes the flag a no-op?

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -436,12 +435,16 @@ def add_annotation_data(self, proc_var, annotation_field_name, data):
             self._add_ambiguous_fields(annotations_list, ambiguous)
           alt._info[annotation_field_name] = annotations_list
 
-  def _find_common_alt_ref_prefix(self, proc_var):
+  def _find_common_alt_ref_prefix_char(self, proc_var):
```
arostamianfar: What is the reasoning behind returning just a single char instead of the prefix? Please add a comment.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
bashir2: Hmm, why is setting it here makes the flag no-op? The reason the default is False is because we want the user be specific in telling us to use ALLELE_NUM; think about pre-annotated VCF scenarios (we discussed this when this feature was added). Here though is where we are running VEP ourselves and we run it with `--allele_number` hence setting the flag.

Once we implement the suggestion in the above TODO, maybe we can actually remove this flag and always look for ALLELE_NUM first. I added an extra note to the end of TODO accordingly but please let me know if I misunderstood your point.

gcp_variant_transforms/libs/processed_variant.py
bashir2: Done (the first part).

gcp_variant_transforms/libs/processed_variant.py
bashir2: Please check the updated comment in `_alt_matches_annotation_alt` (line 578 below). That is the right place to explain this as the function doc has examples with links to VEP documentation that describes the ALT mapping logic in details.

So the gist of it is that VCF spec says there should be only one extra base included before an indel variant region. Of course this is not followed in all VCFs and that is what triggered the issue in my new integration tests. In the new platinum file, there are examples that multiple extra bases are added to the variant region and VEP only drops the first base not the whole common part (in the non-minimal mode).

gcp_variant_transforms/testing/integration/run_tests.py
bashir2: Done; good suggestion.
Actually your comment reminded me that dataset_id is not enough and this should actually be table_id because if we have multiple tests running VEP then they clash in output dir name. Fixed now.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
arostamianfar: ah i see...that makes sense! thanks!

gcp_variant_transforms/libs/processed_variant.py
arostamianfar: I see. Thanks for the explanation. sg.

PR #236, opened 2018-05-09
Partition

samanvp:
Ability to partition output tables based on the partition_config.yaml file.


samanvp:
Hi Asha,
Just note I added a few minor styles edits.
Here is the state right now:
 * I checked the correctness of the code using the two config files for integration test (testing/data/misc/partition_config_integration1.yaml and 2) and manually checked the number of rows in the output tables and everything was correct. If you agree, I am going to add integration tests in a separate PR.
  * Output BQ tables still don't have labels as we intended in our design, that also will be added in a follow up PR.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -218,44 +219,92 @@ def run(argv=None):
       known_args.minimal_vep_alt_matching,
       counter_factory)
 
-  partitioner = variant_partition.VariantPartition()
+  if known_args.partition_config_path:
```
arostamianfar: I've been thinking about this and I think perhaps a combination of your original suggestion (about keeping a single variable named `variants` and passing that around) and some cleanup can be more readable. Something like:
```
create partitioner (can be None if not needed)
read VCF files and filter
if partitioner:
  num_partitions = partitioner.get_num_partitions()
  variants |= 'PartitionVariants' >> beam.Partition(
        partition_variants.PartitionVariants(partitioner), num_partitions)
  # does 'del variants[partitioner.get_residual_partition_index()]' work? if not we need a way to remove it from here.
  # perhaps by reconstructing another array that doesn't have that particular index?
else:
  num_partitions = 1
  variants = [variants]

for i in range(num_partitions):
  variants[i] |= merge
  variants[i] |= convert to processed_variant # add anything else that may come along later

if partitioner.should_flatten():  # based on whether there was a config file
  variants = [variants | 'FlattenPartitions' >> beam.Flatten()]
  num_partitions = 1

for i in range(num_partitions):
  output to BigQuery using table suffix based on i
```
We'd need to think about the merge step when/if we implement gVCF merging as we'd need to take care of regions that span multiple partitions (if not done by chromosome), but we can figure that out later.
Lemme know what you think!

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,26 +43,216 @@
...
+  """Partition variants based on their reference_name and/or position.
+
+  This class has 2 operating modes:
+    1) No config file is given (config_file_path == ''):
+        Automatically partition variants based on their reference_name.
+    2) A config file is given:
+        partitions variants based on input config file.
+  """
+
+  def __init__(self, config_file_path=''):
```
arostamianfar: nit: `=None` is better

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -90,3 +289,29 @@ def _get_auto_partition(self, reference_name):
       # Save partition in _reference_name_to_partition dict for future lookups
       self._reference_name_to_partition_map[reference_name] = partition
     return partition
+
+  def is_default_partition_absent(self):
```
arostamianfar: I think using "residual" here and everywhere is a better name instead of "default".

arostamianfar: It's more readable to have a positive method name rather than double-negative so `has_residual_partition` is a better name.

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,26 +43,216 @@
...
+
+  This class has 2 operating modes:
+    1) No config file is given (config_file_path == ''):
+        Automatically partition variants based on their reference_name.
+    2) A config file is given:
+        partitions variants based on input config file.
+  """
+
+  def __init__(self, config_file_path=''):
+    if not config_file_path:
```
arostamianfar: as previously discussed: it's much cleaner to have a 'state' with all of the parameters initialized to default values (e.g. `self._num_partitions = _DEFAULT_NUM_PARTITIONS`, `self._residual_partition_index = -1`, etc). Then, if config file is given, you'd update the state accordingly. It also makes the code much easier to follow, as currently it's not clear what is the state of the class when a config file is given (e.g. how is `_by_ref_name` different than `_reference_name_to_partition_map`? are they both really needed?).

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -377,3 +377,22 @@ def add_arguments(self, parser):
         help=('The full path of the resolved headers. The file will not be'
               'generated if unspecified. Otherwise, please provide a local '
               'path if run locally, or a cloud path if run on Dataflow.'))
+    parser.add_argument(
```
arostamianfar: why was this added? :)

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,26 +43,216 @@
...
+    def _parse_position(pos_str):
+      return int(pos_str.replace(',', ''))
+
+    with FileSystems.open(config_file_path, 'r') as f:
+      try:
+        partition_configs = yaml.load(f)
+      except yaml.YAMLError as e:
+        raise ValueError('Invalid yaml file: %s' % str(e))
+
+    if len(partition_configs) > _MAX_NUM_PARTITIONS:
```
arostamianfar: instead of adding ValueErrors throughout the actual logic, consider creating a validate() method that validates the content first. You can then go through the contents assuming it's valid. I think for some of them you still need to validate while parsing (e.g. checking for overlapping regions), but it would be simpler.

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,26 +43,216 @@
...
+    self._by_ref_name = defaultdict(RegionIndexer)
+
+    self._suffixes = []
+    table_suffix_duplicate = {}
+
+    # Default partition will contain all remaining readings that do not match to
+    # any other partition. -1 means it does not exist.
+    self._default_partition_index = -1
+    partition_index = 0
+    for partition_config in partition_configs:
```
arostamianfar: I think this logic can be simplified a lot. Let's discuss later on how we can do it.

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -90,3 +289,29 @@ def _get_auto_partition(self, reference_name):
       # Save partition in _reference_name_to_partition dict for future lookups
       self._reference_name_to_partition_map[reference_name] = partition
```
arostamianfar: Can we use the same structure for both auto and config file? i.e. map of chr to Region class.

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -90,3 +289,29 @@ def _get_auto_partition(self, reference_name):
...
+
+    # If default partition is absent in the config file, we still need an index
+    # for all residual variants. In this case we add a dummy paritition indexed
+    # _num_partitions which will be ignored right after the partitioning stage.
+    if self.is_default_partition_absent():
+      return self._num_partitions
+    else:
+      return self._default_partition_index
+
+  def get_suffix(self, index):
```
arostamianfar: get_partition_name?

gcp_variant_transforms/libs/variant_partition.py
samanvp: done.

gcp_variant_transforms/libs/variant_partition.py
samanvp: As we discussed I replaced _reference_name_to_partition_map and now in both modes we are using the same dict.

gcp_variant_transforms/libs/variant_partition.py
samanvp: done.
Note two of the ValueErrors are still remained in _parse_config, if we move those two to validate we will be basically implementing the whole parse logic in validate and only avoid recording the result. To make the distinction between the parse and validate, I left those two ValueErrors in parse. Please let me know what do you think.

gcp_variant_transforms/libs/variant_partition.py
samanvp: I tried to simplify the logic, let me know what you think.

gcp_variant_transforms/libs/variant_partition.py
samanvp: done.

gcp_variant_transforms/options/variant_transform_options.py
samanvp: Sorry, I somehow messed up while syncing.

gcp_variant_transforms/vcf_to_bq.py
samanvp: I implemented this and code now looks much nicer, please take a look and let me know what you think.
Meanwhile I have two test runs and I will add the integration tests.

gcp_variant_transforms/testing/data/misc/partition_config1.yaml
```diff
@@ -0,0 +1,37 @@
...
+       - "chr1:1,000,000-2,000,000"
+-  partition:
+     partition_name: "chr01_part3"
+     regions:
+       - "chr1:2,000,000-999,999,999"
+-  partition:
+     partition_name: "chrom02"
+     regions:
+       - "chr2"
+       - "ch2"
```
arostamianfar: please avoid using "ch2" and "c2" even in tests as they're not 'standard' and can confuse people looking at this test :) You can choose one of the real chromosome names from hg19.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -218,44 +219,74 @@ def run(argv=None):
...
   pipeline = beam.Pipeline(options=beam_pipeline_options)
   variants = _read_variants(pipeline, known_args)
   variants |= 'FilterVariants' >> filter_variants.FilterVariants(
       reference_names=known_args.reference_names)
+  if partitioner:
+    num_partitions = partitioner.get_num_partitions()
+    variants = variants | 'PartitionVariants' >> beam.Partition(
+        partition_variants.PartitionVariants(partitioner), num_partitions)
+    if (not partitioner.should_flatten() and
+        not partitioner.has_residual_partition()):
```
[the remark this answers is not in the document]
samanvp: No, unfortunately len(variants) does not work and we need to know the number of partitions. So we need this if statement anyways to decide whether or not we need the num_partitions -= 1.

For now I added the following loop:
```variants = [variants[i] for i in range(0, num_partitions - 1)]```

Let me know what you think.

File: gcp_variant_transforms/vcf_to_bq.py [hunk 12]
```diff
@@ -218,44 +219,74 @@ def run(argv=None):
...
-       'VariantToBigQuery' >> variant_to_bigquery.VariantToBigQuery(
-           known_args.output_table,
-           header_fields,
-           variant_merger,
-           processed_variant_factory,
-           append=known_args.append,
-           allow_incompatible_records=known_args.allow_incompatible_records,
-           omit_empty_sample_calls=known_args.omit_empty_sample_calls,
-           num_bigquery_write_shards=known_args.num_bigquery_write_shards))
+    variants = apply_transform(
```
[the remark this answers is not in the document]
samanvp: This is what I came up with, after dropping apply_transform function:
```
  proc_variants = []
  for i in range(num_partitions):
    if variant_merger:
      merged = variants[i] | ('MergeVariants' + str(i) >>
                              merge_variants.MergeVariants(variant_merger))
    else:
      merged = variants[i]
    proc_variants.append(
        merged | 'ProcessVaraints' >>
        beam.Map(processed_variant_factory.create_processed_variant).\
            with_output_types(processed_variant.ProcessedVariant))
  
  if partitioner.should_flatten():
    proc_variants = [proc_variants | 'FlattenPartitions' >> beam.Flatten()]
    num_partitions = 1

  for i in range(num_partitions):
    partition_name = partitioner.get_partition_name(i)
    _ = (proc_variants[i] | 'VariantToBigQuery' + partition_name >>
         variant_to_bigquery.VariantToBigQuery(
             known_args.output_table + partition_name,
             header_fields,
             variant_merger,
             processed_variant_factory,
             append=known_args.append,
             allow_incompatible_records=known_args.allow_incompatible_records,
             omit_empty_sample_calls=known_args.omit_empty_sample_calls,
             num_bigquery_write_shards=known_args.num_bigquery_write_shards))

```
If you think this looks cleaner, I will go ahead and apply the change, please let me know.

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +43,295 @@
...
+
+  def get_partition_name(self, index):
+    if self._config_file_path_given:
+      if index >= self._num_partitions or index < 0:
+        raise ValueError('Partition index is outside of expected range: {}'.
+                         format(index))
+      # Note, a '_' is automatically added, so output table name would be:
+      #   known_args.output_table + '_' + partition_name
+      return '_' + self._partition_names[index]
+    else:
```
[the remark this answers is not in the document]
samanvp: How about returning empty string for all else cases? This will make the  vcf_to_bq.py look cleaner (in the loop where we read the partition_name and call variant_to_bigquery.VariantToBigQuery.

Let me know if you think this is reasonable.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -218,44 +219,73 @@ def run(argv=None):
...
+    if partitioner.should_flatten():
+      variants = [variants | 'FlattenPartitions' >> beam.Flatten()]
+      num_partitions = 1
+  proc_variants = apply_transform(
+      variants, num_partitions, 'ProcessVaraints',
+      beam.Map(
+          processed_variant_factory.create_processed_variant).with_output_types(
+              processed_variant.ProcessedVariant))
+  for i in range(num_partitions):
+    partition_name = partitioner.get_partition_name(i)
```
arostamianfar: Also need to take into account when partitioner is None. With the previous suggestion about moving the `_` to here, I suggest something like:
```
table_suffix = '_' + partitioner.get_partition_name(i) if partitioner else ''
```

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,262 @@
...
           '_DEFAULT_NUM_PARTITIONS must be > _RESERVED_AUTO_PARTITIONS')
-    self._reference_name_to_partition_map = {}
+    self._num_partitions = _DEFAULT_NUM_PARTITIONS
+    # This variable determines the operation mode auto (default mode) vs config.
+    self._config_file_path_given = False
+    # Residual partition will contain all remaining readings that do not match
+    # to any other partition. -1 means it does not exist.
+    self._residual_partition_index = -1
+    self._ref_name_to_partitions_map = defaultdict(_ChromosomePartitioner)
+    self._partition_names = []
```
arostamianfar: nit: I think this is more robust and intuitive if it's a dict of int->str (where key is the index). Just so that it's not tied to the particular order of the for loop (even though it is now).

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,262 @@
...
 
-class VariantPartition(object):
-  """Automatically partition variants based on their reference_name."""
+
+class _ChromosomePartitioner(object):
+  """Assigns partition indices to multiple regions inside a chromosome.
+
+  This class logic is implemented using an interval tree, each region is
+  considered as an interval and will be added to the interval tree. Note all
+  regions must be pairwise disjoint, ie no overlapping interval is accepted.
```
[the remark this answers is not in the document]
samanvp: done.

gcp_variant_transforms/vcf_to_bq.py
samanvp: done (except that I used a normal, easier to read if statement).

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,262 @@
...
+      return self._num_partitions - 1
+
+  def get_partition_name(self, partition_index):
+    if self._config_file_path_given:
+      if partition_index >= self._num_partitions or partition_index < 0:
+        raise ValueError('Partition index is outside of expected range: {}'.
+                         format(partition_index))
+      # Note, a '_' is automatically added, so output table name would be:
+      #   known_args.output_table + '_' + partition_name
+      return '_' + self._partition_names[partition_index]
```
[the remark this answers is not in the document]
samanvp: done.

gcp_variant_transforms/libs/variant_partition.py
samanvp: > (even though it is now)

:)
done.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -218,38 +219,64 @@ def run(argv=None):
...
-           processed_variant_factory,
-           append=known_args.append,
-           allow_incompatible_records=known_args.allow_incompatible_records,
-           omit_empty_sample_calls=known_args.omit_empty_sample_calls,
-           num_bigquery_write_shards=known_args.num_bigquery_write_shards))
+    variants[i] |= (
+        'ProcessVaraints' + suffix >>
+        beam.Map(processed_variant_factory.create_processed_variant).\
+            with_output_types(processed_variant.ProcessedVariant))
+  # If this is auto partitioning (i.e. no config is given) flatten partitions.
```
arostamianfar: nit: (same as above) I think this comment is very specific to the implementation logic in should_flatten and not necessary (i.e. it's ok for this to be inside should_flatten, but not here).

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -218,38 +219,64 @@ def run(argv=None):
...
+  if partitioner and partitioner.should_flatten():
+    variants = [variants | 'FlattenPartitions' >> beam.Flatten()]
+    num_partitions = 1
+
+  for i in range(num_partitions):
+    suffix = ''
```
arostamianfar: nit: s/suffix/table_suffix?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -218,38 +219,64 @@ def run(argv=None):
...
-           known_args.output_table,
-           header_fields,
-           variant_merger,
-           processed_variant_factory,
-           append=known_args.append,
-           allow_incompatible_records=known_args.allow_incompatible_records,
-           omit_empty_sample_calls=known_args.omit_empty_sample_calls,
-           num_bigquery_write_shards=known_args.num_bigquery_write_shards))
+    variants[i] |= (
+        'ProcessVaraints' + suffix >>
```
arostamianfar: I dont think this would work for auto partitions as all partitions would have the same (empty string) suffix? In any case, I dont think having `suffix` is necessary here, so just using `i` is probably sufficient.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -218,38 +219,64 @@ def run(argv=None):
...
+    variants = [variants | 'FlattenPartitions' >> beam.Flatten()]
+    num_partitions = 1
+
+  for i in range(num_partitions):
+    suffix = ''
+    if partitioner and partitioner.get_partition_name(i):
+      suffix = '_' + partitioner.get_partition_name(i)
+    _ = (variants[i] | 'VariantToBigQuery' + suffix >>
+         variant_to_bigquery.VariantToBigQuery(
+             known_args.output_table + suffix,
```
arostamianfar: curious: are there any restrictions on the naming here? Just wondering if someone has a very long or strange suffix (ok for table name) that it wont break Dataflow. Again, probably just ok to use `i` here as well, but up to you.
nit: if you do decide to use `i`, then I recommend changing above to be `table_name = known_args.output_table` and then do `table_name += ...` in the if statement.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -218,38 +219,64 @@ def run(argv=None):
...
+  if partitioner:
+    num_partitions = partitioner.get_num_partitions()
+    partitioned_variants = variants | 'PartitionVariants' >> beam.Partition(
+        partition_variants.PartitionVariants(partitioner), num_partitions)
+    variants = []
+    for i in range(num_partitions):
+      if partitioner.should_keep_partition(i):
+        variants.append(partitioned_variants[i])
+      else:
+        #  An extra dummy partition has been added for residual variants. This
```
arostamianfar: nit: I think this comment is too specific to the particular implementation logic inside partitioner (e.g. if we change the implementation of should_keep_partition, there shouldn't be any need to update the comment here). I think it should be something like `remove filtered partitions`, but I think it's obvious from the code itself so there is no need for the comment.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -218,38 +219,64 @@ def run(argv=None):
...
+        #  extra partition will be ignored in the rest of the pipeline.
+        num_partitions -= 1
+  else:
+    # By default we don't partition data, so we have only 1 partition.
```
arostamianfar: super nit: s/data/the data

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
+      return self._should_keep_residual_partition
+
+  def get_partition_name(self, partition_index):
+    # type: (int) -> Optional[str]
+    if self._config_file_path_given:
+      if partition_index >= self._num_partitions or partition_index < 0:
+        raise ValueError(
+            'Given partition index {} is outside of expected range: '
+            '[0, {}]'.format(partition_index, self._num_partitions))
+      # Note, a '_' is automatically added, so output table name would be:
```
arostamianfar: please remove comment :)

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
+
+  def should_flatten(self):
+    # type: (None) -> bool
+    # In case we are in auto mode (ie no config) we flatten partitions and
+    # produce only one output table.
+    return not self._config_file_path_given
+
+  def should_keep_partition(self, partition_index):
+    # type: (int) -> bool
+    if partition_index != self._residual_partition_index:
```
arostamianfar: Now, this logic deserves a brief comment :)

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
       # Save partition in _reference_name_to_partition dict for future lookups
-      self._reference_name_to_partition_map[reference_name] = partition
-    return partition
+      self._ref_name_to_partitions_map[reference_name].add_region(
+          0, sys.maxint, partition_index)
+      return partition_index
+
+  def should_flatten(self):
+    # type: (None) -> bool
+    # In case we are in auto mode (ie no config) we flatten partitions and
```
arostamianfar: nit: s/ie/i.e.

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
 
+# At most 1000 partitions can be set as output of VariantTransform.
+_MAX_NUM_PARTITIONS = 1000
+# Each partition can contain at most 64 regions.
+_MAX_NUM_REGIONS = 64
+# Matches to regions formatted as 'chr12:10,000-20,000' used in par_config.yaml
+_REGION_LITERAL_REGEXP = re.compile(r'^(\S+):([0-9,]+)-([0-9,]+)$')
+# A special literal for identifying residual partition's region name.
+_RESIDUAL_REGION_LITERAL = 'residual'
+PARTITION_NOT_FOUND = -1
```
arostamianfar: How about UNDEFINED_PARTITION_INDEX? PARTITION_NOT_FOULD looks like a boolean value. We're also using it to initialize for `_residual_partition_index`

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
     if _DEFAULT_NUM_PARTITIONS <= _RESERVED_AUTO_PARTITIONS:
       raise ValueError(
           '_DEFAULT_NUM_PARTITIONS must be > _RESERVED_AUTO_PARTITIONS')
-    self._reference_name_to_partition_map = {}
+    self._num_partitions = _DEFAULT_NUM_PARTITIONS
+    # This variable determines the operation mode auto (default mode) vs config.
+    self._config_file_path_given = False
+    # Residual partition will contain all remaining variants that do not match
+    # to any other partition. -1 means it does not exist.
+    self._residual_partition_index = -1
```
arostamianfar: Please use the constant instead of `-1` and you can remove the extra comment about the meaning of `-1` :)

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
+      partition = partition_config.get('partition', None)
+      if partition is None:
+        raise ValueError('Wrong yaml file format, partition field missing.')
+      regions = partition.get('regions', None)
+      if regions is None:
+        raise ValueError('Each partition must have at least one region.')
+      if len(regions) > _MAX_NUM_REGIONS:
+        raise ValueError('At most {} regions per partition, thie partition '
+                         'contains {}'.format(_MAX_NUM_REGIONS, len(regions)))
+      # Check the partition_name is unique.
```
arostamianfar: nit: this comment should be below this if? Again, I don't think it adds any value as the if statement is pretty self-explanatory IMO.

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
+      if not partition.get('partition_name', None):
+        raise ValueError('Each partition must have partition_name field.')
+      partition_name = partition.get('partition_name').strip()
+      if partition_name == '':
```
arostamianfar: nit: please use `if not partition_name`

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
+        raise ValueError('Partition name can not be empty string.')
+      if partition_name in existing_partition_names:
+        raise ValueError('Table names must be unique, '
```
arostamianfar: s/Table/Partition

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
+    partition_configs = self._validate_config(config_file_path)
+
+    self._num_partitions = len(partition_configs)
+    for partition_index in range(self._num_partitions):
+      partition = partition_configs[partition_index].get('partition')
+      self._partition_names[partition_index] = (
+          partition.get('partition_name').strip())
+      regions = partition.get('regions', None)
+
+      # Check whether this is the residual partition.
```
arostamianfar: nit: by changing that logic to a method, you no longer need this comment (one of the points of making this into a method), so please remove.

gcp_variant_transforms/libs/variant_partition.py
```diff
@@ -34,59 +44,266 @@
...
+      if _is_residual_partition(regions):
+        if self._residual_partition_index != -1:
```
arostamianfar: test with constant instead of `-1` here as well.

gcp_variant_transforms/libs/variant_partition_test.py
```diff
@@ -62,3 +66,349 @@ def test_auto_partitioning_invalid_partitions(self):
 
     with self.assertRaisesRegexp(ValueError, 'Cannot partition given input*'):
       partitioner.get_partition('  ', 1)
+
+  def test_config_boundaries(self):
+    partitioner = variant_partition.VariantPartition(
+        "gcp_variant_transforms/testing/data/partition_configs/"
```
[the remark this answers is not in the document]
samanvp: done.

gcp_variant_transforms/libs/variant_partition_test.py
```diff
@@ -62,3 +66,349 @@ def test_auto_partitioning_invalid_partitions(self):
...
+        partitioner.get_partition('chr2_alternate_name1', 0), 3)
+    self.assertEqual(
+        partitioner.get_partition('chr2_alternate_name1', 999999999000), 3)
+    self.assertEqual(partitioner.get_partition('chr2_alternate_name2', 0), 3)
+    self.assertEqual(
+        partitioner.get_partition('CHR2_ALTERNATE_NAME2', 999999999000), 3)
+    self.assertEqual(partitioner.get_partition('2', 0), 3)
+    self.assertEqual(partitioner.get_partition('2', 999999999000), 3)
+
+    # "C4" OR "cr5" OR "chr6:1,000,000-2,000,000"
```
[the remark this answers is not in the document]
samanvp: done.

gcp_variant_transforms/libs/variant_partition_test.py
```diff
@@ -62,3 +66,349 @@ def test_auto_partitioning_invalid_partitions(self):
...
+        "partition_config1.yaml")
+    self.assertFalse(partitioner.should_flatten())
+    self.assertEqual(partitioner.get_num_partitions(), 8)
+    self.assertEqual(partitioner.get_residual_partition_index(), 7)
+    self.assertTrue(partitioner.has_residual_partition())
+
+    # "chr1:0-1,000,000"
+    self.assertEqual(partitioner.get_partition('chr1', 0), 0)
+    self.assertEqual(partitioner.get_partition('Chr1', 0), 0)
+    self.assertEqual(partitioner.get_partition('cHr1', 0), 0)
```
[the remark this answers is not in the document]
samanvp: I just wanted to emphasize the fact of not being case sensitive :) removed a bunch of them.

gcp_variant_transforms/libs/variant_partition_test.py
```diff
@@ -62,3 +66,349 @@ def test_auto_partitioning_invalid_partitions(self):
...
+                     partitioner.get_residual_partition_index())
+    self.assertEqual(partitioner.get_partition('cr5', 0),
+                     partitioner.get_residual_partition_index())
+    self.assertEqual(partitioner.get_partition('chr6', 0),
+                     partitioner.get_residual_partition_index())
+
+  def test_config_failed_missing_region(self):
+    tempdir = temp_dir.TempDir()
+    missing_region = [
+        '-  partition:\n',
```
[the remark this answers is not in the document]
samanvp: done.

PR #238, opened 2018-05-11
Add type annotation for beam.io #189

allieychen:
Add type annotation for `vcf_header_io.py` and `vcfio.py`.

Tested: unit tests
Issue: [189](https://github.com/googlegenomics/gcp-variant-transforms/issues/189)

### DIFF
--- a/gcp_variant_transforms/beam_io/vcf_header_io.py
+++ b/gcp_variant_transforms/beam_io/vcf_header_io.py
@@ -18,11 +18,12 @@ from __future__ import absolute_import
 
 from collections import OrderedDict
 from functools import partial
-from typing import Dict  # pylint: disable=unused-import
+from typing import Dict, Iterable  # pylint: disable=unused-import
 import vcf
 
 import apache_beam as beam
 from apache_beam.io import filebasedsource
+from apache_beam.io import range_trackers  # pylint: disable=unused-import
 from apache_beam.io.filesystem import CompressionTypes
 from apache_beam.io.filesystems import FileSystems
 from apache_beam.io.iobase import Read
@@ -38,7 +39,6 @@ class VcfHeaderFieldTypeConstants(object):
   STRING = 'String'
   FLAG = 'Flag'
   CHARACTER = 'Character'
-  STRING = 'String'
 
 
 class VcfParserHeaderKeyConstants(object):
@@ -63,6 +63,7 @@ class VcfHeader(object):
                contigs=None,  # type: Dict[str, OrderedDict[vcf.parser._Contig]]
                file_name=None  # type: str
               ):
+    # type: (...) -> None
     """Initializes a VcfHeader object.
 
     It keeps the order of values in the input dictionaries. Order is important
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -161,17 +168,18 @@ def __init__(
       compression_type=CompressionTypes.AUTO,
       validate=True,
       **kwargs):
+    # type: (str, str, bool, **str) -> None
```
bashir2: nit: Consider moving type hints to the line where the argument is declared for multi-line argument lists, like `read_records` function you have added above.

bashir2: BTW, is **str correct? I thought for variable-length arguments, you just need to define the type of the values, e.g., `str` in this case. [Here](https://github.com/googlegenomics/gcp-variant-transforms/blob/5512e444267d67b826f5ab323904bcff8ea56ef3/gcp_variant_transforms/libs/annotation/vep/vep_runner.py#L196) is one example but please let me know if I have done it wrong.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -18,7 +18,8 @@
 """
 
 from __future__ import absolute_import
-
+# pylint: disable=unused-import
```
arostamianfar: nit: please put this in front of the next line.

gcp_variant_transforms/beam_io/vcf_header_io.py
allieychen: Sometimes I couldn't decide whether to go for one line or multi-line type hints. Usually when the entire annotation cannot fit into one line, I will go for multi-line arguments. Do you have any kind of standard (for example, like when there are more than 3 arguments) for the multiple lines?

allieychen: Yes, I think it works for python 3. But for python 2, there is a warning in IntelliJ saying it is expecting a dictionary.
The suggested syntax for Python 2.7, quoted from [PEP 484](https://www.python.org/dev/peps/pep-0484/):
``When using the short form, for *args and **kwds, put 1 or 2 stars in front of the corresponding type annotation. (As with Python 3 annotations, the annotation here denotes the type of the individual argument values, not of the tuple/dict that you receive as the special argument value args or kwds.)``

gcp_variant_transforms/beam_io/vcfio.py
allieychen: Done.

gcp_variant_transforms/beam_io/vcf_header_io.py
arostamianfar: I'm not aware of any 'standards' either, but I think in general it's just a judgement call about how easy it is to map args to their values. I think Bashir's suggestion was to always do inline when we have multi-line args, but I think that's an overkill in some situations like here where it's relatively easy to map types to args. There is another example where I suggested moving to inline though as there are many args and it's more difficult.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -375,6 +378,7 @@ def __init__(self,
                buffer_size=DEFAULT_VCF_READ_BUFFER_SIZE,
                validate=True,
                allow_malformed_records=False):
+    # type: (str, List[str], str, int, bool, bool) -> None
```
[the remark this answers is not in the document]
allieychen: Done.

gcp_variant_transforms/beam_io/vcf_header_io.py
allieychen: Thanks for clarifying. I fixed that and some other similar methods.

PR #240, opened 2018-05-22
Add join query to annotation doc.

arostamianfar:


docs/variant_annotation.md
```diff
@@ -138,3 +144,34 @@ variants.
...
+  vcf_imports_external.gnomad_genomes_chr_hg19 AS T, T.alternate_bases AS ALT, ALT.CSQ AS CSQ
+WHERE
+  CSQ.SYMBOL IN (SELECT DB_Object_Symbol
+                 FROM `isb-cgc.genome_reference.GO_Annotations`
+                 WHERE GO_ID = 'GO:0006302')
+  AND CSQ.IMPACT = "HIGH"
+ORDER BY start_position
+```
+
+![A sample query joining annotations with GO database](images/gnomAD_GO_0006302.png)
```
bashir2: In the snapshot, please also include info about the size of the processed data and run-time of this query (maybe the query as well to be consistent with the above snapshot and for easy re-use in presentations, etc.).

docs/variant_annotation.md
```diff
@@ -138,3 +144,34 @@ variants.
...
+
+Other pathway databases are also available in BigQuery including
+[Reactome](https://reactome.org/)
+([table](https://bigquery.cloud.google.com/table/isb-cgc:genome_reference.Ensembl2Reactome?tab=details))
+and [WikiPathways](https://www.wikipathways.org)
+([table](https://bigquery.cloud.google.com/table/isb-cgc:QotM.WikiPathways_20170425_Annotated?tab=details)).
```
bashir2: I think it is better if we add some explicit acknowledgement/reference to [ISB-CGC](https://isb-cgc.appspot.com/) as I think these are all created and published as part of that effort.

docs/variant_annotation.md
```diff
@@ -138,3 +144,34 @@ variants.
...
+[GO (Gene Ontology) biological process (GO:0006302)](http://amigo.geneontology.org/amigo/term/GO:0006302).
+
+```
+#standardSQL
+SELECT
+  reference_name, start_position, reference_bases, ALT.alt, CSQ.*
+FROM
+  vcf_imports_external.gnomad_genomes_chr_hg19 AS T, T.alternate_bases AS ALT, ALT.CSQ AS CSQ
+WHERE
+  CSQ.SYMBOL IN (SELECT DB_Object_Symbol
```
bashir2: In general, do you know if the `SYMBOL` fields use standard/unique terms in these different datasets? IOW, are they coming from the same source?

docs/variant_annotation.md
```diff
@@ -156,20 +153,22 @@ the double-strand break repair pathway as defined by the
 ```
 #standardSQL
 SELECT
-  reference_name, start_position, reference_bases, ALT.alt, CSQ.*
+  reference_name, start_position, reference_bases, ALT.alt, CSQ.Consequence, CSQ.Impact, CSQ.SYMBOL
```
bashir2: nit: You can break this line to be less than 80 chars. In general, I would have preferred the following style (which is also consistent with the one above):
```
SELECT reference_name, start_position, reference_bases, ALT.alt,
    CSQ.Consequence, CSQ.Impact, CSQ.SYMBOL
FROM vcf_imports_external.gnomad_genomes_chr_hg19` AS T,
    T.alternate_bases AS ALT, ALT.CSQ AS CSQ
WHERE ...
```
but up to you.

docs/variant_annotation.md
bashir2: Hmm, okay, then please add a short comment about this in the text.

docs/variant_annotation.md
arostamianfar: I personally find that style to be less readable (per Google's styleguide), but broke up the lines in the middle to make them <80 chars (the "proper" way is to list each arg in one line if we break them, but I think it's overkill here :p ).

docs/variant_annotation.md
arostamianfar: Done.

PR #241, opened 2018-05-25
Add types (issue 189)

allieychen:
Add types annotation for the project. This pull request completes [this project](https://github.com/googlegenomics/gcp-variant-transforms/projects/5).

Issues: [189](https://github.com/googlegenomics/gcp-variant-transforms/issues/189)

### DIFF
--- a/gcp_variant_transforms/libs/annotation/vep/vep_runner.py
+++ b/gcp_variant_transforms/libs/annotation/vep/vep_runner.py
@@ -194,7 +194,7 @@ class VepRunner(object):
     }
 
   def _make_action(self, *args, **kwargs):
-    # type: (str, str) -> Dict
+    # type: (*str, **List[str]) -> Dict
     command_args = list(args)
     action = {
         'commands': command_args,
@@ -250,6 +250,7 @@ class VepRunner(object):
       return 'n1-standard-16'
 
   def wait_until_done(self):
+    # type: () -> None
     """Polls currently running operations and waits until all are done."""
     while self._running_operation_ids:
       self._running_operation_ids = [op for op in self._running_operation_ids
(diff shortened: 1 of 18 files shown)


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -126,6 +128,7 @@ def add_arguments(self, parser):
               'optimize_for_large_inputs are enabled.'))
 
   def validate(self, parsed_args, client=None):
+    # type: (argparse.Namespace) -> None
```
allieychen: I didn't find any function calls pass `client` except the unit tests. I wonder whether we can remove it.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -194,7 +194,7 @@ def _get_api_request_fixed_parts(self):
     }
 
   def _make_action(self, *args, **kwargs):
-    # type: (str, str) -> Dict
+    # type: (*str, **str) -> Dict
```
allieychen: @bashir2 Can you help check this method? There is a warning here. It seems that it is passing a List instead a dictionary.

bashir2: So like we discussed, the correct type is `**List[str]`, thanks for fixing this.

allieychen: Updated. Thanks!

gcp_variant_transforms/options/variant_transform_options.py
arostamianfar: I think unit tests is the main use case for it actually. We should keep it. Could you add the type for it?

PR #244, opened 2018-05-30
Integration tests fail

allieychen:
Related with [Pull 243](https://github.com/googlegenomics/gcp-variant-transforms/pull/243) and [Pull 237](https://github.com/googlegenomics/gcp-variant-transforms/pull/237). Add the `google-cloud-storage` as a dependency. 

Tested: Manually created a trigger in container registry.

### DIFF
--- a/cloudbuild_CI.yaml
+++ b/cloudbuild_CI.yaml
@@ -55,4 +55,4 @@ steps:
       # - '--gs_dir bashir-variant_integration_test_runs'
 images:
   - 'gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
-timeout: 45m
+timeout: 60m
--- a/deploy_and_run_tests.sh
+++ b/deploy_and_run_tests.sh
@@ -213,11 +213,11 @@ export VIRTUAL_ENV_DISABLE_PROMPT="something"
 virtualenv "${temp_dir}"
 source ${temp_dir}/bin/activate;
 trap clean_up EXIT
-pip install --upgrade .[int_test]
 if [[ -n "${run_unit_tests}" ]]; then
   pip install --upgrade .
   python setup.py test
 fi
+pip install --upgrade .[int_test]
 
 color_print "Running integration tests against ${full_image_name}" "${GREEN}"
 python gcp_variant_transforms/testing/integration/run_vcf_to_bq_tests.py \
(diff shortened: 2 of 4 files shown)


cloudbuild_CI.yaml
```diff
@@ -55,4 +55,4 @@ steps:
       # - '--gs_dir bashir-variant_integration_test_runs'
 images:
   - 'gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
-timeout: 45m
+timeout: 60m
```
allieychen: Unfortunately, these two integration tests do not run in parallel.

setup.py
```diff
@@ -26,12 +26,12 @@
     'intervaltree>=2.1.0,<2.2.0',
     'pyvcf<0.7.0',
     'mmh3<2.6',
+    'google-cloud-storage<=1.2.0'
```
bashir2: nit: Consider adding a comment here why you chose 1.2.0 (i.e., the discussion we had yesterday and other version conflicts, not that we, in our own code, rely on anything specific in 1.2.0).

deploy_and_run_tests.sh
```diff
@@ -233,7 +233,6 @@ if [[ -n "${run_preprocessor_tests}" ]]; then
       --staging_location "gs://${gs_dir}/staging" \
       --temp_location "gs://${gs_dir}/temp" \
       --logging_location "gs://${gs_dir}/temp/logs" \
-      --image "${full_image_name}" ${TEST_ARGUMENTS}
```
bashir2: Hmm, why did you drop `TEST_ARGUMENTS`?

cloudbuild_CI.yaml
bashir2: Hmm, can you elaborate why they are not run in parallel? Aren't we using TestRunner.run() in both cases and that is supposed to submit all requests and let them run together?

setup.py
allieychen: Done.

cloudbuild_CI.yaml
bashir2: Oh, now I realized what you mean by "these two integration tests" in your original comment; I thought you are talking about the two tests you added for preprocessor. Okay, I think this is fine for now.

deploy_and_run_tests.sh
bashir2: Okay, that makes sense.

PR #245, opened 2018-05-30
Added validation of input patterns

### LINKED ISSUE #82: Add better argument validation
We currently some argument validation (e.g. valid BigQuery dataset), but we should add more. Particularly, we should catch any exceptions when matching the input_pattern and provide a user-friendly error (e.g. invalid or inaccessible input_pattern currently displays a cryptic HttpError). We can do the same for temp_location and staging_location.

mhsaul:
Added validation of input_pattern before vcf_to_bq pipeline runs. Checks that input_pattern is well formed and that it matches at least one file.

Fixes #82.

### DIFF
--- a/gcp_variant_transforms/options/variant_transform_options.py
+++ b/gcp_variant_transforms/options/variant_transform_options.py
@@ -17,6 +17,8 @@ from __future__ import absolute_import
 import argparse  # pylint: disable=unused-import
 import re
 
+from apache_beam.io.filesystem import BeamIOError
+from apache_beam.io.filesystems import FileSystems
 from apache_beam.io.gcp.internal.clients import bigquery
 from apitools.base.py import exceptions
 from oauth2client.client import GoogleCredentials
@@ -97,10 +99,23 @@ class VcfReadOptions(VariantTransformsOptions):
       raise ValueError('Both --infer_headers and --representative_header_file '
                        'are passed! Please double check and choose at most one '
                        'of them.')
+    try:
+      # Gets at most one pattern match result of type `filesystems.MatchResult`.
+      first_match = FileSystems.match([parsed_args.input_pattern], [1])[0]
+      if not first_match.metadata_list:
+        raise ValueError(
+            'Input pattern {} returned '
+            'no matched files.'.format(parsed_args.input_pattern))
+    except BeamIOError:
+      # Retrieves the HttpError from BeamIOError.
+      raise ValueError('Invalid or inaccessible input pattern {}.'.format(
+          parsed_args.input_pattern))
 
 
 class BigQueryWriteOptions(VariantTransformsOptions):
   """Options for writing Variant records to BigQuery."""
+
+  _BQ_TABLE_URI = r'^((?P<project>.+):)(?P<dataset>\w+)\.(?P<table>[\w\$]+)$'
 
   def add_arguments(self, parser):
     # type: (argparse.ArgumentParser) -> None
@@ -145,8 +160,7 @@ class BigQueryWriteOptions(VariantTransformsOptions):
   def validate(self, parsed_args, client=None):
     # type: (argparse.Namespace, bigquery.BigqueryV2) -> None
     output_table_re_match = re.match(
-        r'^((?P<project>.+):)(?P<dataset>\w+)\.(?P<table>[\w\$]+)$',
-        parsed_args.output_table)
+        self._BQ_TABLE_URI, parsed_args.output_table)
     if not output_table_re_match:
       raise ValueError(
           'Expected a table reference (PROJECT:DATASET.TABLE) '


mhsaul:
I signed it!

mhsaul:
I signed it!

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -89,10 +92,27 @@ def add_arguments(self, parser):
               'variants. This is useful when there are header fields in '
               'variants not defined in the header sections.'))
 
+  def validate(self, parsed_args):
+    input_pattern_re_match = re.match(
```
arostamianfar: actually, the input_pattern can be a local file as well. I actually don't think we need this check as the one below it would cover invalid paths as well.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -89,10 +92,27 @@ def add_arguments(self, parser):
...
+      raise ValueError(
+          'Expected a file input_pattern (SCHEME://BUCKET/OBJECT) '
+          'instead of {}.'.format(parsed_args.input_pattern))
+
+    # Gets at most one pattern match result of type `filesystems.MatchResult`
+    first_match = FileSystems.match([parsed_args.input_pattern], [1])[0]
+    if not first_match.metadata_list:
+      raise ValueError(
+          'File pattern {} returned '
+          'no matched files.'.format(parsed_args.input_pattern))
```
arostamianfar: Could you please also catch the exception where you don't have access to the source? you can test with `--input_pattern=gs://test/*.vcf`.

gcp_variant_transforms/options/variant_transform_options.py
mhsaul: removed.

gcp_variant_transforms/options/variant_transform_options.py
mhsaul: Done.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -89,10 +91,25 @@ def add_arguments(self, parser):
...
 
+  def validate(self, parsed_args):
+    try:
+      # Gets at most one pattern match result of type `filesystems.MatchResult`.
+      first_match = FileSystems.match([parsed_args.input_pattern], [1])[0]
+      if not first_match.metadata_list:
+        raise ValueError(
+            'File pattern {} returned '
+            'no matched files.'.format(parsed_args.input_pattern))
+    except BeamIOError as e:
```
mhsaul: @arostamianfar : This only checks `storage.objects.list` permission as that is what is required for the match condition. To check `storage.objects.get` (which is what a file read specifically uses I think) we would probably need to do a read of the match result. Additionally, if files matching the input pattern have different permissions (some readable, some not readable) the validation could pass if the first file matched has acceptable permissions.

mhsaul:
@arostamianfar Thanks for the review! PTAL

gcp_variant_transforms/options/variant_transform_options.py
arostamianfar: I think those edge cases are ok and we're doing "best effort" here. In all problematic cases that I know of, the users had a typo in their GCS bucket path, which resulted in a cryptic http error message, so they thought something was wrong with our code.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -89,10 +91,25 @@ def add_arguments(self, parser):
...
+      # Retrieves the HttpError from BeamIOError.
+      match_error = e.exception_details[parsed_args.input_pattern]
```
arostamianfar: match_error is not used?
In any case, I think it's better to throw a more user-friendly error message here...something like "Invalid or inaccessible input pattern {}" b/c the original exceptions from Beam can be cryptic.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -89,10 +91,25 @@ def add_arguments(self, parser):
               'variants. This is useful when there are header fields in '
               'variants not defined in the header sections.'))
 
+  def validate(self, parsed_args):
+    try:
+      # Gets at most one pattern match result of type `filesystems.MatchResult`.
+      first_match = FileSystems.match([parsed_args.input_pattern], [1])[0]
+      if not first_match.metadata_list:
+        raise ValueError(
+            'File pattern {} returned '
```
arostamianfar: nit: s/File/Input (just to be consistent with the argument).

mhsaul:
@arostamianfar PTAL

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -17,6 +17,8 @@
 import argparse  # pylint: disable=unused-import
 import re
 
+from apache_beam.io.filesystem import BeamIOError
```
arostamianfar: nit: please do `from apache_beam.io import filesystem` and `from apache_beam.io import filesystems` and then refer to `filesystem.BeamIOError` below (it's a google style thing to not import classes as much as possible).

Issue #86, opened 2018-01-23
Add native support for table partitions by chromosome


arostamianfar:
For large (1TB+) datasets, we should provide an option to output multiple BigQuery tables for each chromosome. This makes querying the data cheaper for certain types of queries (e.g. querying variants in a particular gene). Users can use [BigQuery wildcards](https://cloud.google.com/bigquery/docs/querying-wildcard-tables) to query over all of the data.

Things to consider:
- We currently cannot dynamically change the number of outputs in Beam, so we need to ask the user about the nature and number of partitions prior to running the pipeline. We can provide common ones (e.g. homo sapiens). 
- Need to analyze the performance impact (if any) of having 24+ tables vs. 1 table. This should be documented if there is a significant performance hit (e.g. users may want to create a unified table for queries that span all of the genome).
- Doing a simple partition by "reference name" won't work as some datasets may have a large number (1000s) of reference names. We need a "catch all" partition.

Related:
- BigQuery pricing model (https://cloud.google.com/bigquery/pricing): Pricing is based on the columns accessed and not the number of records from each column. In other words, adding "where reference_name = chr1" does not affect pricing.

PR #246, opened 2018-05-31
Partitioning Part 2 (input config file)

samanvp:
Following the PR #195, here we add the functionality that user can input
a config file to partition output tables to optimise their future BQ
query costs.

Issue #86.


samanvp:
All integration tests passed successfully.

Issue #247, opened 2018-05-31
Add integration tests for Partitioning


samanvp:
After landing PR #246 we need integration tests for testing the correctness of partitioning functionality.

PR #251, opened 2018-05-31
Add integration tests for partitioning

samanvp:
Following PR #246 , here we add two integration tests to validate the correct
functionality of partitioning.

solves issue #247 
TESTED= time ./deploy_and_run_tests.sh --run_presubmit_tests took ~39 min

### DIFF
--- a/gcp_variant_transforms/testing/integration/vcf_to_bq_tests/presubmit_tests/large_tests/platinum_partition_with_residual.json
+++ b/gcp_variant_transforms/testing/integration/vcf_to_bq_tests/presubmit_tests/large_tests/platinum_partition_with_residual.json
@@ -1,0 +1,57 @@
+{
+  "test_name": "platinum-partition-with-residual",
+  "table_name": "platinum_partition_with_residual",
+  "input_pattern": "gs://genomics-public-data/platinum-genomes/vcf/*.vcf",
+  "partition_config": "gcp_variant_transforms/testing/data/partition_configs/integration_with_residual.yaml",
+  "runner": "DataflowRunner",
+  "variant_merge_strategy": "MOVE_TO_CALLS",
+  "worker_machine_type": "n1-highmem-4",
+  "max_num_workers": "60",
+  "num_workers": "60",
+  "assertion_configs": [
+    {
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/testing/integration/run_vcf_to_bq_tests.py
```diff
@@ -165,6 +165,8 @@ def _replace_macros(self, query):
     for macro in self._QueryMacros:
       if macro.name == query:
         return macro.value
+      if query.startswith(macro.name):
+        return macro.value + query[len(macro.name):]
```
arostamianfar: oh, this is very hacky :) it assumes the query ends with `{TABLE_NAME}` and the logic is cryptic.
Please just inline the entire query for your tests (it's ok to have duplicate logic in this case as it's a very specific use case).
`SELECT COUNT(0) AS num_rows FROM {TABLE_NAME}_chr01`, etc.

gcp_variant_transforms/testing/integration/vcf_to_bq_tests/presubmit_tests/large_tests/platinum_partition_with_residual.json
```diff
@@ -0,0 +1,53 @@
...
+  "input_pattern": "gs://genomics-public-data/platinum-genomes/vcf/*.vcf",
+  "partition_config": "gcp_variant_transforms/testing/data/partition_configs/integration_with_residual.yaml",
+  "runner": "DataflowRunner",
+  "variant_merge_strategy": "MOVE_TO_CALLS",
+  "worker_machine_type": "n1-highmem-4",
+  "max_num_workers": "40",
+  "num_workers": "40",
+  "assertion_configs": [
+    {
+      "query": ["NUM_ROWS_QUERY_chr01"],
```
arostamianfar: can you also add a query like:
```
SELECT COUNT(0) AS num_rows FROM `{TABLE_NAME}*`
```
to make sure there are no additional tables created. You can do the same for the no residual one as well to ensure the residual one is not created.

(you may need to have backquotes around `{TABLE_NAME}`.

samanvp: done.

Issue #83, opened 2018-01-23
Automatically modify the BigQuery schema when appending data to an existing table


arostamianfar:
We currently provide a "--append" option to append data to an existing table. However, it requires the schema to be exactly the same. We can use the BigQuery API to update the schema prior to import (in cases where the schema is compatible) or throw an error when the schema is incompatible (e.g. the same field is defined with a different type).

PR #260, opened 2018-06-07
Update the schema #83

allieychen:
- Add a new flag `update_schema` (requires `--append`). When it is set to true, BigQuery schema will be updated to be the merged version of the previous schema and the new schema.
- If the type or the mode is different for the same field id, throw an exception.
- Integration test will be added later.

Issue: [83](https://github.com/googlegenomics/gcp-variant-transforms/issues/83)
Tested: Manually run `vcf_to_bq` to an existing table with a different`input_file` (different schema).

### DIFF
--- a/gcp_variant_transforms/transforms/variant_to_bigquery.py
+++ b/gcp_variant_transforms/transforms/variant_to_bigquery.py
@@ -16,12 +16,20 @@
 
 from __future__ import absolute_import
 
+import exceptions
 import random
+import re
+from typing import Dict, List  # pylint: disable=unused-import
+
 import apache_beam as beam
+from apache_beam.io.gcp.internal.clients import bigquery
+from apitools.base.py import exceptions
+from oauth2client.client import GoogleCredentials
 
 from gcp_variant_transforms.beam_io import vcf_header_io  # pylint: disable=unused-import
 from gcp_variant_transforms.libs import bigquery_row_generator
 from gcp_variant_transforms.libs import bigquery_schema_descriptor  # pylint: disable=unused-import
+from gcp_variant_transforms.libs import bigquery_util
 from gcp_variant_transforms.libs import bigquery_vcf_schema
 from gcp_variant_transforms.libs import processed_variant
 from gcp_variant_transforms.libs import vcf_field_conflict_resolver
@@ -65,6 +73,7 @@ class VariantToBigQuery(beam.PTransform):
       variant_merger=None,  # type: variant_merge_strategy.VariantMergeStrategy
       proc_var_factory=None,  # type: processed_variant.ProcessedVariantFactory
       append=False,  # type: bool
+      update_schema=False,  # type: bool
       allow_incompatible_records=False,  # type: bool
       omit_empty_sample_calls=False,  # type: bool
       num_bigquery_write_shards=1  # type: int
@@ -84,8 +93,10 @@ class VariantToBigQuery(beam.PTransform):
         The latter functionality is what is needed here.
       append: If true, existing records in output_table will not be
         overwritten. New records will be appended to those that already exist.
+      update_schema: If true, BigQuery schema will be updated by combining the
+        existing schema and the new schema if they are compatible.
       allow_incompatible_records: If true, field values are casted to Bigquery
-+        schema if there is a mismatch.
++       schema if there is a mismatch.
       omit_empty_sample_calls: If true, samples that don't have a given call
         will be omitted.
       num_bigquery_write_shards: If > 1, we will limit number of sources which
@@ -108,6 +119,7 @@ class VariantToBigQuery(beam.PTransform):
     self._allow_incompatible_records = allow_incompatible_records
     self._omit_empty_sample_calls = omit_empty_sample_calls
     self._num_bigquery_write_shards = num_bigquery_write_shards
+    self._update_bigquery_schema(update_schema)
 
   def expand(self, pcoll):
     bq_rows = pcoll | 'ConvertToBigQueryTableRow' >> beam.ParDo(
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -65,6 +73,7 @@ def __init__(
       variant_merger=None,  # type: variant_merge_strategy.VariantMergeStrategy
       proc_var_factory=None,  # type: processed_variant.ProcessedVariantFactory
       append=False,  # type: bool
+      update_schema=False,  # type: bool
```
arostamianfar: nit: consider renaming this to `update_schema_on_append` here and everywhere. `update_schema` is generic and b/c this logic is tied to append, I prefer to have the flag and variable names be specific about it.

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -108,6 +119,7 @@ def __init__(
     self._allow_incompatible_records = allow_incompatible_records
     self._omit_empty_sample_calls = omit_empty_sample_calls
     self._num_bigquery_write_shards = num_bigquery_write_shards
+    self._update_bigquery_schema(update_schema)
```
arostamianfar: Please check if update_schema is true before calling this method and remove the check from inside the function (I think you can also assume that self._schema exists; otherwise, the method above would have returned an error).

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -147,3 +159,79 @@ def expand(self, pcoll):
                       beam.io.BigQueryDisposition.WRITE_APPEND
                       if self._append
                       else beam.io.BigQueryDisposition.WRITE_TRUNCATE))))
+
+  def _update_bigquery_schema(self, update_schema):
+    # type: (bool) -> None
+    if not update_schema or not self._schema:
+      return
+    # if table does not exist, do not need to update the schema.
+    output_table_re_match = re.match(
```
arostamianfar: Ideally, this should be combined with the logic in validate(), but please just add a TODO for now :)

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -147,3 +159,79 @@ def expand(self, pcoll):
...
+    original_schema = found_table.schema
+    found_table.schema = new_schema
+    try:
+      client.tables.Update(bigquery.BigqueryTablesUpdateRequest(
+          projectId=project_id,
+          datasetId=dataset_id,
+          table=found_table,
+          tableId=table_id))
+    except exceptions.HttpError:
+      found_table.schema = original_schema
```
arostamianfar: I think it's better to raise the error here and fail. If someone has explicitly set update_schema and we fail for any reason, we don't want to even start the pipeline.
You can do something like: `raise RuntimeError('BigQuery schema update failed: %s' % str(e))`

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -147,3 +159,79 @@ def expand(self, pcoll):
...
+
+
+def _merge_field_schemas(
+    field_schemas_1,  # type: List[bigquery.TableFieldSchema]
+    field_schemas_2  # type: List[bigquery.TableFieldSchema]
+    ):
+  # type: (...) -> List[bigquery.TableFieldSchema]
+  """Merges the `field_schemas_1` and `field_schemas_2`.
+
+  Raises an error if there are fields with the same name, but different modes or
```
arostamianfar: Please use Args/Raises syntax like here: https://github.com/googlegenomics/gcp-variant-transforms/blob/2de52811a0c63fbb11c8deaa1245890dfc8c0202/gcp_variant_transforms/libs/vcf_header_parser.py#L36

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -147,3 +159,79 @@ def expand(self, pcoll):
...
+        r'^((?P<project>.+):)(?P<dataset>\w+)\.(?P<table>[\w\$]+)$',
+        self._output_table)
+    credentials = GoogleCredentials.get_application_default().create_scoped(
+        ['https://www.googleapis.com/auth/bigquery'])
+    client = bigquery.BigqueryV2(credentials=credentials)
+    try:
+      project_id = output_table_re_match.group('project')
+      dataset_id = output_table_re_match.group('dataset')
+      table_id = output_table_re_match.group('table')
+      found_table = client.tables.Get(bigquery.BigqueryTablesGetRequest(
```
arostamianfar: nit: s/found_table/existing_table?

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -147,3 +159,79 @@ def expand(self, pcoll):
...
+          projectId=project_id,
+          datasetId=dataset_id,
+          tableId=table_id))
+    except exceptions.HttpError:
+      return
+
+    new_schema = bigquery.TableSchema()
+    new_schema.fields = _merge_field_schemas(self._schema.fields,
```
arostamianfar: I think it's actually better to swap the order of the args since when appending, I think you'd expect the order of the columns to remain the same as before and new ones just added to the end rather than the order potentially changing. So, you'd call with `(existing_table.schema.fields, self._schema.fields)`. WDYT?

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -147,3 +159,79 @@ def expand(self, pcoll):
...
+      client.tables.Update(bigquery.BigqueryTablesUpdateRequest(
+          projectId=project_id,
+          datasetId=dataset_id,
+          table=found_table,
+          tableId=table_id))
+    except exceptions.HttpError:
+      found_table.schema = original_schema
+
+
+def _merge_field_schemas(
```
arostamianfar: nit: s/_merge_field_schemas/_merge_schema_fields

gcp_variant_transforms/transforms/variant_to_bigquery.py
arostamianfar: Please add a new line with the note: `It returns a new schema with new fields from field_schemas_2 appended to field_schemas_1` so that it's more clear what is being mutated.

gcp_variant_transforms/transforms/variant_to_bigquery.py
arostamianfar: nit: s/_merge_field_schemas/_get_merged_field_schemas
Note the explicit `get` keyword as it indicates that the a new merged schema is returned rather than mutating either one of the passed in args.

gcp_variant_transforms/transforms/variant_to_bigquery_test.py
```diff
@@ -272,3 +273,323 @@ def test_convert_variant_to_bigquery_row_allow_incompatible_recoreds(self):
             self._row_generator, allow_incompatible_records=True)))
     assert_that(bigquery_rows, equal_to([row]))
     pipeline.run()
+
+  def test_merge_field_schemas_no_same_id(self):
```
arostamianfar: Thanks for adding such detailed unit tests!

Issue #230, opened 2018-05-07
Public release of VEP docker image.


bashir2:
It turned out that we are the first who need to publish VEP docker images on [gcr.io](http://gcr.io).

We need to follow our internal processes/policies to make this happen; this is a process tracking issue.
The [license](https://useast.ensembl.org/info/about/legal/code_licence.html) is [Apache 2.0](http://www.apache.org/licenses/LICENSE-2.0).

bashir2:
The VEP docker image is finally released publicly at `gcr.io/gcp-variant-annotation/vep_91`. One VEP cache is also publicly available at `gs://gcp-variant-annotation-vep-cache/vep_cache_homo_sapiens_GRCh38_91.tar.gz`

Closing this issue ...

PR #261, opened 2018-06-08
Default values for VEP image and cache.

bashir2:
Issue #230 

Tested: Manually triggered integration tests on Container Registry by pushing to a `test_CI` branch (running).


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -224,11 +224,12 @@ def add_arguments(self, parser):
               'structure as input files with a suffix added to them.'))
     parser.add_argument(
         '--' + AnnotationOptions._VEP_IMAGE_FLAG,
-        default="",
+        default='gcr.io/gcp-variant-annotation/vep_91',
         help=('The URI of the docker image for VEP.'))
     parser.add_argument(
         '--' + AnnotationOptions._VEP_CACHE_FLAG,
-        default="",
+        default=('gs://gcp-variant-annotation-vep_cache/'
```
arostamianfar: nit: any reason why `vep_cache` is an underscore instead of a dash like the other parts?

bashir2: Done.
In general for file names I prefer '_' over '-' (hence the cache name file below) and the reason I picked 'gcp-variant-annotation' was that it was the name of the project. But I think it is better to be consistent so I changed the bucket name. Also given the trouble that I once had with GATK and '_' in bucket name maybe for bucket names I should change my habit.

Issue #255, opened 2018-06-01
Revisit integration tests (reduce #, combine, use DirectRunner more)


arostamianfar:
As mentioned in issue #250, there is a max limit of 25 dataflow jobs per project, which cannot be increased. Our integration test cases were developed under the assumption that we dont have such a low limit. The tests currently take about an hour to run (~40mins for vcf_to_bq and ~15mins for preprocessor), and we should aim to keep them under an hour. Some ideas:
- Prioritize tests: we should start large tests before smaller ones. This is currently not an issue as majority of tests finish in ~5mins, so large tests wont be starved for more than 5mins. However, this can get worse as we add more tests.
- Remove redundant test cases (e.g. we dont expect anything new caught by "valid_4_1_gz.json" that is not already caught by "valid_4_0_gz.json" and "valid_4_1.json").
- Consider using DirectRunner for some of the small tests
- Combine tests together (e.g. we added `--num_bigquery_write_shards` to an existing platinum test instead of adding a new one just to test that flag).

pgrosu:
Asha, never skimp on your tests - that's what ensures you are validating code-updates against real use cases.    Just create more projects.

arostamianfar:
@pgrosu: we're definitely not reducing our test coverage :) Just removing the redundant tests and perhaps combining multiple tests into one so that we can utilize our resources better. Creating multiple projects is something we can consider as well, but I prefer to use it as a last resort if the proposed workarounds still don't scale.

PR #266, opened 2018-06-08
Add integration tests for update_schema_on_append #83

allieychen:
- Update the code to support sequential runs if two test configs are present in one JSON file. The second test case won't start until the first one finishes.
- Modified an existing test case by taking the flag `update_schema_on_append` and `append` to make sure it still works if the table does not exist. 
- Add one integration tests for the flag `update_schema_on_append` and `append`: append variants from `valid_4_2` to table of `valid_4_0`. Their schemas have some overlapped fields, but also each of them have their unique fields.
- It could be a solution for [Issue 255](https://github.com/googlegenomics/gcp-variant-transforms/issues/255), since it can force the larger tests to run first.

Issue: [83](https://github.com/googlegenomics/gcp-variant-transforms/issues/83)
Tested: ./deploy_and_run_tests


gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -146,25 +162,31 @@ def get_configs(test_file_path, required_keys):
...
+  # type: (str, List[str]) -> List[Dict]
   """Loads an integration test JSON object from a file."""
   with open(filename, 'r') as f:
-    test = json.loads(f.read())
-    _validate_test(test, filename, required_keys)
-    return test
+    tests = json.loads(f.read())
+    # When there are multiple test cases in one JSON file, it is a list. Wrap
+    # the dictionary in a list if there is only one test case in one JSON file.
+    if not isinstance(tests, list):
```
allieychen: Actually I prefer to modify all other JSON files be a list rather than wrapping it here. It would be more readable. WDYT?

gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -41,42 +41,57 @@ class TestRunner(object):
...
-    # type: (List[Union[PreprocessorTestCase, VcfToBQTestCase]]) -> None
+    # type: (List[List[Union[PreprocessorTestCase, VcfToBQTestCase]]]) -> None
     self._tests = tests
     self._service = discovery.build(
         'genomics',
         'v1alpha2',
         credentials=GoogleCredentials.get_application_default())
-    self._operation_names = []  # type: List[str]
-    self._responses = []  # List[Dict]
+    self._operation_names_to_tests = {}  # type: Dict
```
arostamianfar: As discussed offline: Please make this dict be operation_names_to_test_states, where the key is the operation_name and the value is a namedtuple TestCaseState(RunningTest, RemainingTests). This essentially makes a 'queue' for you and you can remove a test from the RemainingTests and run it at each step.

gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -41,42 +41,57 @@ class TestRunner(object):
   """Runs the tests using pipelines API."""
 
   def __init__(self, tests):
-    # type: (List[Union[PreprocessorTestCase, VcfToBQTestCase]]) -> None
+    # type: (List[List[Union[PreprocessorTestCase, VcfToBQTestCase]]]) -> None
```
arostamianfar: As discussed offline: consider making a TestCaseInterface base class that these inherit from. That way, the interface of the test case class is clear (i.e. validate_result method should be implemented in all test case classes).

gcp_variant_transforms/testing/integration/run_tests_common.py
arostamianfar: SGTM!

gcp_variant_transforms/testing/integration/vcf_to_bq_tests/presubmit_tests/small_tests/valid_4_0.json
```diff
@@ -1,20 +0,0 @@
-{
```
arostamianfar: is this removed by accident? :)

arostamianfar: ah, i see...sounds good!

gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -40,43 +54,71 @@ class TestCaseFailure(Exception):
 class TestRunner(object):
   """Runs the tests using pipelines API."""
 
-  def __init__(self, tests):
-    # type: (List[Union[PreprocessorTestCase, VcfToBQTestCase]]) -> None
+  def __init__(self, tests, revalidation=None):
```
arostamianfar: s/revalidation/revalidate
and this can just be a bool?

gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -40,43 +54,71 @@ class TestCaseFailure(Exception):
...
+    # type: (List[List[TestCaseInterface]], str) -> None
+    """Initializes the TestRunner.
+
+    Args:
+      tests: All test cases.
+      revalidation: If not none, skip most of the test (including table creation
```
arostamianfar: Consider rephrasing as "skip most of the test" implies skipping actual test cases. Something like:
`revalidate: If True, only run the result validation part of the tests.`

gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -40,43 +54,71 @@ class TestCaseFailure(Exception):
...
-      # `googleapiclient.discovery.Resource._set_service_methods`.
-      # pylint: disable=no-member
-      request = self._service.pipelines().run(body=test.pipeline_api_request)
-      operation_name = request.execute()['name']
-      self._operation_names.append(operation_name)
+    if self._revalidation:
+      for test_cases in self._tests:
+        for test_case in test_cases:
+          test_case.validate_result()
+          return
```
arostamianfar: this `return` should be on the `if` level? I also think that it's more readable in this case to just do an` else` with the rest of the logic since the `else` part isn't that large.

gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -40,43 +54,71 @@ class TestCaseFailure(Exception):
...
     self._wait_for_all_operations_done()
 
+  def _run_test(self, test_cases):
+    # type: (List[TestCaseInterface]) -> None
+    """Runs the first test case in `test_cases`.
+
+    The first test case and the remaining test cases form `TestCaseState` and
+    are added into `_operation_names_to_test_states` for future usage.
+    """
+
```
arostamianfar: nit: remove extra empty line.

gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -40,43 +54,71 @@ class TestCaseFailure(Exception):
...
-      for operation_name in self._operation_names:
-        if operation_name in running_operation_names:
-          request = operations.get(name=operation_name)
-          response = request.execute()
-          if response['done']:
-            self._handle_failure(response)
-            self._responses.append(response)
-            running_operation_names.remove(operation_name)
+      running_operation_names = self._operation_names_to_test_states.keys()
+      for operation_name in running_operation_names:
```
arostamianfar: nice way of avoiding two iterators! :)

gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -40,43 +54,69 @@ class TestCaseFailure(Exception):
...
-      # is dynamically added to the returned `service` object above. See
-      # `googleapiclient.discovery.Resource._set_service_methods`.
-      # pylint: disable=no-member
-      request = self._service.pipelines().run(body=test.pipeline_api_request)
-      operation_name = request.execute()['name']
-      self._operation_names.append(operation_name)
-    self._wait_for_all_operations_done()
+    if self._revalidate:
+      for test_cases in self._tests:
+        for test_case in test_cases:
```
arostamianfar: As discussed offline: I think we should just validate the last test in this case.

allieychen: Thanks very much on catching this issue! Done now. PTAL.

PR #272, opened 2018-06-13
Remove VariantInfo.field_count

samanvp:
and mondify all code to depend on header to look up the NUM.


arostamianfar:
Thanks, Saman! As discussed offline:
* Please remove VariantInfo completely.
* Please update the logic to use PyVCF ints instead of string but keep your existing make_header logic that takes care of string->int conversion. This would make it easier to use strings when we want to switch to using Nucleus.

samanvp:
I send a new PR that includes the changes we discussed. So if you don't mind I will close this one.

arostamianfar:
SGTM! Closing this one :)

PR #276, opened 2018-06-14
Update zone/region doc.

### LINKED ISSUE #275: Invalid command in documentation on running jobs
In the document, '[Running jobs in a particular region/zone](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/docs/running_in_particular_zone_region.md)', the command stated: 

```
gcloud alpha genomics pipelines run ... \
    --regions europe-west1 \
    --zones europe-west1-b
```

Returns the error:

`ERROR: (gcloud.alpha.genomics.pipelines.run) INVALID_ARGUMENT: Error: validating pipeline: zones and regions cannot be specified together`

arostamianfar:
Preview: https://github.com/arostamianfar/gcp-variant-transforms/blob/zone/docs/setting_zone_region.md
Tested: ran a pipeline in EU using the documented flags.
Fixes #275


docs/setting_zone_region.md
```diff
@@ -16,19 +18,25 @@ docker:
```
bashir2: Is having both region and zone here okay?
Also please consider adding a Tested field to the description if you have tested the config here.

arostamianfar: yes, it's ok for Dataflow flags. It used to be ok even for pipelines API but it has recently changed as 'regions' is a v2 parameter and 'zones' is a v1/v2 parameter. So, with v1 only 'zones' is actually taken into account and the error message was added recently about this.
Also added tested :)

PR #279, opened 2018-06-19
Add documentation for flattening BigQuery table.

### LINKED ISSUE #191: Document BigQuery flattening workaround
See snarfed's comment in #92 for context.

arostamianfar:
Preview: https://github.com/arostamianfar/gcp-variant-transforms/blob/docflat/docs/flattening_table.md

Fixes #191


arostamianfar:
@snarfed FYI; in case you like to review this as well! Thanks for your example query BTW! I ended up modifying it a bit to make it more general.

snarfed:
@arostamianfar this looks great! the generalizations absolutely make sense. glad we were able to help. thanks for writing it up!

smrgit:
I agree, this is super helpful!  One minor quibble and a question, though -- is it appropriate to use the word "join" when you say "To further flatten the BigQuery table on the genotype array ... you can add another explicit join with ..." ?  I didn't think that having comma-separated tables/fields in the FROM clause was equivalent to a JOIN (?)  Specifically, I'm looking at this [page](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#from-clause) in the documentation, where it discusses the ``field_path``.  Also, I had always been using UNNEST() in these kinds of queries, but I just saw (on the same doc page) in the blue "Note" that it seems that one can leave out the UNNEST (as you do in your examples) if there is a table name at the start of the field_path: 

> Note: If a path has only one name, it is interpreted as a table. To work around this, wrap the path using UNNEST, or use the fully-qualified path.

Mentioning that and pointing to that particular doc page in this new documentation here might be helpful to other users.

arostamianfar:
Thanks for the comments @smrgit and @snarfed !
RE using join: I based that on the documentation from [this page](https://cloud.google.com/bigquery/docs/reference/standard-sql/migrating-from-legacy-sql#removing_repetition_with_flatten) where it specifies that the semantics are equivalent to join. Note that we actually use this idea when doing `LEFT JOIN` on `t.alternate_bases` as we want to get variants that don't have an alternate.
RE unnest: Thanks! Updated the doc to include this point.

smrgit:
interesting, thanks @arostamianfar -- I didn't realize that it was indeed equivalent to a JOIN, good to know!

docs/flattening_table.md
```diff
@@ -0,0 +1,208 @@
...
+  IF(genotype > 0, alternate_bases[ORDINAL(genotype)], NULL) AS alternate_bases
+FROM
+  `project.dataset.table` AS t,
+  t.call AS call,
+  call.genotype as genotype
+```
+
+![Flatten call, genotype, alt](images/flatten_call_flatten_genotype_flatten_alt.png)
+
+Note that the semantics of the value of the `genotype` column is has changed
```
bashir2: nit: "is has"

docs/flattening_table.md
```diff
@@ -0,0 +1,208 @@
...
+as each row only contains a single alternate allele. As a result, you may
+decide to reformat that column using
+`IF(genotype > 0, 1, genotype) AS alt_genotype`, which results to:
+  * `0` implying reference match.
+  * `-1` implying not called.
```
bashir2: Where does `-1` come from?

docs/flattening_table.md
```diff
@@ -0,0 +1,208 @@
...
+
+```
+#standardSQL
+SELECT
+  reference_name, start_position, end_position, reference_bases,
+  call.name AS call_name, genotype
+FROM
+  `project.dataset.table` AS t,
+  t.call AS call,
+  call.genotype as genotype
```
bashir2: nit: `as` -> `AS` here and elsewhere

docs/flattening_table.md
```diff
@@ -0,0 +1,208 @@
...
+  IF(genotype > 0, 1, genotype) AS alt_genotype,
+  IF(genotype > 0, alts.alt, NULL) AS alt
+FROM
+  `project.dataset.table` AS t,
+  t.call AS call,
+  call.genotype as genotype
+LEFT JOIN
+  t.alternate_bases AS alts WITH OFFSET AS a_index
+WHERE
+  genotype IN(a_index + 1, 0, -1)
```
bashir2: nit: space after `IN` (below as well)

docs/flattening_table.md
```diff
@@ -0,0 +1,208 @@
...
+[Standard SQL](https://cloud.google.com/bigquery/docs/reference/standard-sql/)
+and not
+[Legacy SQL](https://cloud.google.com/bigquery/docs/reference/legacy-sql).
+
+
+## Flattening basics
+
+Consider the following BigQuery row:
+
+![Flatten original row](images/flatten_original_row.png)
```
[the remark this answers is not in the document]
arostamianfar: yeap, removed the extra blank spaces :)

docs/flattening_table.md
arostamianfar: Done.

docs/flattening_table.md
arostamianfar: Done. Thanks for catching!

docs/flattening_table.md
arostamianfar: It actually comes from the GA4GH API. I realize that we should update those as part of the documentation update work in Q3 as well (likely in the "understanding the BQ schema" doc). For now, I just added a note about it.

docs/flattening_table.md
arostamianfar: Done.

docs/flattening_table.md
bashir2: I see, thanks for clarification.

Issue #248, opened 2018-05-31
Add labels to output BigQuery tables


samanvp:
After landing PR #246 we need to add labels to output BigQuery tables so we will be able to query them efficiently.

PR #280, opened 2018-06-20
Add labels to output BigQuery tables

samanvp:
In order to automize finding relevant tables for each query based on its
`WHERE` clause, we need to add labels to each table to highlight their
content.
solves issue #248 

TESTED=unit_tests and presubmit_tests


arostamianfar:
Closing this as it's now a lower priority with [native BigQuery clustering](https://cloud.google.com/bigquery/docs/clustered-tables). We may revisit this later.

Issue #258, opened 2018-06-06
Catch more errors with the validator/preprocessor tool


arostamianfar:
The preprocessor tool currently catches VCF parsing errors, header mismatch errors, and undefined header errors. There are a few other errors that would be ideal to catch. Particularly:
- Header definition mismatch errors (e.g. header is defined as integer, but a float value is provided). This should ideally result in changing the recommendation for the field with the mismatch. 
- Alternate allele errors: catch cases where Number=A INFO values do not have the same cardinality as alternate bases.

allieychen:
The two types of mismatches listed above are added into the report by running the preprocessor tool with the flag `--report_all_conflicts`. The related fields and suggested resolutions are listed in `Inferred Headers` category. Please refer to [VCF files preprocessor doc](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/docs/vcf_files_preprocessor.md) for more details.

PR #283, opened 2018-06-26
Infer mismatched headers.

allieychen:
- Infer the headers when the definition mismatches the value, which handles the two cases listed in [258](https://github.com/googlegenomics/gcp-variant-transforms/issues/258).
- Also, add this type of conflicts into the preprocessor report.
- Update the integration tests.
- There are still some other places using infer_undefined_headers (e.g. the flag --infer_undefined_headers), another PR will be sent shortly after this PR as a clean up.

Issue: [258](https://github.com/googlegenomics/gcp-variant-transforms/issues/258).
Tested: unit tests & integration tests.


gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -78,16 +90,109 @@ def _can_cast_to(self, value, cast_type):
...
+
+    It handles one special case PyVCF cannot handle, i.e., the defined type is
+    `Integer`, but the provided value is float. In this case, correct the type
+    to be `Float`.
+
+    Note that if `value` is a float instance but with an integer value
+    (e.g. 2.0), the type will stay the same as `defined_type`.
+    """
+    if defined_type == _HeaderTypeConstants.INTEGER:
+      if isinstance(value, float) and not float(value).is_integer():
```
arostamianfar: nit: this can just be `value.is_integer()` as you have already verified that it's a float.
P.S. Nice find on the is_integer method! did not know about it :)

allieychen: Ah, good point! Thanks.

PR #284, opened 2018-06-26
Temporarily add <=2.4 for Beam version to fix the build

bashir2:
Tested: Reproduced the unit-test failure, applied this change, `pip install --upgrade .`, checked that the test passes.


bashir2:
Thanks @samanvp 
@allieychen please remove the extra condition and the TODO once you figured the root cause with Beam 2.5 (and thanks).

PR #287, opened 2018-06-28
Moving annotation parsing logic into its own module.

### LINKED ISSUE #186: Refactor ProcessedVariants module
The [ProcessedVariant](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/libs/processed_variant.py) module is now complicated since it has a lot of annotation parsing logic within it. We should refactor the module to create a new 'annotation' module that handles that logic.

bashir2:
Fixes #186 

Tested: unit-tests updated

Some notes about this PR:
* Counter handling is kept in the `processed_variant`; the new `annotation_parser` is purely responsible for annotation parsing and to signal errors it uses different exceptions. The caller (i.e., `processed_variant`) converts those exceptions to counter increments.
* This is mostly a refactoring and from a user's perspective it is almost a no-op; the exceptions are:
* There is no `ambiguous_allele` column in the output BigQuery table (or the map out of the parser). This column was conditional on the `minimal_match` case before but I decided to remove it completely and instead drop the annotations that have ambiguous ALT matching. In general, I think the `minimal_match` mode should be avoided if possible, but if it is used, it is better to drop ambiguous annotations.
* There was a `_convert_annotation_strs_to_alt_map` in the old code which was producing `Dict[str, List[Dict[str, str]]]`. This was ugly, hard to understand, and I never liked it. The reason for this structure was some minor performance gains. Instead of relying on this complex structure, each annotation string is matched independently. 
* As a result of the above change, the semantics of the `annotation_alt_match_counter` has slightly changed. Note that independent processing of annotation strings was done for `_use_allelel_num` even in the old code, so this change fixes the inconsistency between the counter semantics.
* `annotation_alt_minimal_match_counter` is dropped; all matching alts are counted under `annotation_alt_match`

### DIFF
--- a/gcp_variant_transforms/libs/annotation/annotation_parser.py
+++ b/gcp_variant_transforms/libs/annotation/annotation_parser.py
@@ -1,0 +1,334 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Module for parsing annotation fields.
+
+The main class is `Parser`; see its documentation for usage. There are also
+some helper methods that can be used in different contexts.
+"""
+
+from __future__ import absolute_import
+
+import re
+
+from typing import Dict, List, Tuple  # pylint: disable=unused-import
+
+# The key in annotation maps that keeps the original alternate allele in an
+# annotation string (and the field name in the BigQuery table that holds that
+# information).
+ANNOTATION_ALT = 'allele'
+
+# The annotation field that VEP uses to record the index of the alternate
+# allele (i.e., ALT) that an annotation list is for.
+_ALLELE_NUM_ANNOTATION = 'ALLELE_NUM'
+
+# The representation of a deletion variant in VEP.
+_COMPLETELY_DELETED_ALT = '-'
+
+# Regular expressions to identify symbolic and breakend ALTs used in
+# annotation alt matching.
+# Check the VCF spec for symbolic and breakend ALT formats.
+_SYMBOLIC_ALT_RE = re.compile(r'^<(?P<ID>.*)>$')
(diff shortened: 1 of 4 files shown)


bashir2:
@Jessime this is mostly FYI since you are looking at annotation related code; that said, you should feel free to comment on this code or any other PR in this repo.

@arostamianfar to review this, it might help a little bit if you diff the new `annotation_parser.py` with the `processed_variant.py` at HEAD. I have tried to keep the function order the same for a better diff. However, the amount of changes in the `annotation_parser` module is big enough that I like you to review the whole module in isolation as well. Some of the reasons for changes are explained in the PR description; thanks in advance.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -373,12 +345,10 @@ def __init__(self,
       if field not in header_fields.infos:
         raise ValueError('{} INFO not found in the header'.format(field))
       header_desc = header_fields.infos[field][_HeaderKeyConstants.DESC]
-      self._annotation_names_map[field] = _extract_annotation_names(
-          header_desc)
+      self._annotation_names_map[
```
arostamianfar: nit: I find it more readable to break on the `=` sign like:
```
self._annotation_names_map[field] = (
    annotation_parser.extract_annotation_names(header_desc))
```

gcp_variant_transforms/libs/annotation/annotation_parser.py
```diff
@@ -0,0 +1,334 @@
...
+_COMPLETELY_DELETED_ALT = '-'
+
+# Regular expressions to identify symbolic and breakend ALTs used in
+# annotation alt matching.
+# Check the VCF spec for symbolic and breakend ALT formats.
+_SYMBOLIC_ALT_RE = re.compile(r'^<(?P<ID>.*)>$')
+_BREAKEND_ALT_RE = (re.compile(
+    r'^(?P<up_to_chr>.*([\[\]]).*):(?P<pos>.*)([\[\]]).*$'))
+
+class AnnotationParserException(Exception):
```
arostamianfar: nit: styleguide requires two blank lines before every class definition. I know it would make this block longer, but I prefer to conform to the styleguide.

bashir2: Done.
Yes I knew I am breaking the style guide, in fact I even wanted `pass` to be on the same line but pylint complained :-)
I think this is a good/concise exception pattern but unfortunately the style guide rules make it too verbose (4 lines for each instead of one).

gcp_variant_transforms/libs/processed_variant.py
bashir2: Done.
FYI, IntelliJ formatted it that way.

PR #289, opened 2018-06-29
Beam filesystem

### LINKED ISSUE #286: VEP logs get written to wrong location if trailing slash is in --annotation_output_dir
If I do `--annotation_output_dir gs://jessime_test_bucket/vep_test/` instead of `--annotation_output_dir gs://jessime_test_bucket/vep_test`, then the log file gets written to the base directory. 

The fix should be as simple as changing:

```python
  def _get_output_log_path(self, output_dir, vm_index):
    # type: (str, int) -> str
    return '{}/logs/output_VM_{}.log'.format(output_dir, vm_index)
```

to

```python
  def _get_output_log_path(self, output_dir, vm_index):
    # type: (str, int) -> str
    return filesystems.FileSystems.join(output_dir, 'logs/output_VM_{}.log'.format(vm_index))
```

But I'm going to have to figure out testing before I can make the change.

Jessime:
Makes the location of the VEP log file robust to a trailing forward slash in the output_dir. 

Fixes #286


gcp_variant_transforms/libs/annotation/vep/vep_runner_test.py
```diff
@@ -101,6 +101,11 @@ def test_get_output_pattern(self):
     output_pattern = self._create_test_instance().get_output_pattern()
     self.assertEqual(output_pattern, _OUTPUT_DIR + '/**_vep_output.vcf')
 
+  def test_get_output_log_path(self):
+    test_instance = self._create_test_instance()
+    log_path = test_instance._get_output_log_path(_OUTPUT_DIR, 0)
```
bashir2: Please also add a test where `output_dir` parameter of `_get_output_log_path` has a `/` at the end and check that it is properly handled. This is basically the usual way of fixing a bug, i.e., adding a test that exposes the bug and fails without the fix but passes after the fix.

PR #293, opened 2018-07-05
clean up infer_undefined_headers #258

allieychen:
Follow up on [PR 283](https://github.com/googlegenomics/gcp-variant-transforms/pull/283).

Now the `infer headers` applies to both undefined headers and mismatched headers. This PR replaces all `infer_undefined_headers` with `infer_headers`. Update the corresponding doc as well.

Issue: [258](https://github.com/googlegenomics/gcp-variant-transforms/issues/258).

### DIFF
--- a/gcp_variant_transforms/options/variant_transform_options.py
+++ b/gcp_variant_transforms/options/variant_transform_options.py
@@ -82,12 +82,13 @@ class VcfReadOptions(VariantTransformsOptions):
               'Note that each VCF file must still contain valid header files '
               'even if this is provided.'))
     parser.add_argument(
-        '--infer_undefined_headers',
+        '--infer_headers',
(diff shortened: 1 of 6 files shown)


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -82,12 +82,13 @@ def add_arguments(self, parser):
               'Note that each VCF file must still contain valid header files '
               'even if this is provided.'))
     parser.add_argument(
-        '--infer_undefined_headers',
+        '--infer_headers',
```
arostamianfar: Please keep both versions to make this change backwards compatible. You can just add both like this:
```
parser.add_argument(
    '--infer_headers',
    '--infer_undefined_headers',
    type = 'bool', ...

```

Issue #85, opened 2018-01-23
Add BigQuery->VCF pipeline


arostamianfar:
There are certain use cases where users like to transform variants from BigQuery to VCF (e.g. to use tools that expect a VCF file). This task is to write such a pipeline.

As an extension, once this is completed, we can provide VCF->VCF and BQ->BQ transforms as well (e.g. just use our merging logic to merge a large number of VCF files and output a single merged VCF file).

PR #300, opened 2018-07-13
BQ to VCF Prototype #85

allieychen:
First version (simplest) of BQ to VCF.
- only variants (not sorted) are exported.
- it assumes each variant' list of calls contain all samples.

Issue: [85](https://github.com/googlegenomics/gcp-variant-transforms/issues/85)
Tested: unit tests

### DIFF
--- a/gcp_variant_transforms/beam_io/vcfio.py
+++ b/gcp_variant_transforms/beam_io/vcfio.py
@@ -711,7 +711,7 @@ class WriteToVcf(PTransform):
     self._header = headers and '\n'.join([h.strip() for h in headers]) + '\n'
 
   def expand(self, pcoll):
-    return pcoll | 'Write to VCF' >> textio.WriteToText(
+    return pcoll | 'WriteToVCF' >> textio.WriteToText(
         self._file_path,
         append_trailing_newlines=False,
         num_shards=self._num_shards,
--- a/gcp_variant_transforms/beam_io/vcfio.py
+++ b/gcp_variant_transforms/beam_io/vcfio.py
@@ -711,7 +711,7 @@ class WriteToVcf(PTransform):
     self._header = headers and '\n'.join([h.strip() for h in headers]) + '\n'
 
   def expand(self, pcoll):
-    return pcoll | textio.WriteToText(
+    return pcoll | 'Write to VCF' >> textio.WriteToText(
         self._file_path,
         append_trailing_newlines=False,
         num_shards=self._num_shards,
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -1,0 +1,72 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+r"""Pipeline for downloading BigQuery table to one VCF file.
+
+Run locally:
+python -m gcp_variant_transforms.bq_to_vcf \
+  --output_file <path to VCF file> \
+  --input_table projectname:bigquerydataset.tablename
+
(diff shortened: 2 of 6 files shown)


gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -711,7 +711,7 @@ def __init__(self,
     self._header = headers and '\n'.join([h.strip() for h in headers]) + '\n'
 
   def expand(self, pcoll):
-    return pcoll | textio.WriteToText(
+    return pcoll | 'Write to VCF' >> textio.WriteToText(
         self._file_path,
         append_trailing_newlines=False,
         num_shards=self._num_shards,
```
allieychen: FYI: `num_shards` is the number of output files. The default is 1, and we expect the output will be only one VCF file.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def run(argv=None):
+  # type: (List[str]) -> None
+  """Runs BigQuery to VCF pipeline."""
+  logging.info('Command: %s', ' '.join(argv or sys.argv))
+  known_args, pipeline_args = vcf_to_bq_common.parse_args(argv,
+                                                          _COMMAND_LINE_OPTIONS)
+  bq_source = bigquery.BigQuerySource(
+      query=''.join(['select * from `',
```
arostamianfar: nit: please extract the query as a template like:
```
_BASE_QUERY_TEMPLATE = 'SELECT * FROM {INPUT_TABLE};'

def run(...):
  query = _BASE_QUERY_TEMPLATE.format(INPUT_TABLE=table)
```

Longer term, we want to let the user customize which info/format fields and which region(s) they actually want to extract, so the query gets more complex.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -401,3 +401,49 @@ def add_arguments(self, parser):
...
+  """Options for BigQuery to VCF pipelines."""
+
+  def add_arguments(self, parser):
+    # type: (argparse.ArgumentParser) -> None
+    parser.add_argument(
+        '--output_file',
+        required=True,
+        help='The full path of the VCF file to store the result.')
+    parser.add_argument(
+        '--input_table',
```
arostamianfar: Just FYI: longer term we also want to support wildcard tables (for partitioned use case). I don't think anything needs to be changed here though.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -401,3 +401,49 @@ def add_arguments(self, parser):
...
+        required=True,
+        help=('BigQuery table that will be downloaded. It must be in the '
```
arostamianfar: s/downloaded/loaded to VCF

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -0,0 +1,72 @@
...
+                     known_args.input_table.replace(':', '.', 1),
```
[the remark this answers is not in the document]
allieychen: Thanks for pointing out the problem. Created a helper method in `bigquery_util`.

gcp_variant_transforms/transforms/bigquery_to_variant.py
```diff
@@ -0,0 +1,109 @@
...
+class BigQueryToVariant(beam.PTransform):
+  """Transforms BigQuery table rows to PCollection of `Variant`."""
+
+  def expand(self, pcoll):
+    return (pcoll | 'BigQueryToVariant' >> beam.Map(
+        self._convert_bq_row_to_variant))
+
+  def _convert_bq_row_to_variant(self, row):
+    # type: (Dict[str, Any]) -> vcfio.Variant
+    return vcfio.Variant(
```
arostamianfar: I just noticed that this should be in a separate module (similar to VCF to BQ where the BQ row generation is inside /libs) so that it can be reused for other use cases. Also one of the PTransform 'styleguide'. Please just add a TODO for now.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -0,0 +1,74 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+r"""Pipeline for downloading BigQuery table to one VCF file.
+
+Run locally:
```
arostamianfar: nit: Please add a disclaimer that this is experimental for now.

Issue #205, opened 2018-04-23
Unrecognized flags should stop the execution early.


bashir2:
Currently, because some of the arguments are Beam specific, we do not fail on unrecognized flags. But we should check the remaining flags and if there is any that is not recognized by Beam either, then we should fail the execution early and point out the error to the user.

arostamianfar:
Bumping up the priority as users have been hit by this issue multiple times (e.g. typo in the flag names).

PR #302, opened 2018-07-24
raise error on unknown flags  #205

allieychen:
Raise an error and fail the pipeline if there are unrecognized flags.

Issue: [#205](https://github.com/googlegenomics/gcp-variant-transforms/issues/205)
Tested: unit tests & integration tests

### DIFF
--- a/gcp_variant_transforms/vcf_to_bq_common.py
+++ b/gcp_variant_transforms/vcf_to_bq_common.py
@@ -25,6 +25,7 @@ import enum
 import apache_beam as beam
 from apache_beam import pvalue  # pylint: disable=unused-import
 from apache_beam.io import filesystems
+from apache_beam.options import pipeline_options
 
 from gcp_variant_transforms.beam_io import vcf_header_io
 from gcp_variant_transforms.transforms import merge_headers
@@ -33,6 +34,7 @@ from gcp_variant_transforms.transforms import merge_headers
 # headers will be merged in beam.
 _SMALL_DATA_THRESHOLD = 100
 _LARGE_DATA_THRESHOLD = 50000
+_COMMAND_LINE_ARG_PREFIX = '--'
 
 
 class PipelineModes(enum.Enum):
@@ -59,6 +61,7 @@ def parse_args(argv, command_line_options):
   known_args, pipeline_args = parser.parse_known_args(argv)
   for transform_options in options:
     transform_options.validate(known_args)
+  _raise_error_on_unrecognized_flags(pipeline_args)
   return known_args, pipeline_args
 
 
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/vcf_to_bq_common_test.py
```diff
@@ -87,3 +87,21 @@ def test_default_optimize_for_large_inputs(self):
     with mock.patch.object(FileSystems, 'match', return_value=[match]):
       self.assertEqual(vcf_to_bq_common.get_pipeline_mode(args.input_pattern),
                        PipelineModes.MEDIUM)
+
+  def test_fail_on_unrecognized_flags_1(self):
```
arostamianfar: nit: instead of '1' and '2', consider renaming to `test_fail_on_unrecognized_flags` and `test_fail_on_unrecognized_flags_no_failure`.

gcp_variant_transforms/vcf_to_bq_common.py
```diff
@@ -115,3 +118,19 @@ def write_headers(merged_header, file_path):
...
   _ = (merged_header | 'WriteHeaders' >>
        vcf_header_io.WriteVcfHeaders(file_path))
+
+
+def _raise_error_on_unrecognized_flags(pipeline_args):
+  # type: (List[str]) -> None
+  """Raises an error if there are unrecognized flags."""
+  options = pipeline_options.PipelineOptions(pipeline_args).get_all_options()
+  for flag in _get_flag_names(pipeline_args):
+    if not any(str.startswith(option, flag) for option in options):
```
arostamianfar: nit: any reason to not use `option.startswith(flag)`? it's slightly more readable.

arostamianfar: Please add a short comment why you're doing startswith instead of equals.

gcp_variant_transforms/vcf_to_bq_common.py
```diff
@@ -115,3 +118,19 @@ def write_headers(merged_header, file_path):
...
+      raise ValueError('The flag {} is unrecognized.'.format(flag))
+
+
+def _get_flag_names(pipeline_args):
+  # type: (List[str]) -> List[str]
+  """Returns a list of flag names that starts with `COMMAND_LINE_ARG_PREFIX`."""
+  return [arg[len(_COMMAND_LINE_ARG_PREFIX):] for arg in pipeline_args
```
arostamianfar: nit: I think it's easier if you just do `arg.strip('-')` so that it covers both `-f` and `--flag` (none of the existing beam options have a single dash, but they could). It would work as long as you don't have a flag ending with a `-`, but I think that's ok.

arostamianfar: ah, actually, there's an `lstrip('-')` command that does exactly that :)

allieychen: Cool! Thanks!

gcp_variant_transforms/vcf_to_bq_common_test.py
allieychen: Done. Thanks.

gcp_variant_transforms/vcf_to_bq_common.py
allieychen: Right. Updated.

allieychen: Done.

PR #304, opened 2018-07-25
Update large inputs doc to be more intuitive and complete.

arostamianfar:
Preview: https://github.com/arostamianfar/gcp-variant-transforms/blob/li/docs/large_inputs.md

### DIFF
--- a/docs/large_inputs.md
+++ b/docs/large_inputs.md
@@ -18,7 +18,20 @@ Default settings:
   --partition_config_path <default None> \
 ```
 
-### Important note about adjusting quota
+### Important notes
+
+#### Running preprocessor/validator tool
+
+Since processing large inputs can take a long time and can be costly, we highly
+recommend running the [preprocessor/validator tool](vcf_files_preprocessor.md)
+prior to loading the full VCF to BigQuery pipeline to find out about any
+invalid/inconsistent records. This can avoid failures due to invalid records
+and can save time/cost. Depending on the quality of the input files, you may
+consider running with `--report_all_conflicts` to get the full report (it takes
+longer, but is more accurate and is highly recommended when you're not sure
+about the quality of the input files).
+
+#### Adjusting quota
 
 Compute Engine enforces quota on maximum amount of resources that can be used
 at any time for variety of reasons. Please see
--- a/docs/large_inputs.md
+++ b/docs/large_inputs.md
@@ -48,7 +48,7 @@ The main Compute Engine quotas to be adjusted are:
 
 Note that the value assigned to these quotas will be the upper limit of
 available resources for your job. For example, if the quota for
-`In-use IP addresses` is 20, but you try to run with `--max_num_workers 20`,
+`In-use IP addresses` is 10, but you try to run with `--max_num_workers 20`,
 your job will be running with at most 10 workers because that's all your GCP
 project is allowed to use.
 
--- a/docs/large_inputs.md
+++ b/docs/large_inputs.md
@@ -1,48 +1,75 @@
 # Handling large inputs
 
 The Variant Transforms pipeline can process hunderds of thousands of files,
-millions of samples, and billions of records. Please consider adjusting the
-default settings as described below when processing any dataset that is more
-than a few gigabytes or contains more than 50,000 files (even if the individual
-files are small).
+millions of samples, and billions of records. There are a few settings that
+may need to be adjusted depending on the size of the dataset. Each of these
+settings are explained in the sections below.
 
-In addition, it is recommended to run the pipeline with
-`--optimize_for_large_inputs` to process large datasets more efficiently.
+Default settings:
 
-## Numer of workers
+```
+/opt/gcp_variant_transforms/bin/vcf_to_bq ... \
+  --optimize_for_large_inputs <default false> \
+  --max_num_workers <default is automatically determined> \
+  --worker_machine_type <default n1-standard-1> \
+  --disk_size_gb <default 250> \
+  --worker_disk_type <default PD> \
+  --num_bigquery_write_shards <default 1> \
+  --partition_config_path <default None> \
+```
 
-The default maximum number of workers is 15. You can adjust this by adding
-`--max_num_workers <num workers>` to the pipeline args.
-The actual number of workers will be dynamically adjusted based on Dataflow's
-autoscaling algorithm. If you prefer to not use the autoscaling algorithm, you
-may set `--num_workers <num workers>` to use the exact
-number of workers.
+## Important note about adjusting quota
 
-### Adjusting quota
+Compute Engine enforces quota on maximum amount of resources that can be used
+at any time for variety of reasons. Please see
+https://cloud.google.com/compute/quotas for more details. As a result, you may
+need to adjust the quota to satisfy the job requirements.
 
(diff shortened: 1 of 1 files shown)


docs/large_inputs.md
```diff
@@ -1,63 +1,127 @@
...
-and `num_workers * 250GB` of disk in the zone you are running the pipeline.
-Please see https://cloud.google.com/compute/quotas for details on how to
-adjust your quota.
+The main Compute Engine quotas to be adjusted are:
+* `In-use IP addresses`: One per worker.
+* `CPUs`: At least one per worker. More if larger machine type is used.
+* `Persistent Disk Standard (GB)`: At least 250GB per worker. More if larger
+  disk is used.
+* `Persistent Disk SSD (GB)`: Only needed if `--worker_disk_type` is set to SSD.
+  Required quota size is the same as `Persistent Disk Standard`.
```
samanvp: It would be nice if we could somehow explain these will be the upper limits of available resources, for example:

> The value assigned to these quotas will be the upper limit of available resources for your job. For example, if you set `In-use IP addresses` to 10 while set `--max_num_workers 20` your job will be running with at most 10 workers because that's all your GCP project is allowed to use.

docs/large_inputs.md
```diff
@@ -1,63 +1,127 @@
...
 
-Note: the `+1` in the above figures is to account for the additional VM used as
-the "pipeline orchastrator" when using the
-[pipelines API with the docker image](../README.md#using-docker).
+## `--optimize_for_large_inputs`
```
samanvp: Perhaps we can make all the following heading one level smaller (###) so it will be more obvious they are sub items of the main heading.

docs/large_inputs.md
```diff
@@ -1,63 +1,127 @@
...
+This flag should be set to true when loading more than 50,000 files and/or
+[merging](variant_merging.md) is enabled for a large number (>3 billion)
+of variants. This flag optimizes the Dataflow pipeline for large inputs, which
+can save significant cost/time, but the additional overhead may hurt cost/time
+for small inputs.
+
+## `--max_num_workers`
+
+By default, Dataflow uses its autoscaling algorithm to adjust the number of
+workers assigned to each job. You may adjust the maximum number of workers using
```
samanvp: > workers assigned to each job to reach a limit set by Compute Engine quota

docs/large_inputs.md
```diff
@@ -1,63 +1,127 @@
...
-https://cloud.google.com/compute/pricing#predefined_machine_types for a list of
-available machine types.
+3.75GB of RAM. You may need to request a larger machine for large datasets.
+Please see https://cloud.google.com/compute/pricing#predefined_machine_types
+for a list of available machine types.
+
+We have observed that Dataflow performs more efficiently when running
+with a large number of small machines rather than a small number of large
+machines (e.g. 200 `n1-standard-4` workers instead of 25 `n1-standard-32`
+workers). This is due to disk/network IOPS (input/output operations per second)
```
samanvp: (e.g. having `CPUs` quota set to 800, 200 `n1-standard-4` workers instead of 25 `n1-standard-32` workers)

docs/large_inputs.md
```diff
@@ -1,63 +1,127 @@
...
+
+In addition, if [merging](variant_merging.md) or
+[--num_bigquery_write_shards](#--num_bigquery_write_shards) is enabled, you may
+need more disk per worker (e.g. 500GB) as the same variants need to be
+aggregated together on one machine.
+
+## `--worker_disk_type`
+
+SSDs provide significantly more IOPS than standard persistent disks, but are
+more expensive. However, when choosing a large machine (e.g. `n1-standard-16`),
```
samanvp: You referred to a large machine earlier as `n1-standard-8`, I think it's better to be consistent.

docs/large_inputs.md
```diff
@@ -1,63 +1,127 @@
...
+they can reduce cost as they can avoid idle CPU cycles due to disk IOPS
+limitations.
+
+As a result, we recommend using SSDs if [merging](variant_merge.md) or
+[--num_bigquery_write_shards](#--num_bigquery_write_shards) is enabled: these
+operations require "shuffling" the data (i.e. redistributing the data among
+workers), which require significant disk I/O.
+
+Set
+`--worker_disk_type compute.googleapis.com/projects//zones//diskTypes/pd-ssd`
```
samanvp: On a not related note to this PR:
Can't we isolate the users from this strange flag value by having a flag such as --use_ssd_for_workers and then we set the internal variable to `compute.googleapis.com/projects//zones//diskTypes/pd-ssd`?

docs/large_inputs.md
```diff
@@ -1,63 +1,127 @@
...
+As a result, we recommend setting `--num_bigquery_write_shards 20` when loading
+any data that has more than 1 billion rows (after merging) or 1TB of final
+output. Note that using a larger value (e.g. 50) can cause BigQuery write to
+fail as there is a maximum limit on the number of concurrent writes per table.
+
+## `--partition_config_path`
 
-## Pipeline resource estimator tool
+Partitioning the output can save significant query costs once the data is in
+BigQuery. It can also optimize the pipeline (e.g. it natively shards the
```
samanvp: It can also optimize the cost/time of the pipeline

samanvp:
Also please feel free to ignore my comments if you think they don't make the overall doc any better.

docs/large_inputs.md
arostamianfar: Good point. Updated!

docs/large_inputs.md
arostamianfar: ah, that one was "larger than" 8 :p rephrased... :)

docs/large_inputs.md
arostamianfar: I agree that this is strange and that's a good suggestion; but like to avoid duplicating flags as much as possible (e.g. if there is a new disk type, we'd need to add another flag for it ourselves). I think this is a rare enough use case that it's ok to ask users to provide this flag for now.

docs/large_inputs.md
```diff
@@ -1,63 +1,127 @@
 # Handling large inputs
 
 The Variant Transforms pipeline can process hunderds of thousands of files,
-millions of samples, and billions of records. Please consider adjusting the
-default settings as described below when processing any dataset that is more
-than a few gigabytes or contains more than 50,000 files (even if the individual
-files are small).
+millions of samples, and billions of records. There are a few settings that
+may need to be adjusted depending on the size of the dataset. Each of these
```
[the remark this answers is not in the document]
arostamianfar: Good idea! Updated.

docs/large_inputs.md
arostamianfar: Done.

docs/large_inputs.md
arostamianfar: Updated the text.

docs/large_inputs.md
arostamianfar: I do like to avoid coupling this point with quota here; as this can be independent of the quota (project can still have more CPUs but user may want to just use 800 CPUs).

docs/large_inputs.md
arostamianfar: Done.

samanvp:
Everything looks good, thanks again Asha!

PR #309, opened 2018-07-26
write vcf files based on key and compose files to one

allieychen:
1. replaced the textio.WriteToText with file_system.
2. Generates a separate file for variants with the same key in `bucket_name`.
3. Combine the files to the `output_file` provided at the end.

Issue: [issue 85](https://github.com/googlegenomics/gcp-variant-transforms/issues/85)
Tested: unit tests & manual pipeline test

### DIFF
--- a/gcp_variant_transforms/beam_io/vcfio.py
+++ b/gcp_variant_transforms/beam_io/vcfio.py
@@ -19,6 +19,9 @@ The 4.2 spec is available at https://samtools.github.io/hts-specs/VCFv4.2.pdf.
 
 from __future__ import absolute_import
 
+from typing import Dict, Iterable, List, Optional, Tuple  # pylint: disable=unused-import
+import logging
+from collections import namedtuple
 from functools import partial
 
 import vcf
@@ -446,7 +449,6 @@ class ReadAllFromVcf(PTransform):
     return pvalue | 'ReadAllFiles' >> self._read_all_files
 
 
-
 class WriteToVcf(PTransform):
   """A PTransform for writing to VCF files."""
 
@@ -494,13 +496,14 @@ class WriteToVcf(PTransform):
 class _WriteVcfDataLinesFn(beam.DoFn):
   """A function that writes variants to one VCF file."""
 
-  def process(self, file_path_and_variants, *args, **kwargs):
-    # type: (tuple[str, list]) -> None
-    file_path, variants = file_path_and_variants
-    coder = _ToVcfRecordCoder()
+  def __init__(self):
+    self._coder = _ToVcfRecordCoder()
+
+  def process(self, (file_path, variants), *args, **kwargs):
+    # type: (Tuple[str, List[Variant]]) -> None
     with filesystems.FileSystems.create(file_path) as file_to_write:
       for variant in variants:
-        file_to_write.write(coder.encode(variant))
+        file_to_write.write(self._coder.encode(variant))
 
 
 class WriteVcfDataLines(PTransform):
--- a/gcp_variant_transforms/beam_io/vcfio.py
+++ b/gcp_variant_transforms/beam_io/vcfio.py
@@ -507,7 +507,11 @@ class _WriteVcfDataLinesFn(beam.DoFn):
 
 
 class WriteVcfDataLines(PTransform):
-  """A PTransform for writing VCF data lines."""
+  """A PTransform for writing VCF data lines.
 
+  This PTransform takes PCollection<`file_path`, `variants`> as input, and
+  writes `variants` to `file_path`. The PTransform `WriteToVcf` takes
+  PCollection<`Variant`> as input, and writes all variants to the same file.
+  """
   def expand(self, pcoll):
     return pcoll | 'WriteToVCF' >> beam.ParDo(_WriteVcfDataLinesFn())
--- a/gcp_variant_transforms/beam_io/vcfio.py
+++ b/gcp_variant_transforms/beam_io/vcfio.py
@@ -171,6 +171,124 @@ class _ToVcfRecordCoder(coders.Coder):
     return MISSING_FIELD_VALUE
 
 
+class _ToVcfRecordCoder(coders.Coder):
+  """Coder for encoding :class:`Variant` objects as VCF text lines."""
+
+  def encode(self, variant):
+    # type: (Variant) -> str
+    """Converts a :class:`Variant` object back to a VCF line."""
+    encoded_info = self._encode_variant_info(variant)
+    format_keys = self._get_variant_format_keys(variant)
+    encoded_calls = self._encode_variant_calls(variant, format_keys)
+
+    columns = [
+        variant.reference_name,
+        None if variant.start is None else variant.start + 1,
+        ';'.join(variant.names),
+        variant.reference_bases,
+        ','.join(variant.alternate_bases),
+        variant.quality,
+        ';'.join(variant.filters),
+        encoded_info,
+        ':'.join(format_keys),
+    ]
+    if encoded_calls:
+      columns.append(encoded_calls)
+    columns = [self._encode_value(c) for c in columns]
+
+    return '\t'.join(columns) + '\n'
+
+  def _encode_value(self, value):
+    """Encodes a single :class:`Variant` column value for a VCF file line."""
+    if not value and value != 0:
+      return MISSING_FIELD_VALUE
+    elif isinstance(value, list):
+      return ','.join([self._encode_value(x) for x in value])
+    return str(value)
+
+  def _encode_variant_info(self, variant):
+    """Encodes the info of a :class:`Variant` for a VCF file line."""
+    encoded_infos = []
+    # Set END in info if it doesn't match start+len(reference_bases). This is
+    # usually the case for non-variant regions.
+    if (variant.start is not None
+        and variant.reference_bases
+        and variant.end
+        and variant.start + len(variant.reference_bases) != variant.end):
+      encoded_infos.append('END=%d' % variant.end)
+    # Set all other fields of info.
+    for k, v in variant.info.iteritems():
+      if v is True:
+        encoded_infos.append(k)
+      else:
+        encoded_infos.append('%s=%s' % (k, self._encode_value(v)))
+    return ';'.join(encoded_infos)
+
+  def _get_variant_format_keys(self, variant):
+    """Gets the format keys of a :class:`Variant`."""
+    if not variant.calls:
+      return []
+
+    format_keys = [GENOTYPE_FORMAT_KEY]
+    for call in variant.calls:
+      # If any calls have a set phaseset that is not `DEFAULT_PHASESET_VALUE`,
+      # the key will be added to the format field.
+      if self._is_alternate_phaseset(call.phaseset):
+        format_keys.append(PHASESET_FORMAT_KEY)
+      format_keys.extend([k for k in call.info])
+
+    # Sort all keys and remove duplicates after GENOTYPE_FORMAT_KEY
+    format_keys[1:] = sorted(list(set(format_keys[1:])))
+
+    return format_keys
+
+  def _encode_variant_calls(self, variant, format_keys):
+    """Encodes the calls of a :class:`Variant` in a VCF line."""
+    # Ensure that genotype is always the first key in format_keys
+    assert not format_keys or format_keys[0] == GENOTYPE_FORMAT_KEY
+    encoded_calls = []
+    for call in variant.calls:
+      encoded_call_info = [self._encode_genotype(call.genotype, call.phaseset)]
+      for key in format_keys[1:]:
+        if key == PHASESET_FORMAT_KEY:
+          encoded_call_info.append(self._encode_phaseset(call.phaseset))
+        else:
+          encoded_call_info.append(self._encode_call_info_value(call.info, key))
+
+      encoded_calls.append(':'.join(encoded_call_info))
+
+    return '\t'.join(encoded_calls)
+
+  def _encode_genotype(self, genotype, phaseset):
+    """Encodes the genotype of a :class:`VariantCall` for a VCF file line."""
+    if genotype == MISSING_GENOTYPE_VALUE:
+      return MISSING_FIELD_VALUE
+    encoded_genotype = []
+    for allele in genotype:
+      if allele == MISSING_GENOTYPE_VALUE:
+        encoded_genotype.append(MISSING_FIELD_VALUE)
+      else:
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -56,19 +63,42 @@ def run(argv=None):
...
+                                            _NUMBER_OF_VARIANTS_PER_SHARD)
+         | 'GroupVariantsByKey' >> beam.GroupByKey()
+         | vcfio.WriteVcfDataLines(file_path_prefix))
+
+  client = storage.Client(project)
+  bucket = client.get_bucket(known_args.bucket_name)
+  output_file_blob = bucket.blob(known_args.output_file)
+  output_file_blob.content_type = 'text/plain'
+  # Compose all VCF files generated in this pipeline into one.
+  output_file_blob.compose(bucket.list_blobs(prefix=blob_prefix))
```
allieychen: Note: There is a limit (currently 32) to the number of components that can be composed in a single operation for [compose](https://cloud.google.com/storage/docs/gsutil/commands/compose). To merge a large number of files, possible solutions can be found in [here](https://stackoverflow.com/questions/43024522/how-to-concatenate-a-million-files-in-google-cloud-storage), which I plan to try in my next PR.

allieychen: Updated the code.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -48,6 +54,9 @@
 
 _BASE_QUERY_TEMPLATE = 'SELECT * FROM `{INPUT_TABLE}`;'
 _COMMAND_LINE_OPTIONS = [variant_transform_options.BigQueryToVcfOptions]
+# The maximum number of variants to include in a single VCF file.
+_NUMBER_OF_VARIANTS_PER_SHARD = 1000
```
arostamianfar: I think this should be (1) a flag with a (potentially larger) default value and (2) named more generally as # of variants is misleading (I realize you mention 'maximum', but it can still lead to confusion about how this value is used; IOW, ppl may think that we split based on the number of variants rather than regions). What is the name of this variable in other contexts? I think I've see the word `bucket` used previously.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -56,19 +65,78 @@ def run(argv=None):
   logging.info('Command: %s', ' '.join(argv or sys.argv))
   known_args, pipeline_args = vcf_to_bq_common.parse_args(argv,
                                                           _COMMAND_LINE_OPTIONS)
+  (project, dataset, table) = bigquery_util.parse_table_reference(
```
arostamianfar: nit: you dont need brackets around `(project, dataset, table)`

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -56,19 +65,78 @@ def run(argv=None):
...
+  file_path_prefix = '/'.join(['gs:/', known_args.bucket_name, blob_prefix])
   with beam.Pipeline(options=options) as p:
     _ = (p | 'ReadFromBigQuery ' >> beam.io.Read(bq_source)
          | bigquery_to_variant.BigQueryToVariant()
          | densify_variants.DensifyVariants()
-         | vcfio.WriteToVcf(known_args.output_file))
+         | 'PairVariantWithKey' >> beam.Map(_pair_variant_with_key,
+                                            _NUMBER_OF_VARIANTS_PER_SHARD)
+         | 'GroupVariantsByKey' >> beam.GroupByKey()
+         | vcfio.WriteVcfDataLines(file_path_prefix))
```
arostamianfar: I think the interface of WriteVcfDataLines class is confusing as it accepts a list and a key to use to generate the file name (it should really just get a list of variants and the full path of the file name). What do you think of introducing another simple transform after GroupByKey to generate the full file path, sort the variants, and then pass `(file_path, variants)` to a DoFn that actually writes the file?
Combined with my earlier suggestion of keeping the Coder class, the DoFn will just be a simple for loop over variants as it already has the full file path in it.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -438,9 +438,13 @@ def add_arguments(self, parser):
...
         '--output_file',
         required=True,
-        help='The full path of the VCF file to store the result.')
+        help='The path of the VCF file to store the result inside the bucket.')
     parser.add_argument(
         '--input_table',
         required=True,
         help=('BigQuery table that will be loaded to VCF. It must be in the '
               'format of (PROJECT:DATASET.TABLE).'))
+    parser.add_argument(
```
arostamianfar: I think this API is a bit confusing; what do you think of using the `--temp_location` flag (just dynamically create a new folder under there), store all of the intermediate VCF files there, and then combine them into the output_file. That way, the user ends up with a single file under output_file, and it would be atomic (i.e. the file would only be copied if all intermediate ones were successful). This is actually the idea that BQ I/O uses as well.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -56,19 +65,78 @@ def run(argv=None):
...
+
+  Args:
+    bucket: the bucket in which the VCF files will be composed.
+    blob_prefix: the prefix used to filter blobs. Only the VCF files with this
+      prefix will be composed.
+
+  Returns:
+    The final blob that all VCFs composed to.
+  """
+  blobs_to_be_composed = list(bucket.list_blobs(prefix=blob_prefix))
```
arostamianfar: to clarify: does `list_blobs` return the files in a sorted order?

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -56,19 +65,78 @@ def run(argv=None):
...
+
+
+def _pair_variant_with_key(variant, number_of_variants_per_shard):
+  return ('_%s_%011d' % (variant.reference_name,
+                         variant.start / number_of_variants_per_shard *
+                         number_of_variants_per_shard),
+          variant)
+
+
+def _compose_multiple_files_to_one(bucket, blob_prefix):
```
arostamianfar: Very cool to get this working in code!
How long does this take for a whole genome?

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -56,19 +65,78 @@ def run(argv=None):
...
     _ = (p | 'ReadFromBigQuery ' >> beam.io.Read(bq_source)
          | bigquery_to_variant.BigQueryToVariant()
          | densify_variants.DensifyVariants()
-         | vcfio.WriteToVcf(known_args.output_file))
+         | 'PairVariantWithKey' >> beam.Map(_pair_variant_with_key,
+                                            _NUMBER_OF_VARIANTS_PER_SHARD)
+         | 'GroupVariantsByKey' >> beam.GroupByKey()
+         | vcfio.WriteVcfDataLines(file_path_prefix))
+
+  client = storage.Client(project)
```
arostamianfar: Please extract this logic into another method.

gcp_variant_transforms/bq_to_vcf.py
allieychen: I tested on a table with 10,000 variants, with 3865 blobs initially generated (I guess the number `1000` max variants is way to small for this case, since there are only 3 variants on average in each file), it takes about 20 minutes to do the compose. I am also interested to know whether there is any parallel solutions for it.

gcp_variant_transforms/bq_to_vcf.py
allieychen: Thanks! I added a flag as number_of_bases_per_shard (same as in the Java version) with default value 10000.

gcp_variant_transforms/options/variant_transform_options.py
allieychen: Thanks! My concern is that if using `--temp_location`, then even for the local runner this flag is required, which is not the case right now. 

By providing the `bucket_name`, all intermediate files are stored here (we can create an temp file). It is also atomic since the final VCF file will be formed by composing the meta-information file, data header file, and one composed VCF file (not implemented yet).

Another reason I replaced the full path of the output_file with `blob_name`(I didn't change the flag to blob_name though) and `bucket_name` is that it seems that there is no parser to parse a gs link (e.g., gs://bucket_name/blob_name), but we need those for composing purposes.

arostamianfar: For DirectRunner: you can just create a tempfolder; we already do this for merging headers (see https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/vcf_to_bq.py#L169). You do raise an important point though that for DirectRunner, we shouldn't use GCS compose and probably safe to set the bucket size to max value (i.e. just one per chromosome) as the expectation is that you're running with small amount of data. Note that DirectRunner support is important as (depending on how efficient it is), we would use it for htsget and there shouldn't be any dependency on GCS.

For gs parser: I suspect you wouldn't need it if you just use the `--temp_location` flag as you can create that yourself?

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -678,45 +559,160 @@ def expand(self, pvalue):
...
+    # type: (tuple[str, list]) -> None
+    (key, variants) = key_and_variants
+    variants = sorted(list(variants), key=lambda a: a.start)
+
+    file_name = ''.join([self._file_prefix, key, self._FILE_EXTENSION])
+    with filesystems.FileSystems.create(file_name) as file_to_write:
+      for variant in variants:
+        file_to_write.write(self._variant_to_vcf_line(variant))
+
+  def _variant_to_vcf_line(self, variant):
```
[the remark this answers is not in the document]
arostamianfar: yeah, the `vcfio.py` file is huge. We mainly did that based on Beam's style (all I/Os are one file), but we can consider splitting them up here and merge it back when syncing with the Beam SDK (someday!). For now, feel free to put it inside vcfio.py though.

gcp_variant_transforms/options/variant_transform_options.py
allieychen: Thanks. I removed the `bucket_name` flag as suggested. I happened to find a gcs path parser in `gcsio`, which solves the parsing problem.
FYI, the [gsutil ](https://github.com/GoogleCloudPlatform/gsutil) also supports parsing and provides various cloud storage APIs. But we would end up with dependency compatibility issues if using it. To be more specific, `apache-beam` requires `google-apitools >=0.5.18, < 0.5.20`, where `gsutil 4.28` depends on `google-apitools ==0.5.16` and `gsutil 4.29` requires `google-apitools == 0.5.21`.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -720,3 +719,23 @@ def expand(self, pcoll):
...
+
+  def process(self, (file_path, variants), *args, **kwargs):
+    # type: (Tuple[str, List[Variant]]) -> None
+    with filesystems.FileSystems.create(file_path) as file_to_write:
+      for variant in variants:
+        file_to_write.write(self._coder.encode(variant))
+
+
+class WriteVcfDataLines(PTransform):
+  """A PTransform for writing VCF data lines."""
```
arostamianfar: Please add a brief note about how this is different from WriteToVcf transform above now that we're keeping both.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -56,19 +63,137 @@ def run(argv=None):
...
+  bq_to_vcf_temp_folder = filesystems.FileSystems.join(
+      google_cloud_options.temp_location, shards_folder)
+
+  _bigquery_to_vcf_shards(known_args, options, bq_to_vcf_temp_folder)
+  _compose_temp_files(google_cloud_options.project,
+                      known_args.output_file,
+                      bq_to_vcf_temp_folder)
+
+
+def _bigquery_to_vcf_shards(known_args, options, bq_to_vcf_temp_folder):
```
arostamianfar: nit: s/options/pipeline_options as the context is not clear in this API.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -56,19 +63,137 @@ def run(argv=None):
...
+
+  Args:
+    file_name: The file name that associated with the `variants`, which is used
+      to form the file path where the `variants` are written to.
+    variants: A collection of variants within a contiguous region of the genome.
+    file_path_prefix: The common file prefix for all VCF files generated in the
+      pipeline. The files written will begin with this prefix, followed by the
+      `file_name`.
+  """
+  file_path = '/'.join([file_path_prefix, file_name])
```
arostamianfar: Please use FileSystems.join here as well.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -56,19 +63,137 @@ def run(argv=None):
...
+  Returns:
+    The final blob that all VCFs composed to.
+  """
+  blobs_to_be_composed = list(bucket.list_blobs(prefix=blob_prefix))
+  if len(blobs_to_be_composed) == 1:
+    return blobs_to_be_composed[0]
+  new_blob_prefix = filesystems.FileSystems.join(blob_prefix, 'composed_')
+  for blobs_chunk in (_break_list_in_chunks(blobs_to_be_composed,
+                                            _MAX_NUM_OF_BLOBS_PER_COMPOSE)):
+    _, file_name = os.path.split(blobs_chunk[0].name)
```
arostamianfar: nit: does FileSystem(s) have a method for this?

gcp_variant_transforms/bq_to_vcf.py
allieychen: I didn't use FileSystems here since it throws a 
`NameError: global name 'filesystems' is not defined [while running 'ParDo(CallableWrapperDoFn)']`
I have to explicitly import `filesystems` inside this method to avoid the above error. Please let me know if you still prefer use `FileSystems.join`

gcp_variant_transforms/bq_to_vcf.py
allieychen: Thank you! Nice advice!

gcp_variant_transforms/bq_to_vcf.py
arostamianfar: ah i see...let's continue to use FileSystems to be consistent. 
You can import it within the method and disable pylint warnings `pylint:disable=wrong-import-order,wrong-import-position`

Issue #305, opened 2018-07-25
Let the user know when both --infer_headers and --representative_header_file is passed


arostamianfar:
In this case, we'll ignore --infer_headers arg and use --representative_header_file. We should warn the user that this is happening (likely fail the pipeline).

PR #310, opened 2018-07-26
Add warning

allieychen:
Add warnings when both infer_headers and representative_header_file are used.

Issues: [issue 305](https://github.com/googlegenomics/gcp-variant-transforms/issues/305)
Tested: unit tests

### DIFF
--- a/gcp_variant_transforms/options/variant_transform_options.py
+++ b/gcp_variant_transforms/options/variant_transform_options.py
@@ -15,6 +15,7 @@
 from __future__ import absolute_import
 
 import argparse  # pylint: disable=unused-import
+import logging
 import re
 
 from apache_beam.io.gcp.internal.clients import bigquery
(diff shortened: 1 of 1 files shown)


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -91,6 +92,14 @@ def add_arguments(self, parser):
               'variants not defined in the header sections, or the definition '
               'of the the header fields do not match the field values.'))
 
+  def validate(self, parsed_args):
+    # type: (argparse.Namespace) -> None
+    if parsed_args.infer_headers and parsed_args.representative_header_file:
+      logging.warning('Both --infer_headers and --representative_header_file '
```
arostamianfar: I think we should just fail and force the user to choose the right flags. Users rarely look at warning logs :)

allieychen: Thanks Asha! It makes sense. I also have this concern. When I was testing I had to keep my eyes open since the warning was overwhelmed by lots of info logs.

PR #311, opened 2018-07-27
Add the full variant in case of errors for easier debugging.

### LINKED ISSUE #306: Report the full variant that contains malformed Number=A field
Even though we can't report the file where the variant came from*, we can at least report the full variant. This will make it easier to debug and find the file as it contains the sample name(s).

arostamianfar:
Fixes #306.

Tested: ran pipelines that generate errors and verified the logs and exceptions thrown.

P.S. pylint for logging.warning requires "%s" format, hence the difference in formatting.

### DIFF
--- a/gcp_variant_transforms/transforms/infer_headers.py
+++ b/gcp_variant_transforms/transforms/infer_headers.py
@@ -16,6 +16,7 @@
 
 from __future__ import absolute_import
 
+import logging
 from typing import Any, Dict, Iterable, List, Optional, Union  # pylint: disable=unused-import
 
 import apache_beam as beam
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -213,6 +216,8 @@ def _infer_info_fields(self, variant, defined_headers):
             defined_headers.infos.get(info_field_key),
             len(variant.alternate_bases))
         if corrected_info:
+          logging.warning('Incorrect INFO field "%s" in variant "%s"',
```
allieychen: Providing the conflicting info may also help. Something similar to `expected (the definition found in defined_headers), found (the INFO field value in variant), correct the header to (corrected_header)`.

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -252,6 +259,8 @@ def _infer_format_fields(self, variant, defined_headers):
           corrected_format = self._infer_mismatched_format_field(
               format_key, format_value, defined_headers.formats.get(format_key))
           if corrected_format:
+            logging.warning('Incorrect FORMAT field "%s" in variant "%s"',
```
allieychen: ditto.

gcp_variant_transforms/transforms/infer_headers.py
arostamianfar: ah, this was actually the conflicting info case :) Rephrased the warning msg to make it more clear (also added expected vs. actual). Thanks!

gcp_variant_transforms/transforms/infer_headers.py
arostamianfar: Done here as well.

PR #313, opened 2018-07-30
Make setup_file flag required.

arostamianfar:
setup_file is required when running w/ Dataflow for Variant Transforms. If it's not provided, no error is actually thrown and the pipeline just hangs; very difficult to debug and identify the problem.

### DIFF
--- a/gcp_variant_transforms/vcf_to_bq_common_test.py
+++ b/gcp_variant_transforms/vcf_to_bq_common_test.py
@@ -25,7 +25,7 @@ from gcp_variant_transforms.vcf_to_bq_common import PipelineModes
 
 
 class VcfToBqCommonTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/vcf_to_bq_common_test.py
```diff
@@ -24,7 +24,7 @@
 from gcp_variant_transforms.vcf_to_bq_common import PipelineModes
 
 
-class DataProcessorTest(unittest.TestCase):
+class VcfToBqCommonTest(unittest.TestCase):
   """Tests cases for the ``general_process`` script."""
```
allieychen: Can you also help update the doc string here? s/general_process/vcf_to_bq_common.

arostamianfar: Done! Thanks for catching this one too!

Issue #221, opened 2018-05-01
Add integration test to cover inf/-inf/nan values


arostamianfar:
See #219 for context. We should have integration tests for these values to ensure we can handle them properly from now on (also becomes important when we switch to Nucleus).

PR #317, opened 2018-08-03
update integration tests #221

allieychen:
1. Fixed a bug related to [#219](https://github.com/googlegenomics/gcp-variant-transforms/issues/219). If the field is defined as `num = A`, and the value is `inf`, the pipeline fails.
2. Add integration test for the 'inf' float values.

Issue: [#221](https://github.com/googlegenomics/gcp-variant-transforms/issues/221), [#219](https://github.com/googlegenomics/gcp-variant-transforms/issues/219)
Tested: unit tests & integration tests


pgrosu:
@allieychen By you already have `annotation_fields` defined as follows in `ProcessedVariant`:

https://github.com/googlegenomics/gcp-variant-transforms/blob/6d175c2e7dd0cdc6d314d5fc608ec7c046184a54/gcp_variant_transforms/libs/processed_variant.py#L199-L201

pgrosu:
By the way BigQuery is misspelled here: 

https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/libs/bigquery_util.py#L153

`BgiQuery API does not allow null values in lists (note that the entire`

allieychen:
@pgrosu Thanks for pointing out the typo!
Re `annotation_fileds` in `ProcessedVariant`: The field is defined in `ProcessedVariantFactory`, and I think it might be easier to add this field to `ProcessedVariant`. :)

arostamianfar:
@allieychen: as discussed offline, annotation fields are actually a property of alternates in the context of `ProcessedVariant` and they're really providing more context about the keys of the info fields inside the alternates. So, let's put them inside `AlternateBaseData`.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -235,6 +237,9 @@ def create_processed_variant(self, variant):
             proc_var, key, variant_info_data)
       else:
         proc_var._non_alt_info[key] = variant_info_data
+    for alternate_data in proc_var._alternate_datas:
```
arostamianfar: I think it's cleaner if this is moved to [here](https://github.com/googlegenomics/gcp-variant-transforms/blob/a15886c2a92353a5caed3d046c20bb3ba73bda1e/gcp_variant_transforms/libs/processed_variant.py#L407)
Basically, you'd add `alt_datas._annotation_field_names.add(annotation_field_name)` there.

gcp_variant_transforms/libs/bigquery_row_generator.py
```diff
@@ -162,7 +162,9 @@ def _get_base_row_from_variant(self, variant, allow_incompatible_records):
       alt_record = {bigquery_util.ColumnKeyConstants.ALTERNATE_BASES_ALT:
                     alt.alternate_bases}
       for key, data in alt.info.iteritems():
-        alt_record[bigquery_util.get_bigquery_sanitized_field_name(key)] = data
+        if key not in alt._annotation_field_names:
```
arostamianfar: Please make `alt._annotation_field_names` a property so that you can refer to it without private syntax.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -153,13 +153,15 @@ def __init__(self, alt_bases):
...
     self._info = {}  # type: Dict[str, Any]
+    self._annotation_field_names = []  # type: Set[str]
 
   def __repr__(self):
     return ', '.join([str(self._alt_bases), str(self._info)])
 
   def __eq__(self, other):
     return (isinstance(other, AlternateBaseData) and
-            vars(self) == vars(other))
+            self._alt_bases == other._alt_bases and
```
arostamianfar: Are there cases where the new `_annotation_field_names` does not work anymore?

allieychen: Some unit tests didn't pass. We can also update the unit tests, but it just seems weird and unnecessary that every time we create `AlternateBaseData` in the unit tests, we have to set the the `annotation_field_names`.

pgrosu: Don't worry @allieychen, it's just temporary until the recursive check implementation.

PR #321, opened 2018-08-09
change the null numeric value replacement #221

allieychen:
Update the default `null_numeric_value_replacement` value. 

The pipeline fails when a float value is `nan`. The reason is that the default `null_numeric_value_replacement`value (-sys.maxint) is too large to convert to a float. The converted value and the original value are incompatible.

Issues: [221](https://github.com/googlegenomics/gcp-variant-transforms/issues/221)
Tested: unit tests and manually ran `vcf_to_bq` pipeline.

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_util.py
+++ b/gcp_variant_transforms/libs/bigquery_util.py
@@ -75,6 +75,7 @@ _FALLBACK_FIELD_NAME_PREFIX = 'field_'
 # A big number to represent infinite float values. The division by 10 is to
 # prevent unintentional overflows when doing subsequent operations.
 _INF_FLOAT_VALUE = sys.float_info.max / 10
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -75,6 +75,7 @@ class TableFieldConstants(object):
 # A big number to represent infinite float values. The division by 10 is to
 # prevent unintentional overflows when doing subsequent operations.
 _INF_FLOAT_VALUE = sys.float_info.max / 10
+_DEFAULT_NULL_NUMERIC_VALUE_REPLACEMENT = -pow(2, 32)
```
arostamianfar: Please use 2^31 (it's equivalent to -sys.maxint in 32bit machines).

allieychen:
Right! Our integration tests do work great!

pgrosu:
Awesome :)

PR #322, opened 2018-08-09
VcfParser abstract class

### LINKED ISSUE #316: Provide more error details for invalid vcf formatted files
Related to #113, when encountering issues with fields could you provide more details about the error such as possibly the vcf file and line number?  E.g. the following error tells me what's wrong but not where:

```
      '@type': type.googleapis.com/google.genomics.v2alpha1.ContainerStoppedEvent
      actionId: 1
      exitStatus: 1
      stderr: |2+
           op.start()
          File "dataflow_worker/native_operations.py", line 38, in dataflow_worker.native_operations.NativeReadOperation.start
            def start(self):
          File "dataflow_worker/native_operations.py", line 39, in dataflow_worker.native_operations.NativeReadOperation.start
            with self.scoped_start_state:
          File "dataflow_worker/native_operations.py", line 44, in dataflow_worker.native_operations.NativeReadOperation.start
            with self.spec.source.reader() as reader:
          File "dataflow_worker/native_operations.py", line 48, in dataflow_worker.native_operations.NativeReadOperation.start
            for value in reader:
          File "/usr/local/lib/python2.7/dist-packages/gcp_variant_transforms/beam_io/vcfio.py", line 396, in read_records
            for line in record_iterator:
          File "/usr/local/lib/python2.7/dist-packages/gcp_variant_transforms/beam_io/vcfio.py", line 475, in next
            raise ValueError('Invalid record in VCF file. Error: %s' % str(e))
        ValueError: Invalid record in VCF file. Error: invalid literal for float(): 20,0
```

samanvp:
This PR implements a generic VcfParser class and a derived class based
on Pyvcf.
This is a preparation for a second PR in which we add the support for
Nucleus based parser.

This PR also fixes #316


gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,449 @@
...
+    # type: (...) -> None
+    # If `representative_header_lines` is given, header lines in `file_name`
+    # are ignored; refer to _process_header_lines() logic.
+    self._header_lines = []
+    self._representative_header_lines = representative_header_lines
+    self._current_line = None
+    self._file_name = file_name
+    self._allow_malformed_records = allow_malformed_records
+
+    self._header_processed = False
```
arostamianfar: This is also not needed.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,449 @@
...
+
+  def __iter__(self):
+    return self
+
+  def _init_with_header(self, header_lines):
+    # type: (List[str]) -> None
+    """ This method will be called by _process_header_lines()."""
+    raise NotImplementedError
+
+  def _get_record(self, line):
```
arostamianfar: s/line/data_line may be a better name here (but I dont feel strongly about it).
...it may be obvious, but this the body of this method should only care about parsing data lines and not header lines, so renaming the variable makes it even more clear.
and 'data lines' terminology is from VCF spec: http://samtools.github.io/hts-specs/VCFv4.3.pdf

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,449 @@
...
+    # type: (str) -> Variant
+    """ This method will be called by next(), one line at a time."""
+    raise NotImplementedError
+
+class PyvcfParser(VcfParser):
+  """An Iterator for processing a single VCF file using pyvcf."""
```
arostamianfar: nit: s/pyvcf/PyVCF

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,449 @@
...
+      yield self._current_line.encode('utf-8')
+
+  def _get_record(self, line):
+    # _line_generator will consume this line.
+    self._current_line = line
+    try:
+      record = next(self._vcf_reader)
+      return self._convert_to_variant_record(record, self._vcf_reader.formats)
+    except (LookupError, ValueError) as e:
+      if self._allow_malformed_records:
```
arostamianfar: I think it's a better design to change the 'contract' of this method (i.e. in the comments) to either return a `Variant` or a `MalformedVcfRecord` (you need to specify this now anyway). That way, you just return the MalformedVcfRecord here. The `next()` method in `VcfParser` then checks whether the type is a Variant or Malformed one and does the right thing depending on allow_malfored_records. That way, for Nucleus, you don't have to re-implement this logic.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -34,205 +29,30 @@
...
-MISSING_FIELD_VALUE = '.'  # Indicates field is missing in VCF record.
-PASS_FILTER = 'PASS'  # Indicates that all filters have been passed.
-END_INFO_KEY = 'END'  # The info key that explicitly specifies end of a record.
-GENOTYPE_FORMAT_KEY = 'GT'  # The genotype format key in a call.
-PHASESET_FORMAT_KEY = 'PS'  # The phaseset format key.
-DEFAULT_PHASESET_VALUE = '*'  # Default phaseset value if call is phased, but
-                              # no 'PS' is present.
-MISSING_GENOTYPE_VALUE = -1  # Genotype to use when '.' is used in GT field.
 
+__all__ = ['ReadFromVcf', 'ReadAllFromVcf', 'MalformedVcfRecord']
```
arostamianfar: I think you can remove this now (required by Beam, but not by our code). We kept it here for Beam-compatibility, but now we have to change a bunch of stuff anyway, so let's remove it. Ditto for vcf_parser.py.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -34,205 +29,30 @@
...
-    return ', '.join(
-        [str(s) for s in [self.name, self.genotype, self.phaseset, self.info]])
 
 
 class _ToVcfRecordCoder(coders.Coder):
   """Coder for encoding :class:`Variant` objects as VCF text lines."""
 
   def encode(self, variant):
-    # type: (Variant) -> str
+    # type: (vcf_parser.Variant) -> str
```
arostamianfar: Let's keep everything as 'Variant' now (it would be valid as it's still declared here). We'd need to change everything if we want to use the vcf_parser version not just this particular one.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,449 @@
...
+
+  def _line_generator(self):
+    header_processed = False
+    if not header_processed and self._header_lines:
+      for header in self._header_lines:
+        self._current_line = header
+        yield self._current_line
+      header_processed = True
+    # Infinit loop; VcfParser.next() rasies StopIteration when EOF is reached.
+    while self._current_line:
```
arostamianfar: Thinking about this more: I think it's better to be explicit and just say `while True:` since we know that the string will never be set to None (see now below). The comment then becomes something like `Continue to process the next line indefinitely. The next line is set inside _get_record and this method is indirectly called in get_record`.
Also note that this behavior is independent of `VcfParser.next()`; the relationship is really between get_record and this method.

note: the current logic has a bug as it should check with `is not None`, otherwise, we'd terminate the pipeline if there's an empty line (PyVcf [skips empty lines](https://github.com/jamescasbon/PyVCF/blob/master/vcf/parser.py#L280)). I think we should add an integration test with an empty line to make sure this continues to work as we make changes. Could you please change, say, valid-4.2.vcf and add an empty line in the middle of the file?

gcp_variant_transforms/beam_io/vcfio.py
samanvp: done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -34,205 +29,30 @@
 from apache_beam.io.iobase import Read
 from apache_beam.transforms import PTransform
 
-__all__ = ['ReadFromVcf', 'ReadAllFromVcf', 'Variant', 'VariantCall',
-           'MalformedVcfRecord']
+from gcp_variant_transforms.beam_io import vcf_parser
 
+# All other modules depend on vcfio for the following const values.
```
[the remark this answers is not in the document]
samanvp: Filed https://github.com/googlegenomics/gcp-variant-transforms/issues/324

I also added another line to the comment. Please let me know if you think we should use some other reasons here.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,455 @@
...
+    # Continue to process the next line indefinitely. The next line is set
+    # inside _get_variant() and this method is indirectly called in get_variant.
+    while self._next_line_to_process:
+      # PyVCF has explicit str() calls when parsing INFO fields, which fails
+      # with UTF-8 decoded strings. Encode the line back to UTF-8.
+      self._current_line = self._next_line_to_process
+      self._next_line_to_process = None
+      yield self._current_line.encode('utf-8')
+      # Making sure _get_variant() assigned a new value before consuming it.
+      assert self._next_line_to_process
```
arostamianfar: Please add a comment on this assert:
`assert self._next_line_to_process, 'Internal error: A data line is requested to be processed more than once. Please file a bug if you see this!'`

gcp_variant_transforms/beam_io/vcfio.py
arostamianfar: Looks like this was added back in the latest sync.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,455 @@
...
+
+    text_source = textio._TextSource(
+        file_pattern,
+        0,  # min_bundle_size
+        compression_type,
+        True,  # strip_trailing_newlines
+        coders.StrUtf8Coder(),  # coder
+        validate=False,
+        header_processor_fns=(
+            lambda x: not x or not x.strip() or x.startswith('#'),
```
arostamianfar: Shouldn't this be `x.strip() or x.startswith('#')`? Otherwise, this would treat the entire file as a header. The unit tests would still pass given that PyVCF does't care about header vs. data lines, but integration tests would fail (likely run out of memory or take way too long for large samples).
Also, you dont need to worry about x being None here.

gcp_variant_transforms/beam_io/vcf_parser.py
arostamianfar: Also, I'd relax this assert to `self._next_line_to_process is not None` (I know that we skip empty lines, but even if we didn't, this means that something was processed).

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,455 @@
...
+GENOTYPE_FORMAT_KEY = 'GT'  # The genotype format key in a call.
+PHASESET_FORMAT_KEY = 'PS'  # The phaseset format key.
+DEFAULT_PHASESET_VALUE = '*'  # Default phaseset value if call is phased, but
+                              # no 'PS' is present.
+MISSING_GENOTYPE_VALUE = -1  # Genotype to use when '.' is used in GT field.
+
+
+class Variant(object):
+  """A class to store info about a genomic variant.
+  Each object corresponds to a single record in a VCF file.
```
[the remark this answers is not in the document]
samanvp: sorry, I don't know how this happened since I copied vcfio.py and started editing it!

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,455 @@
...
+                      self._file_name, data_line, str(e))
+      return MalformedVcfRecord(self._file_name, data_line, str(e))
+
+  def _line_generator(self):
+    for header in self._header_lines:
+      yield header
+    # Continue to process the next line indefinitely. The next line is set
+    # inside _get_variant() and this method is indirectly called in get_variant.
+    while self._next_line_to_process:
+      # PyVCF has explicit str() calls when parsing INFO fields, which fails
```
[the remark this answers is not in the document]
samanvp: done.

gcp_variant_transforms/beam_io/vcf_parser.py
samanvp: done.

gcp_variant_transforms/beam_io/vcf_parser.py
samanvp: I actually made the same mistake first and later realized it has to be `no x.strip()` otherwise all not empty lines would be considered as header since they all satisfy the `x.strip()` condition.

samanvp:
We looked at performance metrics offline together and it seems there is no significant increase.
Thanks,

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,460 @@
...
+    # inside _get_variant() and this method is indirectly called in get_variant.
+    while self._next_line_to_process:
+      self._current_line = self._next_line_to_process
+      self._next_line_to_process = None
+      # PyVCF has explicit str() calls when parsing INFO fields, which fails
+      # with UTF-8 decoded strings. Encode the line back to UTF-8.
+      yield self._current_line.encode('utf-8')
+      # Making sure _get_variant() assigned a new value before consuming it.
+      assert self._next_line_to_process is not None, (
+          'Internal error: A data line is requested to be processed more than' +
```
arostamianfar: nit: you don't need `+`. Also, please add an extra space after `than`

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -0,0 +1,460 @@
...
+               alternate_bases=None,  # type: List[str]
+               names=None,  # type: List[str]
+               quality=None,  # type: float
+               filters=None,  # type: List[str]
+               info=None,  # type: Dict[str, Any]
+               calls=None  # type: List[VariantCall]
+              ):
+    # type: (...) -> None
+    """Initialize the ``Variant`` object.
+    Args:
```
arostamianfar: Please also add a blank line after the first line for all methods.

gcp_variant_transforms/beam_io/vcf_parser.py
arostamianfar: ah, good call! you're right! ...and i'm glad to hear that our unit tests did catch the issue too :)

samanvp:
Both done, thanks again for your patience

Issue #239, opened 2018-05-16
Infer types from annotation fields


arostamianfar:
Currently we output all annotation fields as "string" as the type is not specified in the header. We can use similar ideas from header inference to infer the types of these fields as well.

PR #323, opened 2018-08-09
Infer annotation types

Jessime:
Adds inferred types to annotation columns in BQ.

Uses the first pass over the variants to create fake info headers in the representative header file, which are later read while creating the BQ schema.
    
Fixes: #239
Tested: unit tests


gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -310,12 +312,96 @@ def expand(self, pcoll):
...
+                             1,
+                             variant_merged_type,
+                             '',  # NO_DESCRIPTION
+                             '',  # UNKNOWN_SOURCE
+                             '')  # UNKNOWN_VERSION
+    inferred_header_infos = vcf_header_io.VcfHeader(infos=infos)
+    yield inferred_header_infos
+
+
+class InferAnnotationTypes(beam.PTransform):
```
bashir2: Please see my performance comment in `vcf_to_bq.py` and verify what I am saying is actually correct. To fix that multiple pass issue (i.e., going twice through all variants to infer headers), I think you can implement the logic you have in this `PTransform` inside `InferHeaderFields` instead. For example, one way that I can think of fixing this is by changing [`_infer_info_fields` method](https://github.com/googlegenomics/gcp-variant-transforms/blob/10aa5f8632f7b750fc2cf5ebcd434fb04566a29f/gcp_variant_transforms/transforms/infer_headers.py#L182) and add the logic for annotation header field generation in there. Then you can avoid the second pass and do everything in one `InferHeaderFields` pass.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -121,24 +121,44 @@ def _get_variant_merge_strategy(known_args  # type: argparse.Namespace
...
+    merged_header = _do_inferred_merge(inferred_headers,
+                                       merged_header,
+                                       known_args,
+                                       'Headers')
+  if known_args.infer_annotation_types:
+    inferred_headers = (filtered_variants | 'InferAnnotationTypes' >>
+                        infer_headers.InferAnnotationTypes(
+                            pvalue.AsSingleton(merged_header),
+                            known_args.annotation_fields))
+    merged_header = _do_inferred_merge(inferred_headers,
```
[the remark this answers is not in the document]
pgrosu: @bashir2 One is a subset of the other.  See the following:

https://github.com/googlegenomics/gcp-variant-transforms/blob/9690496f17c247cbd56de8915d608dde8e0fb6ef/gcp_variant_transforms/transforms/infer_headers.py#L326-L335

pgrosu:
@bashir2 What are the architectural/performance related issues?

bashir2:
@pgrosu re. your performance question, please see my comment in `vcf_to_bq.py` where [`merged_headers` is created](https://github.com/googlegenomics/gcp-variant-transforms/pull/323/commits/f9c103ed5fcb12e037d2cde7404356de7b7d053b#diff-7738b31bc91f7662490654314f8e6c6fR158).

pgrosu:
@Jessime Do you have any benchmark results?

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -242,6 +332,8 @@ def _infer_format_fields(self, variant, defined_headers):
       the field values.
     """
     formats = {}
+    if not self._infer_headers:
```
bashir2: The first time I read this I thought this is dropping all FORMAT fields from the generated header when `infer_headers` is false but `infer_annotation_types` is true. But looking at `vcf_to_bq` module I realized that we keep the originals. While I don't like comments inside code (the code should self describe itself) but this is one of those case that maybe it is a good idea to have a sentence clarifying why it is okay to drop FORMAT fields here.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -32,6 +32,9 @@
 from gcp_variant_transforms.beam_io import vcfio
 
 
+BASE_TYPE_KEY = '{}_{}_TYPE'
```
bashir2: Please add a reference to `annotation` in this name, e.g., `BASE_ANNOTATION_TYPE_KEY`, this should make the related changes easier to understand.

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -223,6 +248,71 @@ def _infer_info_fields(self, variant, defined_headers):
               str(defined_header.get(_HeaderKeyConstants.NUM)),
               str(info_field_value), str(variant))
           infos[info_field_key] = corrected_info
+
+  def _infer_type_info_fields(self, variant, infos, defined_headers):
```
bashir2: nit: `_infer_annotation_type_info_fields`

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -223,6 +248,71 @@ def _infer_info_fields(self, variant, defined_headers):
...
+    # type: (vcfio.Variant, Dict[str, Info], vcf_header_io.VcfHeader) -> None
+    """Updates `infos` with inferred annotation type info fields.
+
+    All annotation headers in each annotation field are converted to Info header
+    lines where the new ID corresponds to the given annotation field and header,
+    and the new TYPE corresponds to inferred type of the original header. Since
```
bashir2: nit: Why is `TYPE` all caps? If no reason, use `Type` to be consistent with VCF?

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -223,6 +248,71 @@ def _infer_info_fields(self, variant, defined_headers):
...
+    resolver = vcf_field_conflict_resolver.FieldConflictResolver(
+        resolve_always=True)
+    for field in self._annotation_fields:
+      if field not in variant.info:
+        continue
+      annotation_headers = annotation_parser.extract_annotation_names(
+          defined_headers.infos[field][
+              vcf_header_io.VcfParserHeaderKeyConstants.DESC], with_alt=True)
+      annotation_values = [annotation_parser.extract_annotation_list_with_alt(
+          annotation) for annotation in variant.info[field]]
```
bashir2: And of course if you decided to leave 'allele' out of type inference, then you need to drop it from here too.

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -223,6 +248,71 @@ def _infer_info_fields(self, variant, defined_headers):
...
+        `variant` that is not defined in the header or the definition mismatches
+        the field values.
+      defined_headers: header fields defined in header section of VCF files.
+    """
+    resolver = vcf_field_conflict_resolver.FieldConflictResolver(
+        resolve_always=True)
+    for field in self._annotation_fields:
+      if field not in variant.info:
+        continue
+      annotation_headers = annotation_parser.extract_annotation_names(
```
bashir2: nit: I think `annotation_names` is probably better here.

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -223,6 +248,71 @@ def _infer_info_fields(self, variant, defined_headers):
...
+          defined_headers.infos[field][
+              vcf_header_io.VcfParserHeaderKeyConstants.DESC], with_alt=True)
+      annotation_values = [annotation_parser.extract_annotation_list_with_alt(
+          annotation) for annotation in variant.info[field]]
+      annotation_values = zip(*annotation_values)
```
bashir2: Do we care to do a sanity check on list sizes here and report if the lengths are inconsistent?

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -223,6 +248,71 @@ def _infer_info_fields(self, variant, defined_headers):
...
+          if v != '':
+            variant_merged_type = resolver.resolve_attribute_conflict(
+                vcf_header_io.VcfParserHeaderKeyConstants.TYPE,
+                variant_merged_type,
+                self._get_field_type(v))
+        key_id = vcf_header_io.BASE_TYPE_KEY.format(field, header)
+        infos[key_id] = Info(key_id,
+                             1,
+                             variant_merged_type,
+                             '',  # NO_DESCRIPTION
```
bashir2: I think we can have a description because these will be saved in the merged header that we generate, correct? E.g., something like 'The inferred type field for annotation X'.

gcp_variant_transforms/transforms/infer_headers.py
bashir2: Please ignore this comment, I suppose you are referring to `VcfParserHeaderKeyConstants.TYPE` and that is fine.

gcp_variant_transforms/beam_io/vcf_header_io.py
arostamianfar: +1 to be explicit that this new logic is only for annotations; I even think that it's probably better to remove this constant from here, and instead move it inside infer_headers.py (as a private constant), and instead make a public `get_inferred_annotation_type_header_key` that returns the correct string. That way, all the annotation logic is isolated from the generic vcf_header_io, and there is a proper 'api' for getting this key.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -302,15 +306,48 @@ def create_alt_bases_field_schema(self):
...
     return alternate_bases_record
 
+  def _type_keys_from_names(self, annotation_names, annot_field):
+    #  type: (List[str], str) -> str
+    for annotation_name in annotation_names:
+      type_key = vcf_header_io.BASE_TYPE_KEY.format(annot_field,
+                                                    annotation_name)
+      yield type_key
+
+  def gen_all_info_type_keys(self):
```
arostamianfar: Curious: why are you using generators instead of just returning a list/set? I personally find simple list/set return types to be easier to work with especially if you end up creating a list/set from these anyway.

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -302,15 +306,48 @@ def create_alt_bases_field_schema(self):
...
             description=annotation_descs.get(annotation_name, '')))
       alternate_bases_record.fields.append(annotation_record)
     return alternate_bases_record
 
+  def _type_keys_from_names(self, annotation_names, annot_field):
+    #  type: (List[str], str) -> str
+    for annotation_name in annotation_names:
+      type_key = vcf_header_io.BASE_TYPE_KEY.format(annot_field,
+                                                    annotation_name)
+      yield type_key
```
arostamianfar: Consider returning a tuple of `annotation_name, type_key`; that way, the type and name are associated together and you can avoid the `zip` calls.

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -39,12 +41,37 @@
...
     values.
+  if `infer_annotation_types` is True:
+  - Fields containing type information of corresponding annotation Info fields.
   """
 
+  def __init__(
+      self,
+      annotation_fields=None,  # type: Union[None, List[str]]
+      infer_headers=False,  # type: bool
+      infer_annotation_types=False  # type: bool
```
arostamianfar: I think it's better to just use one variable here (e.g. `annotation_fields_to_infer` or a better name :p) that just contains the list of annotation fields. That way, the API is simpler and you dont need to worry about cases where flag is true, but annotation_fields is empty or vice versa.

gcp_variant_transforms/libs/processed_variant.py
arostamianfar: Great point! Thanks for the clarification!

gcp_variant_transforms/libs/bigquery_vcf_schema.py
```diff
@@ -135,10 +135,13 @@ def generate_schema_from_header_fields(
 
   # Add info fields.
   info_keys = set()
+  info_type_keys_set = set(key for name, key in
+                           proc_variant_factory.gen_annotation_info_type_keys())
   for key, field in header_fields.infos.iteritems():
     # END info is already included by modifying the end_position.
```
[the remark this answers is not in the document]
Jessime: Done.

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -39,12 +42,33 @@
...
 
+  def __init__(
+      self,
+      annotation_fields_to_infer=None,  # type: Optional[List[str]]
+      infer_headers=False,  # type: bool
+      ):
+    """
+    Args:
+      annotation_fields_to_infer: list of info fields treated as annotation
+        fields (e.g. ['CSQ_VT'])
```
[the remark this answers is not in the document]
Jessime: I've added 'CSQ' to match other comments and demonstrate that multiple fields are fine.

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -223,6 +245,81 @@ def _infer_info_fields(self, variant, defined_headers):
...
+        continue
+      annotation_names = annotation_parser.extract_annotation_names(
+          defined_headers.infos[field][
+              vcf_header_io.VcfParserHeaderKeyConstants.DESC])
+      annotation_values = [annotation_parser.extract_annotation_list_with_alt(
+          annotation)[1:] for annotation in variant.info[field]]
+      self._check_annotation_lists_lengths(annotation_names, annotation_values)
+      annotation_values = zip(*annotation_values)
+      for name, values in zip(annotation_names, annotation_values):
+        variant_merged_type = vcf_header_io.VcfHeaderFieldTypeConstants.INTEGER
```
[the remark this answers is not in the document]
pgrosu: Guys you are complicating this too much.  Just use [`set`](https://www.geeksforgeeks.org/sets-in-python/) from Python, and resolve it like Asha mentioned at the end when creating the `Info` object.  Adding to a set is a unique, and if you want you can check by intersecting with `Integer` so you don't downcast to `String`.  You can even start with a `None` element in the initial set, if that is preferred but is not necessary.  Here's a quick tutorial on how to use sets in Python: [https://www.geeksforgeeks.org/sets-in-python/](https://www.geeksforgeeks.org/sets-in-python/)

Hope it helps,
~p

gcp_variant_transforms/libs/processed_variant_test.py
```diff
@@ -488,3 +489,52 @@ def test_create_alt_bases_field_schema_descriptions(self):
...
     a2_fields = [field for field in schema.fields if field.name == 'A2']
     self.assertEqual(len(a2_fields), 1)
+
+  def test_create_alt_bases_field_schema_types(self):
+    _, header_fields = self._get_sample_variant_and_header_with_csq()
+    for hfi in header_fields.infos.values():
+      hfi['type'] = 'String'
+    header_fields.infos.update({
+        'CSQ_Allele_TYPE':
+        OrderedDict([('id', 'CSQ_Allele_TYPE'),
```
arostamianfar: Please use `vcf.parser._Info` for these. Some tests explicitly import `_Info as Info` to make it less awkward.
We'll replace this with VcfHeader class from nucleus eventually, so it would be nicer, so using the Info class makes it easier to find and replace all instances of PyVCF headers.

gcp_variant_transforms/transforms/infer_headers_test.py
```diff
@@ -53,6 +53,17 @@ def _get_sample_header_fields(self):
         ('PS', Format('PS', 1, 'Integer', 'Special PS key'))])
     return vcf_header_io.VcfHeader(infos=infos, formats=formats)
 
+  def _get_sample_header_fields_with_annotation(self):
+    header = self._get_sample_header_fields()
+    header.infos['CSQ'] = OrderedDict([
```
arostamianfar: Please use `Info` and below here as well.

gcp_variant_transforms/transforms/infer_headers_test.py
```diff
@@ -401,3 +417,141 @@ def test_pipeline(self):
       assert_that(inferred_headers,
                   asserts.header_fields_equal_ignore_order([expected]))
       p.run()
+
+  def test_infer_annotation_types_no_conflicts(self):
+    anno_fields = ['CSQ']
+    header = self._get_sample_header_fields_with_annotation()
+    variant = self._get_sample_variant_1()
+    variant.info['CSQ'] = ['A|GENE1|100|1.2', 'TT|GENE1|101|1.3']
+    infer_header_fields = infer_headers._InferHeaderFields(False, anno_fields)
```
arostamianfar: nit: please use `_InferHeaderFields(infer_headers=False, annotation_fields_to_infer=anno_fields)` so that it's more clear what the `False` stands for.

gcp_variant_transforms/transforms/infer_headers_test.py
```diff
@@ -401,3 +417,141 @@ def test_pipeline(self):
...
+    new_header = next(infer_header_fields.process(variant, header))
+    desc = 'Inferred type field for annotation {}.'
+    expected = vcf_header_io.VcfHeader(infos={
+        'CSQ_Gene_TYPE':
```
arostamianfar: I think some of the checks here are unnecessary (e.g. if we change the description format, we have to change all of these tests too; can be saved by reusing a constant, but I think it's not needed).
What do you think of the following approach:

Create a dict like:
```
expected_types = {
    'CSQ_Gene_TYPE': 'String',
    'CSQ_Position_TYPE': 'Integer', ...
}
```
and then just loop through the inferred_headers and check that the types match. You can also check the length of the two dicts to make sure they match.

gcp_variant_transforms/transforms/infer_headers_test.py
```diff
@@ -401,3 +417,141 @@ def test_pipeline(self):
...
                   asserts.header_fields_equal_ignore_order([expected]))
       p.run()
+
+  def test_infer_annotation_types_no_conflicts(self):
+    anno_fields = ['CSQ']
+    header = self._get_sample_header_fields_with_annotation()
+    variant = self._get_sample_variant_1()
+    variant.info['CSQ'] = ['A|GENE1|100|1.2', 'TT|GENE1|101|1.3']
+    infer_header_fields = infer_headers._InferHeaderFields(False, anno_fields)
+    new_header = next(infer_header_fields.process(variant, header))
```
arostamianfar: s/new_header/inferred_headers

gcp_variant_transforms/transforms/infer_headers_test.py
```diff
@@ -401,3 +417,141 @@ def test_pipeline(self):
...
+    expected = vcf_header_io.VcfHeader(infos={
+        'CSQ_Gene_TYPE':
+        Info('CSQ_Gene_TYPE', 1, 'String', desc.format('Gene'), '', ''),
+        'CSQ_Position_TYPE':
+        Info('CSQ_Position_TYPE', 1, 'Integer',
+             desc.format('Position'), '', ''),
+        'CSQ_Score_TYPE':
+        Info('CSQ_Score_TYPE', 1, 'Float', desc.format('Score'), '', '')})
+    self.assertEqual(expected, new_header)
+    return new_header, expected
```
arostamianfar: nit: remove

gcp_variant_transforms/transforms/infer_headers_test.py
```diff
@@ -401,3 +417,141 @@ def test_pipeline(self):
...
+
+  def test_infer_annotation_types_with_conflicts(self):
```
arostamianfar: nit: s/_conflicts/_type_conflicts (it may not be obvious that you're referring to type conflicts (it could be name or other things)).

gcp_variant_transforms/transforms/infer_headers.py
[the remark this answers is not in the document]
arostamianfar: @pgrosu : thanks for the suggestion, but the challenge here is that we need to merge these in a distributed environment, so we cannot use a `set` directly. However, it's an interesting suggestion to make the type string "look like" a set and then merge those together, but I think we'd end up storing more info than needed. We have some other ideas about fixing this, but we'll look into it after this PR is merged.

pgrosu:
@arostamianfar With distributed it is still a set, since your key can be `VariantMergedType` and the values are the types that you get from each worker.  It is still a set in the end if you combine it using `CombinePerKey` and transform it get the unique ones.  The intersection operation would still apply.  I wouldn't overthink it.

gcp_variant_transforms/transforms/infer_headers_test.py
Jessime: Done. As discussed offline, I've modified `_get_sample_header_fields` so I can use `Info` here.

gcp_variant_transforms/libs/processed_variant_test.py
```diff
@@ -108,14 +110,29 @@ def test_create_processed_variant_move_alt_info(self):
...
+    """
+    # type:  (
     variant = self._get_sample_variant()
     variant.info['CSQ'] = ['A|C1|I1|S1|G1', 'TT|C2|I2|S2|G2', 'A|C3|I3|S3|G3']
-    header_fields = vcf_header_util.make_header({'CSQ': '.',
-                                                 'A1': '1', 'A2': 'A'})
-    header_fields.infos['CSQ'][
-        vcf_header_io.VcfParserHeaderKeyConstants.DESC] = (
-            'some desc Allele|Consequence|IMPACT|SYMBOL|Gene')
+    infos = OrderedDict([
```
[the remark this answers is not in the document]
pgrosu: FYI, since you guys are Python 2.7 I would stay away from OrderedDict, as noted here:  

https://www.georgevreilly.com/blog/2017/02/21/OrderedDictInitialization.html

pgrosu:
@Jessime You're doing too much hard-coding in your https://github.com/googlegenomics/gcp-variant-transforms/blob/118b8f6dab5e57b8d178f32ea4e8e97c1ae8fcf4/gcp_variant_transforms/transforms/infer_headers_test.py file.  Consider having a smaller input test-file that will be loaded up through some grammar and processed accordingly.  Based on your current method of testing eventually you'll discover that your current file will become gigantic and hard to manage.

gcp_variant_transforms/transforms/infer_headers.py
```diff
@@ -179,23 +219,21 @@ def _infer_mismatched_format_field(self,
...
 
-  def _infer_info_fields(self, variant, defined_headers):
-    # type: (vcfio.Variant, vcf_header_io.VcfHeader) -> Dict[str, Info]
-    """Returns inferred info fields.
+  def _infer_standard_info_fields(self, variant, infos, defined_headers):
+    # type: (vcfio.Variant, Dict[str, Info], vcf_header_io.VcfHeader) -> None
+    """Updates `infos` with inferred info fields.
 
-    Two types of info fields are inferred:
+    Three types of info fields are inferred:
```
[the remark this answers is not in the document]
Jessime: Done.

PR #327, opened 2018-08-13
Remove chrM from the default partition config.

arostamianfar:
Rationale:
* Mitochondrial DNA is missing from the "primary assemblies": [hg19](https://www.ncbi.nlm.nih.gov/assembly/GCF_000001405.13/#/def_asm_Primary_Assembly), [hg38](https://www.ncbi.nlm.nih.gov/assembly/GCF_000001405.38) 
* Looks like `chrMT` is more commonly used than `chrM`. I originally thought about including `chrMT` here as well, but then I figured that it's a small enough chromosome, so it's ok to be included in residual and not worry about the semantics on our end.
* Users can easily add it back if they need to.


arostamianfar:
Good point about integration tests. I think it would be lower priority as we dont expect to change this config again.

PR #330, opened 2018-08-15
compose files in multiple processes #85

allieychen:
In BQ to VCF pipeline, after generating all VCF shards, the composing process could take hours since each time only 32 objects can be composed. 

This PR composes the files in parallel to speed up the process.

Tested: Manually run BQ to VCF. For a table with about 9000 records, it took 3 to 4 minutes to compose the temp files.

### DIFF
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -39,21 +39,18 @@ from typing import Iterable, List, Tuple  # pylint: disable=unused-import
 import apache_beam as beam
 from apache_beam.io import filesystems
 from apache_beam.io.gcp import bigquery
-from apache_beam.io.gcp import gcsio
 from apache_beam.options import pipeline_options
-
-from google.cloud import storage
 
 from gcp_variant_transforms import vcf_to_bq_common
 from gcp_variant_transforms.beam_io import vcfio
 from gcp_variant_transforms.libs import bigquery_util
+from gcp_variant_transforms.libs import bq_to_vcf_temp_files_composer
 from gcp_variant_transforms.options import variant_transform_options
 from gcp_variant_transforms.transforms import bigquery_to_variant
 from gcp_variant_transforms.transforms import densify_variants
 
 _BASE_QUERY_TEMPLATE = 'SELECT * FROM `{INPUT_TABLE}`;'
 _COMMAND_LINE_OPTIONS = [variant_transform_options.BigQueryToVcfOptions]
-_MAX_NUM_OF_BLOBS_PER_COMPOSE = 32
 
 
 def run(argv=None):
@@ -74,9 +71,9 @@ def run(argv=None):
       google_cloud_options.temp_location, shards_folder)
 
   _bigquery_to_vcf_shards(known_args, options, bq_to_vcf_temp_folder)
-  _compose_temp_files(google_cloud_options.project,
-                      known_args.output_file,
-                      bq_to_vcf_temp_folder)
+  bq_to_vcf_temp_files_composer.compose_temp_files(google_cloud_options.project,
+                                                   known_args.output_file,
+                                                   bq_to_vcf_temp_folder)
 
 
 def _bigquery_to_vcf_shards(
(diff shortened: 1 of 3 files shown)


gcp_variant_transforms/libs/bq_to_vcf_temp_files_composer.py
```diff
@@ -0,0 +1,144 @@
...
+    blobs_to_be_composed = list(self._bucket.list_blobs(prefix=blob_prefix))
+    if len(blobs_to_be_composed) == 1:
+      return blobs_to_be_composed[0]
+    new_blob_prefix = filesystems.FileSystems.join(blob_prefix, 'composed_')
+
+    proc_pool = multiprocessing.Pool(processes=self._NUM_PROCS)
+    for blob_names in self._break_list_in_chunks(blobs_to_be_composed,
+                                                 _MAX_NUM_OF_BLOBS_PER_COMPOSE):
+      _, file_name = filesystems.FileSystems.split(blob_names[0])
+      new_blob_name = ''.join([new_blob_prefix + file_name])
```
arostamianfar: Did you mean to use a comma instead of a plus? (I think they're both plain strings?)

gcp_variant_transforms/libs/bq_to_vcf_temp_files_composer.py
```diff
@@ -0,0 +1,144 @@
...
+  bucket_name, blob_name = gcsio.parse_gcs_path(file_path)
+  file_blob = client.get_bucket(bucket_name).blob(blob_name)
+  file_blob.content_type = 'text/plain'
+  return file_blob
+
+
+class MultiProcessComposer(object):
+  """Class to compose (a large number of) files in GCS in parallel."""
+
+  _NUM_PROCS = 63
```
arostamianfar: Curious why you chose this number? Is it related to a GCS limit?

gcp_variant_transforms/libs/bq_to_vcf_temp_files_composer.py
```diff
@@ -0,0 +1,144 @@
...
+    self._bucket_name = bucket_name
+    self._blob_prefix = blob_prefix
+    self._bucket = storage.Client(project).get_bucket(bucket_name)
+
+  def run(self):
+    # type: () -> storage.Blob
+    """Returns the final blob that all blobs composed to."""
+    return self._compose_vcf_shards_to_one(self._blob_prefix)
+
+  def _compose_vcf_shards_to_one(self, blob_prefix):
```
arostamianfar: nit: since this logic is not VCF-specific, consider removing VCF from the name and comments. I think `_compose_blobs_to_one` may be a better name?

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -39,21 +39,18 @@
...
 from apache_beam.io.gcp import bigquery
-from apache_beam.io.gcp import gcsio
 from apache_beam.options import pipeline_options
 
-from google.cloud import storage
-
 from gcp_variant_transforms import vcf_to_bq_common
 from gcp_variant_transforms.beam_io import vcfio
 from gcp_variant_transforms.libs import bigquery_util
+from gcp_variant_transforms.libs import bq_to_vcf_temp_files_composer
```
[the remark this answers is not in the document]
allieychen: Done.

gcp_variant_transforms/libs/bq_to_vcf_temp_files_composer.py
allieychen: I have done some experiments with this number, it seems that it is not necessary to specify this value.

Quote from [composite doc](https://cloud.google.com/storage/docs/composite-objects):
`There is a per-project rate limit on how many source objects you can compose of approximately 200 source objects per second.`

So I think roughly 6 processes can be created to run in parallel.

To compose 595 blobs, without this number, it takes roughly took about 20s. With `_NUM_PROCS = 6`, `_NUM_PROCS = 7`, `_NUM_PROCS = 16`, it took about the same time (25s). With `_NUM_PROCS = 1`, it took about 95 s. So I think it might be best to let the system decide the number of processes.

I will try to run with a bigger data sets to make sure it won't break.

gcp_variant_transforms/libs/vcf_file_composer.py
```diff
@@ -0,0 +1,145 @@
...
+    bucket_name: The name of the bucket where the `components` and the new
+      composite are saved.
+    blob_names: A list of blob object names.
+    composite_name: Name of the new composite.
+  """
+  bucket = storage.Client(project).get_bucket(bucket_name)
+  output_file_blob = bucket.blob(composite_name)
+  output_file_blob.content_type = 'text/plain'
+  blobs = [bucket.get_blob(blob_name) for blob_name in blob_names]
+  output_file_blob.compose(blobs)
```
pgrosu: @allieychen This is nice, but where are you doing any integrity checks?

pgrosu: No problem Allie, that's fine - it was only to ensure the tests have might not be sprinkled with false positives.  One other thing to keep in mind is some sort of retry policy - funny things with high-throughput data can sneak up on you unexpectedly even when chunking at the max of 32 :)

arostamianfar:
Awesome! ....and very cool that we can actually output a 6TB VCF file :D I think that would be the upper threshold of having a usable VCF file, so it looks like the scalability problem for BQ->VCF is solved too! Great work!

allieychen:
@arostamianfar My mistake. To clarify, 6.19 TB is the disk space when running the dataflow. The actually VCF file is 1.84 TB. :). We definitely can try with a larger public dataset later.

pgrosu:
Great work, now try to push that limit higher.  For example the Broad produces 20 TB of sequencing data per day, and you want to make your product dynamically-scalable as new data comes in.  Don't get me wrong, this is still a great start!

Great job!
~p

PR #337, opened 2018-08-21
add data header in BQ to VCF pipeline #85

allieychen:
- Add the VCF data header in the BQ to VCF pipeline.
- The call names in the header will be sorted.

Tested: unit tests & Manually run BQ to VCF.

### DIFF
--- a/gcp_variant_transforms/transforms/combine_call_names.py
+++ b/gcp_variant_transforms/transforms/combine_call_names.py
@@ -14,7 +14,7 @@
 
 """A PTransform to combine call names from all variants."""
 
-from typing import List  # pylint: disable=unused-import
+from typing import List, Tuple  # pylint: disable=unused-import
 
 import apache_beam as beam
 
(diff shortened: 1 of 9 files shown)


gcp_variant_transforms/transforms/densify_variants_test.py
```diff
@@ -30,13 +30,25 @@
...
 
-  def test_add_missing_calls(self):
-    transform = densify_variants.DensifyVariants()
-    variant = vcfio.Variant(calls=[vcfio.VariantCall(name='sample2')])
-    new_variant = transform._densify_variants(
-        variant, ['sample1', 'sample2', 'sample3'])
-    call_names = [call.name for call in new_variant.calls]
-    self.assertItemsEqual(call_names, ['sample1', 'sample2', 'sample3'])
+  def test_densify_variants_pipeline_no_calls(self):
+    call_names = ['sample1', 'sample2', 'sample3']
```
arostamianfar: nit: since call_names is now only used once (and are intentionally ignored), consider just hardcoding it inside `VariantCall(name=...)`. You can add a comment about why the calls are ignored too.

gcp_variant_transforms/transforms/combine_call_names.py
```diff
@@ -0,0 +1,59 @@
...
+  """A PTransform to combine call names from all variants."""
+
+  def _get_call_names(self, variant):
+    # type: (vcf_parser.Variant) -> Tuple[str]
+    """Returns the names of all calls for the variant."""
+    call_names = [call.name for call in variant.calls]
+    if len(call_names) != len(set(call_names)):
+      raise ValueError('There are duplicate call names in the variant {}'.
+                       format(variant))
+    return tuple(call_names)
```
arostamianfar: curious: why change this to a tuple instead of just a list? is it to emphasize immutability?

PR #339, opened 2018-08-22
schema to header

allieychen:
In BQ to VCF, if the representative header is not present, we will infer the headers on a best effort basis. This PR 
- add one module to generate the header fields from Big Query schema.
- unit tests are included.

Tested: unit tests.


gcp_variant_transforms/libs/bigquery_schema_to_vcf_header_test.py
```diff
@@ -0,0 +1,141 @@
...
+
+from gcp_variant_transforms.beam_io import vcf_header_io
+from gcp_variant_transforms.libs import bigquery_schema_to_vcf_header
+from gcp_variant_transforms.libs import bigquery_util
+
+
+class VcfHeaderSourceTest(unittest.TestCase):
+  """Test cases for the `generate_header_fields_from_schema` function."""
+
+  def _get_table_schema(self):
```
pgrosu: You might want to have this in a separate file than hard-coding it.  If the schema changes it'll become a headache with many changes or any transformations during migration.

allieychen: Thanks, Paul! It is a very good point, and I actually need this sample schema for testing in some other modules as well. Done!

pgrosu: No problemo Allie, I was fairly sure you were aware of the abstraction :)

gcp_variant_transforms/libs/bigquery_schema_to_vcf_header.py
```diff
@@ -0,0 +1,101 @@
...
+  for field in schema.fields:
+    if field.name not in _NON_RECORD_CONSTANT_FIELDS:
+      if field.name == bigquery_util.ColumnKeyConstants.CALLS:
+        _add_format_fields(field, formats)
+      elif field.name == bigquery_util.ColumnKeyConstants.ALTERNATE_BASES:
+        _add_info_fields_from_alternate_bases(field, infos)
+      else:
+        infos.update({
+            field.name: _Info(field.name,
+                              vcfio.MISSING_FIELD_VALUE,
```
arostamianfar: Sadly, this is actually an integer (I know it's very confusing, but something that PyVCF decided to do). Please use `vcf.parser.field_counts[vcfio.MISSING_FIELD_VALUE]` (which converts it to None). Longer term, we should update to use Nucleus' protos, where it is a string and much more convenient.

gcp_variant_transforms/libs/bigquery_schema_to_vcf_header.py
```diff
@@ -0,0 +1,101 @@
...
+_BIG_QUERY_TYPE_TO_VCF_TYPE_MAP = bigquery_util.BIG_QUERY_TYPE_TO_VCF_TYPE_MAP
+_Format = parser._Format
+_Info = parser._Info
+
+
+def generate_header_fields_from_schema(schema):
+  # type: (bigquery.TableSchema) -> vcf_header_io.VcfHeader
+  """Returns header fields converted from BigQuery schema.
+
+  This is done on a best effort basis. Only INFO and FORMAT are considered. For
```
arostamianfar: s/This is done on a best effort basis./This is a best effort reconstruction of header fields.

gcp_variant_transforms/libs/bigquery_schema_to_vcf_header.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2018 Google Inc.  All Rights Reserved.
```
arostamianfar: Any reason for not using the existing `bigquery_vcf_schema.py` file? It seems natural to have both `generate_header_fields_from_schema` and `generate_schema_from_header_fields` in one file.
I think we should also have a unit test with header->schema->header to make sure all the assumptions are met and no fields are dropped/added (you can choose the original header to be compatible with 'best effort' reconstruction so that you can exactly match on the reconstructed header). This will also catch things like integer vs. string in the Number field.

allieychen: Very nice advice! Updated. The combined test of header->schema->header is awesome!

gcp_variant_transforms/testing/bigquery_schema_util.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+from apache_beam.io.gcp.internal.clients import bigquery
+
+from gcp_variant_transforms.libs import bigquery_util
+
+
+def get_sample_table_schema():
+  # type: () -> bigquery.TableSchema
+  """Creates a sample BigQuery table schema."""
```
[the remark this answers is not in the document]
allieychen: Done.

allieychen:
@arostamianfar Thanks for the review. The name is updated. PTAL.

PR #346, opened 2018-08-24
Adding the second VcfParser class: NucleusParser 

samanvp:
After defining the VCFParser interface and PyVcfParser in a previous PR, here we add
NucleusParser. Following points must be highlighted:
 * This new parser is hidden behind a flag (use_nucleus=False).
 * Since we are not still able to create a Wheel file that have the
 from_string() method, we use a hack to make Nucleus work: write the whole
 file to a local temp file and send it as input to parser.
 * Currently we don't have any unit tests or integration test. We
 manually cheked the sanity of Nucleus output by comparing its output
 table with PyVcf output. This must be done in the next step.

### DIFF
--- a/gcp_variant_transforms/beam_io/vcf_parser.py
+++ b/gcp_variant_transforms/beam_io/vcf_parser.py
@@ -29,7 +29,7 @@ import nucleus
 import vcf
 
 from apache_beam.coders import coders
-from apache_beam.io.filesystems import FileSystems
+from apache_beam.io import filesystems
 from apache_beam.io import textio
 
 
@@ -48,7 +48,7 @@ PHASESET_FORMAT_KEY = 'PS'  # The phaseset format key.
 DEFAULT_PHASESET_VALUE = '*'  # Default phaseset value if call is phased, but
                               # no 'PS' is present.
 MISSING_GENOTYPE_VALUE = -1  # Genotype to use when '.' is used in GT field.
-
+FIRST_HEADER_LINE = '##fileformat=VCFv4.2'
 
 class Variant(object):
   """A class to store info about a genomic variant.
@@ -496,14 +496,14 @@ class NucleusParser(VcfParser):
     # This member will be properly initiated in _init_with_header().
     self._vcf_reader = None
     # These members will be properly initiated in _extract_header_fields().
-    self._is_info_repeated = {}
-    self._is_format_repeated = {}
+    self._header_infos = {}
+    self._header_formats = {}
     # This is a temporary solution until from_string will be fixed.
     self._temp_local_file = self._write_to_local_file(file_name)
 
   def _write_to_local_file(self, remote_file_name):
-    (temp_file, temp_file_name) = tempfile.mkstemp(text=True)
-    with FileSystems.open(remote_file_name) as f:
+    temp_file, temp_file_name = tempfile.mkstemp(text=True)
+    with filesystems.FileSystems.open(remote_file_name) as f:
       while True:
         line = f.readline()
         if line:
@@ -513,22 +513,15 @@ class NucleusParser(VcfParser):
           break
     return temp_file_name
 
-  def _store_to_temp_local_file(self, header_lines):
-    (temp_file, temp_file_name) = tempfile.mkstemp(text=True)
-    for line in header_lines:
-      os.write(temp_file, line)
-    os.close(temp_file)
-    return temp_file_name
-
   def _init_with_header(self, header_lines):
-    # This header line is needed by Nucleus.
-    header_lines = ['##fileformat=VCFv4.2'] + header_lines
+    # The first header line must be similar to '##fileformat=VCFv4.2'.
+    if header_lines and not header_lines[0].startswith('##fileformat'):
+      header_lines = [FIRST_HEADER_LINE] + header_lines
     try:
       # This is a temporary solution until from_string will be fixed.
-      self._vcf_reader = nucleus.io.vcf.VcfReader(
-          self._temp_local_file, use_index=False)
-
-    except SyntaxError as e:
+      self._vcf_reader = nucleus.io.vcf.VcfReader(self._temp_local_file,
+                                                  use_index=False)
+    except ValueError as e:
       raise ValueError(
           'Invalid VCF header in %s: %s' % (self._file_name, str(e)))
     self._extract_header_fields()
@@ -536,50 +529,62 @@ class NucleusParser(VcfParser):
   def _extract_header_fields(self):
     header = self._vcf_reader.header
     for info in header.infos:
-      if info.number in ('0', '1'):
-        self._is_info_repeated[info.id] = False
-      else:
-        self._is_info_repeated[info.id] = True
+      self._header_infos[info.id] = info
 
     for format_info in header.formats:
-      if format_info.number in ('0', '1'):
-        self._is_format_repeated[format_info.id] = False
-      else:
-        self._is_format_repeated[format_info.id] = True
+      self._header_formats[format_info.id] = format_info
+
+  def _is_info_repeated(self, info_id):
+    info = self._header_infos.get(info_id, None)
+    if not info or not info.number:
+      return False
+    else:
+      return self._is_repeated(info.number)
+
+  def _is_format_repeated(self, format_id):
+    format_info = self._header_formats.get(format_id, None)
+    if not format_info or not format_info.number:
+      return False
+    else:
+      return self._is_repeated(format_info.number)
+
+  def _is_repeated(self, number):
+    if number in ('0', '1'):
+      return False
+    else:
+      return True
 
   def _get_variant(self, data_line):
     try:
       # This is a temporary solution until from_string will be fixed.
-      record = next(self._vcf_reader)
-      return self._convert_to_variant(record)
+      variant_proto = next(self._vcf_reader)
+      return self._convert_to_variant(variant_proto)
     except (LookupError, ValueError) as e:
-      logging.warning('VCF record read failed in %s for line %s: %s',
+      logging.warning('VCF variant_proto read failed in %s for line %s: %s',
                       self._file_name, data_line, str(e))
       return MalformedVcfRecord(self._file_name, data_line, str(e))
 
-  def _convert_to_variant(self, record):
-    # type: (nucleus_proto.Variant) -> Variant
+  def _convert_to_variant(self, variant_proto):
+    # type: (nucleus.protos.variants_pb2.Variant) -> Variant
     return Variant(
-        reference_name=record.reference_name,
-        start=record.start,
-        end=record.end,
-        reference_bases=(record.reference_bases
-                         if record.reference_bases != MISSING_FIELD_VALUE
+        reference_name=variant_proto.reference_name,
+        start=variant_proto.start,
+        end=variant_proto.end,
+        reference_bases=(variant_proto.reference_bases
+                         if variant_proto.reference_bases != MISSING_FIELD_VALUE
                          else None),
-        alternate_bases=(
-            map(str, record.alternate_bases) if record.alternate_bases else []),
-        names=map(str, record.names) if record.names else [],
-        quality=record.quality,
-        filters=(
-            [PASS_FILTER] if record.filter == [] else map(str, record.filter)),
-        info=self._get_variant_info(record),
-        calls=self._get_variant_calls(record))
+        alternate_bases=variant_proto.alternate_bases,
+        names=map(str, variant_proto.names),
+        quality=variant_proto.quality,
+        filters=map(str, variant_proto.filter),
+        info=self._get_variant_info(variant_proto),
+        calls=self._get_variant_calls(variant_proto))
 
-  def _get_variant_info(self, record):
+  def _get_variant_info(self, variant_proto):
     info = {}
-    for k in record.info:
-      data = self._convert_list_value(record.info[k],
-                                      self._is_info_repeated.get(k, False))
+    for k in variant_proto.info:
+      data = self._convert_list_value(variant_proto.info[k],
+                                      self._is_info_repeated(k))
       # Prevents including missing flags as 'false' flag.
       if isinstance(data, bool) and not data:
         continue
@@ -612,38 +617,39 @@ class NucleusParser(VcfParser):
 
     if is_repeated:
       return output_list
-    if not output_list:
-      return None
-    if len(output_list) == 1:
-      return output_list[0]
-    raise ValueError('a not repeated field has more than 1 value')
+    else:
+      if len(output_list) > 1:
+        raise ValueError('a not repeated field has more than 1 value')
+      if len(output_list) == 1:
+        return output_list[0]
+      else:
+        return None
 
-  def _get_variant_calls(self, record):
+  def _get_variant_calls(self, variant_proto):
     calls = []
-    for sample in record.calls:
+    for call_proto in variant_proto.calls:
       call = VariantCall()
-      call.name = sample.call_set_name
-      if not sample.genotype:
+      call.name = call_proto.call_set_name
+      if not call_proto.genotype:
         call.genotype.append(MISSING_GENOTYPE_VALUE)
       else:
-        for v in sample.genotype:
-          call.genotype.append(v)
+        call.genotype = list(call_proto.genotype)
 
       phaseset_from_format = (
-          sample.info[PHASESET_FORMAT_KEY].values[0].string_value
-          if sample.info.get(PHASESET_FORMAT_KEY, None)
+          self._convert_list_value(call_proto.info[PHASESET_FORMAT_KEY], False)
+          if call_proto.info.get(PHASESET_FORMAT_KEY, None)
           else None)
       # Note: Call is considered phased if it contains the 'PS' key regardless
       # of whether it uses '|'.
-      if phaseset_from_format or sample.is_phased:
+      if phaseset_from_format or call_proto.is_phased:
         call.phaseset = (phaseset_from_format if phaseset_from_format
                          else DEFAULT_PHASESET_VALUE)
-      for k in sample.info:
+      for k in call_proto.info:
         # Genotype and phaseset (if present) are already included.
         if k in (GENOTYPE_FORMAT_KEY, PHASESET_FORMAT_KEY):
           continue
-        is_repeated = self._is_format_repeated.get(k, False)
-        data = self._convert_list_value(sample.info[k], is_repeated)
+        data = self._convert_list_value(call_proto.info[k],
+                                        self._is_format_repeated(k))
         call.info[k] = data
       calls.append(call)
     return calls
--- a/gcp_variant_transforms/beam_io/vcf_parser.py
+++ b/gcp_variant_transforms/beam_io/vcf_parser.py
@@ -25,8 +25,8 @@ from collections import namedtuple
 import os
 import tempfile
 
-from nucleus.io.python import vcf_reader as nucleus
-from nucleus.protos import variants_pb2 as nucleus_proto
+from nucleus.io.python import vcf_reader as nucleus_vcf_reader
+from nucleus.protos import variants_pb2
 import vcf
 
 from apache_beam.coders import coders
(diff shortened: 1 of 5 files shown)


gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+      while True:
+        line = f.readline()
+        if line:
+          os.write(temp_file, line)
+        else:
+          os.close(temp_file)
+          break
+    return temp_file_name
+
+  def _store_to_temp_local_file(self, header_lines):
```
arostamianfar: this is not used?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+                                        range_tracker,
+                                        file_pattern,
+                                        compression_type,
+                                        allow_malformed_records,
+                                        representative_header_lines,
+                                        **kwargs)
+    # This member will be properly initiated in _init_with_header().
+    self._vcf_reader = None
+    # These members will be properly initiated in _extract_header_fields().
+    self._is_info_repeated = {}
```
arostamianfar: Please create a dict of info and format that point to `VcfInfo` and `VcfFormatInfo` instead; that is more future-proof (e.g. if we want to also check the description or type for another purpose, we could just reuse those dicts later on without introducing more specialized dicts). You can keep the is_repeated arg in `_convert_list_value` as it's a nice abstraction, but you can make a helper function that given a number string would return true/false if the field is repeated.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+  def _store_to_temp_local_file(self, header_lines):
+    (temp_file, temp_file_name) = tempfile.mkstemp(text=True)
+    for line in header_lines:
+      os.write(temp_file, line)
+    os.close(temp_file)
+    return temp_file_name
+
+  def _init_with_header(self, header_lines):
+    # This header line is needed by Nucleus.
+    header_lines = ['##fileformat=VCFv4.2'] + header_lines
```
arostamianfar: Please only add this if the first like is not `##fileformat`.

arostamianfar: Please make this a constant.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+    try:
+      # This is a temporary solution until from_string will be fixed.
```
arostamianfar: this part is not temporary? we'll still write to a temp file for parsing headers.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+      self._vcf_reader = nucleus.io.vcf.VcfReader(
+          self._temp_local_file, use_index=False)
+
+    except SyntaxError as e:
```
arostamianfar: Is this actually what Nucleus throws? SyntaxError was thrown by PyVCF in case of malformed headers.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+        self._is_format_repeated[format_info.id] = False
+      else:
+        self._is_format_repeated[format_info.id] = True
+
+  def _get_variant(self, data_line):
+    try:
+      # This is a temporary solution until from_string will be fixed.
+      record = next(self._vcf_reader)
+      return self._convert_to_variant(record)
+    except (LookupError, ValueError) as e:
```
arostamianfar: Again, we should determine what errors Nucleus throws in case of invalid records and adjust accordingly.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+    # type: (nucleus_proto.Variant) -> Variant
+    return Variant(
+        reference_name=record.reference_name,
+        start=record.start,
+        end=record.end,
+        reference_bases=(record.reference_bases
+                         if record.reference_bases != MISSING_FIELD_VALUE
+                         else None),
+        alternate_bases=(
+            map(str, record.alternate_bases) if record.alternate_bases else []),
```
arostamianfar: Is this map needed? We needed this in PyVCF as the type wasn't string. I think in Nucleus it's a string.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+        names=map(str, record.names) if record.names else [],
```
arostamianfar: ditto here. Not sure if the map is needed.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+        quality=record.quality,
```
arostamianfar: I think you should check for has_quality first, as otherwise, this would always return 0 if unspecified (we should set it to null if unspecified).

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+        filters=(
+            [PASS_FILTER] if record.filter == [] else map(str, record.filter)),
```
arostamianfar: I think this is also an artifact of PyVCF (it used an empty list for PASS; i think Nucleus keeps it as PASS?).

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+      elif value.HasField('bool_value'):
+        output_list.append(value.bool_value)
+      elif value.HasField('struct_value'):
+        output_list.append(value.struct_value)
+      elif value.HasField('list_value'):
+        output_list.append(self._convert_list_value(value.list_value, True))
+      else:
+        raise ValueError('ListValue object has an unexpected value: %s' % value)
+
+    if is_repeated:
```
arostamianfar: I think it's more readable to be more explicit about the repeated vs. nonrepeated cases like:
```
if is_repeated:
  return output_list
else:
  if len(output_list) > 1:
     raise Error
  if len(output_list) == 1:
    return output_list[0]
  else:
    return None
  
```

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+      return output_list
+    if not output_list:
+      return None
+    if len(output_list) == 1:
+      return output_list[0]
+    raise ValueError('a not repeated field has more than 1 value')
```
arostamianfar: Interestingly PyVCF doesn't throw an error in these cases and just picks the first value; I wonder if Nucleus treats the entire record as malformed or it actually returns multiple values?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+      call = VariantCall()
+      call.name = sample.call_set_name
+      if not sample.genotype:
+        call.genotype.append(MISSING_GENOTYPE_VALUE)
+      else:
+        for v in sample.genotype:
+          call.genotype.append(v)
+
+      phaseset_from_format = (
+          sample.info[PHASESET_FORMAT_KEY].values[0].string_value
```
arostamianfar: phaseset is actually an integer in the VCF spec (we use a string as we want to use a "*").
Also, should check whether it has a value using convert_list_value

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+    # This member will be properly initiated in _init_with_header().
+    self._vcf_reader = None
+    # These members will be properly initiated in _extract_header_fields().
+    self._is_info_repeated = {}
+    self._is_format_repeated = {}
+    # This is a temporary solution until from_string will be fixed.
+    self._temp_local_file = self._write_to_local_file(file_name)
+
+  def _write_to_local_file(self, remote_file_name):
+    (temp_file, temp_file_name) = tempfile.mkstemp(text=True)
```
arostamianfar: nit: no need for the surrounding brackets

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+      return output_list
+    if not output_list:
+      return None
+    if len(output_list) == 1:
+      return output_list[0]
+    raise ValueError('a not repeated field has more than 1 value')
+
+  def _get_variant_calls(self, record):
+    calls = []
+    for sample in record.calls:
```
arostamianfar: s/sample/call_proto

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+    try:
+      # This is a temporary solution until from_string will be fixed.
+      record = next(self._vcf_reader)
+      return self._convert_to_variant(record)
+    except (LookupError, ValueError) as e:
+      logging.warning('VCF record read failed in %s for line %s: %s',
+                      self._file_name, data_line, str(e))
+      return MalformedVcfRecord(self._file_name, data_line, str(e))
+
+  def _convert_to_variant(self, record):
```
arostamianfar: s/record/variant_proto (we used 'record' before b/c that was the PyVCF naming).

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+    # type: (nucleus_proto.Variant) -> Variant
```
arostamianfar: The type looks wrong as there is no "nucleus_proto" module?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+        alternate_bases=(
+            map(str, record.alternate_bases) if record.alternate_bases else []),
+        names=map(str, record.names) if record.names else [],
+        quality=record.quality,
+        filters=(
+            [PASS_FILTER] if record.filter == [] else map(str, record.filter)),
+        info=self._get_variant_info(record),
+        calls=self._get_variant_calls(record))
+
+  def _get_variant_info(self, record):
```
arostamianfar: s/record/variant_proto

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,179 @@ def _get_variant_calls(self, record, formats):
...
+
+  def _get_variant_calls(self, record):
+    calls = []
+    for sample in record.calls:
+      call = VariantCall()
+      call.name = sample.call_set_name
+      if not sample.genotype:
+        call.genotype.append(MISSING_GENOTYPE_VALUE)
+      else:
+        for v in sample.genotype:
```
arostamianfar: I think you can just do `call.genotype = sample.genotype` (or if it's really not a list, just cast to a list like `list(sample.genotype)`)

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -186,7 +186,8 @@ def __init__(self,
                compression_type=CompressionTypes.AUTO,  # type: str
                buffer_size=DEFAULT_VCF_READ_BUFFER_SIZE,  # type: int
                validate=True,  # type: bool
-               allow_malformed_records=False  # type: bool
+               allow_malformed_records=False,  # type: bool
+               use_nucleus=False  # type: bool
```
arostamianfar: it's more future proof to make an enum for this like `VcfParserType` and use that here (in case we decide to add a third parser). There is even a 'simpler' way of just using the class string and let python actually construct it (so you dont even need an if statement) but that may be less readable.

gcp_variant_transforms/beam_io/vcf_parser.py
samanvp: Now that we have the `from_string()` I added this back.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -571,8 +563,8 @@ def _convert_to_variant(self, variant_proto):
         reference_bases=(variant_proto.reference_bases
                          if variant_proto.reference_bases != MISSING_FIELD_VALUE
                          else None),
-        alternate_bases=variant_proto.alternate_bases,
-        names=map(str, variant_proto.names),
+        alternate_bases=list(variant_proto.alternate_bases),
+        names=variant_proto.names[0].split(';') if variant_proto.names else [],
```
samanvp: We should report this as a bug to Nucleus team.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -633,15 +625,12 @@ def _get_variant_calls(self, variant_proto):
...
-      phaseset_from_format = (
-          self._convert_list_value(call_proto.info[PHASESET_FORMAT_KEY], False)
-          if call_proto.info.get(PHASESET_FORMAT_KEY, None)
-          else None)
-      # Note: Call is considered phased if it contains the 'PS' key regardless
-      # of whether it uses '|'.
-      if phaseset_from_format or call_proto.is_phased:
-        call.phaseset = (phaseset_from_format if phaseset_from_format
-                         else DEFAULT_PHASESET_VALUE)
+      if call_proto.is_phased:
```
samanvp: I modified this if statement, I think this is more straightforward. Let me know what you think.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -22,6 +22,11 @@
 import logging
 from collections import namedtuple
 
+import os
+import tempfile
+
+from nucleus.io.python import vcf_reader as nucleus
```
arostamianfar: s/nucleus/nucleus_vcf_reader so that it doesn't conflict with `import nucleus` itself (in case someone wants to do that in the future).

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -22,6 +22,11 @@
...
+from nucleus.protos import variants_pb2 as nucleus_proto
```
arostamianfar: Please use `variants_pb2` here. The one above is fine as you like to distinguish between pyvcf and nucleus parsers, but you should leave this one as is (e.g. it's easier for searching).

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -41,7 +46,7 @@
 DEFAULT_PHASESET_VALUE = '*'  # Default phaseset value if call is phased, but
                               # no 'PS' is present.
 MISSING_GENOTYPE_VALUE = -1  # Genotype to use when '.' is used in GT field.
-
+FIRST_HEADER_LINE = '##fileformat=VCFv4.2\n'
```
arostamianfar: To clarify:  do all `header_lines` passed via `_init_with_header` include  a `\n`?

arostamianfar: I think we should use 4.0 as the 'least common denominator'. Not sure if it matters for the parser though.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,174 @@ def _get_variant_calls(self, record, formats):
...
+    try:
+      variant_proto = self._vcf_reader.from_string(data_line)
+      return self._convert_to_variant(variant_proto)
+    except (LookupError, ValueError) as e:
+      logging.warning('VCF variant_proto read failed in %s for line %s: %s',
+                      self._file_name, data_line, str(e))
+      return MalformedVcfRecord(self._file_name, data_line, str(e))
+
+  def _convert_to_variant(self, variant_proto):
+    # type: (nucleus.protos.variants_pb2.Variant) -> Variant
```
arostamianfar: nit: just `variants_pb2.Variant` is fine here :)

gcp_variant_transforms/beam_io/vcfio.py
arostamianfar: I think we should. It's very reasonable to have other types of parsers in the future (e.g. we're thinking about a bgzip parser; which may end up being a layer on top of Nuclues). In general, it's a good software engineering practice to not use bools for things that can possibly have more than two values.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -464,3 +469,174 @@ def _get_variant_calls(self, record, formats):
...
+    calls = []
+    for call_proto in variant_proto.calls:
+      call = VariantCall()
+      call.name = call_proto.call_set_name
+      if not call_proto.genotype:
+        call.genotype.append(MISSING_GENOTYPE_VALUE)
+      else:
+        call.genotype = list(call_proto.genotype)
+
+      if call_proto.is_phased:
```
arostamianfar: This is a slight change in logic: previously, we considered a variant phased if it contained a PS field regardless of whether it used `|`. Im in favor of keeping the old logic. What is your reasoning for changing it?

gcp_variant_transforms/beam_io/vcf_parser.py
arostamianfar: What do you think of this suggestion: define `FILE_FORMAT_HEADER_TEMPLATE='##fileformat=VCF{VERSION}'`.
You can then do `startswith(FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='')` ...if it doesn't match, you can add `FILE_FORMAT_HEADER_TEMPLATE.format('v4.0')`. I think it's ok to hard-code the 'v.4.0' for simplicity.

gcp_variant_transforms/beam_io/vcf_parser.py
arostamianfar: I actually don't see `LookupError` in their code. Looks like they only throw `ValueError` in their from_string signature.

gcp_variant_transforms/beam_io/vcf_parser.py
arostamianfar: This is strange. We should investigate this further later. OK for this PR.

gcp_variant_transforms/beam_io/vcf_parser.py
arostamianfar: Right; it seems like they use `-1` instead of null for quality field. We should eventually revisit this and special case on that. OK to leave as TODO for this PR.

gcp_variant_transforms/beam_io/vcf_parser.py
arostamianfar: We should only throw an error if allow_malformed_records is true. But let's first see what Nucleus does. OK to just add a TODO for this PR though.

gcp_variant_transforms/beam_io/vcf_parser.py
arostamianfar: I think in this context, it's better to use consistent names within each class rather than across classes. There are some terminologies that are PyVCF specific that we shouldn't leak into Nucleus ;) (e.g. when/if we remove PyVCF).

PR #347, opened 2018-08-27
export annotations for BQ to VCF pipeline

allieychen:
For BigQuery to VCF pipeline, this PR
- Support the exporting of annotations if there are annotation fields in BigQuery table. 
- Fixed on bug on utf-8 unicode.

Tested: unit tests & manually ran the BQ to VCF.

### DIFF
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -45,7 +45,7 @@ import logging
 import sys
 import tempfile
 from datetime import datetime
-from typing import Iterable, List, Tuple  # pylint: disable=unused-import
+from typing import Dict, Iterable, List, Tuple  # pylint: disable=unused-import
 
 import apache_beam as beam
 from apache_beam.io import filesystems
@@ -160,11 +160,11 @@ def _bigquery_to_vcf_shards(
                                       validate=True,
                                       use_standard_sql=True)
   schema = _get_schema(known_args.input_table)
-  annotation_fields = _extract_annotation_fields(schema)
+  annotation_names = _extract_annotation_names(schema)
   with beam.Pipeline(options=beam_pipeline_options) as p:
     variants = (p
                 | 'ReadFromBigQuery ' >> beam.io.Read(bq_source)
-                | bigquery_to_variant.BigQueryToVariant())
+                | bigquery_to_variant.BigQueryToVariant(annotation_names))
     call_names = (variants
                   | 'CombineCallNames' >>
                   combine_call_names.CallNamesCombiner())
@@ -182,18 +182,7 @@ def _bigquery_to_vcf_shards(
          beam.Map(_pair_variant_with_key, known_args.number_of_bases_per_shard)
          | 'GroupVariantsByKey' >> beam.GroupByKey()
          | beam.ParDo(_get_file_path_and_sorted_variants, vcf_data_temp_folder)
-         | vcfio.WriteVcfDataLines(annotation_fields))
-
-
-def _extract_annotation_fields(schema):
-  # type: (bigquery_v2.TableSchema) -> List[str]
-  for table_field in schema.fields:
-    if table_field.name == bigquery_util.ColumnKeyConstants.ALTERNATE_BASES:
-      for sub_field in table_field.fields:
-        if (sub_field.type.lower() ==
-            bigquery_util.TableFieldConstants.TYPE_RECORD):
-          return [field.name for field in sub_field.fields]
-  return []
+         | vcfio.WriteVcfDataLines())
 
 
 def _get_schema(input_table):
@@ -228,6 +217,27 @@ def _get_bigquery_query(known_args):
   if not conditions:
     return base_query
   return ' '.join([base_query, 'WHERE', ' OR '.join(conditions)])
+
+
+def _extract_annotation_names(schema):
+  # type: (bigquery_v2.TableSchema) -> Dict[str, List[str]]
+  """Returns a mapping of annotation id to the corresponding annotation names.
+
+  The annotation names are useful for reconstructing the annotation str (e.g.,
+  'A|upstream_gene_variant|MODIFIER|PSMF1|||||'). Sample returned map:
+  {'CSQ': ['Consequence', 'IMPACT', 'SYMBOL'],
+   'CSQ_2': ['Codons', 'STRAND']}
+  """
+  annotation_names = {}
+  for table_field in schema.fields:
+    if table_field.name == bigquery_util.ColumnKeyConstants.ALTERNATE_BASES:
+      for sub_field in table_field.fields:
+        if (sub_field.type.lower() ==
+            bigquery_util.TableFieldConstants.TYPE_RECORD):
+          annotation_names.update({
+              sub_field.name: [field.name for field in sub_field.fields]
+          })
+  return annotation_names
 
 
 def _write_vcf_header_with_call_names(sample_names,
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -50,8 +50,11 @@ from typing import Iterable, List, Tuple  # pylint: disable=unused-import
 import apache_beam as beam
 from apache_beam.io import filesystems
 from apache_beam.io.gcp import bigquery
+from apache_beam.io.gcp.internal.clients import bigquery as bigquery_v2
 from apache_beam.options import pipeline_options
 from apache_beam.runners.direct import direct_runner
+
+from oauth2client import client
 
 from gcp_variant_transforms import vcf_to_bq_common
 from gcp_variant_transforms.beam_io import vcf_header_io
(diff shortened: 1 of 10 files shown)


gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -135,7 +139,30 @@ def _bigquery_to_vcf_shards(
...
+
+
+def _extract_annotation_fields(schema):
+  # type: (bigquery_v2.TableSchema) -> List[str]
+  for table_field in schema.fields:
+    if table_field.name == bigquery_util.ColumnKeyConstants.ALTERNATE_BASES:
+      for sub_field in table_field.fields:
+        if (sub_field.type.lower() ==
+            bigquery_util.TableFieldConstants.TYPE_RECORD):
+          return [field.name for field in sub_field.fields]
```
arostamianfar: s/return/yield to support multiple annotation fields.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -135,7 +139,30 @@ def _bigquery_to_vcf_shards(
          beam.Map(_pair_variant_with_key, known_args.number_of_bases_per_shard)
          | 'GroupVariantsByKey' >> beam.GroupByKey()
          | beam.ParDo(_get_file_path_and_sorted_variants, vcf_data_temp_folder)
-         | vcfio.WriteVcfDataLines())
+         | vcfio.WriteVcfDataLines(annotation_fields))
+
+
+def _extract_annotation_fields(schema):
```
arostamianfar: Actually, I think you don't need to precalculate this. For any variant, you can treat any record under alternate_bases as annotation, which can be done entirely within the BigQueryToVariant transform (i.e. just looking at the json record).

allieychen: If I understand correctly, the record we got in `BigQueryToVariant` is a dictionary, which does not preserve the original order anymore.The extraction of annotation fields here is mainly for annotation fields order recovering (e.g., `Format: Allele|Consequence|IMPACT|SYMBOL|Gene|Feature_type|Feature|`).

pgrosu: @allieychen Ask yourself when and if order is important.  It is good to think abstractly of what your are analyzing or processing.  Do you have different Transformer patterns - ideally one per sub-domain - that will do the work for you on the fly?  The answer should be yes, so as to write as little code as possible, and keep reusing these modular, functional components in different compositions.

arostamianfar: @allieychen: I see. You're right. I think in this case we should keep the order as otherwise it may be difficult to merge/compare values from multiple VCF files. Please add a comment about why it's needed though.

pgrosu: @arostamianfar  A merge requires a subset of schema equivalency.  A schema is different than data.  A comparison is an intersection of schema filtered data.  Order is part of the projection operator and not really a requirement in your input data, and joins have those as defaults.  Making that your comparison and merge strategy is the more optimal approach as you scale up.

allieychen:
Thanks Asha on the comments! PTAL :)

gcp_variant_transforms/libs/bigquery_vcf_data_converter.py
```diff
@@ -299,6 +316,32 @@ def _get_variant_calls(self, variant_call_records):
...
+                       'cannot be reconstructed since the order is not '
+                       'provided.')
+    annotation_str = []
+    for annotation_map in alt_base.get(annotation_id):
+      annotation_values = [
+          alt_base.get(bigquery_util.ColumnKeyConstants.ALTERNATE_BASES_ALT)]
+      for annotation_name in self._annotation_id_to_annotation_names.get(
+          annotation_id):
+        value = annotation_map.get(annotation_name)
+        if not value and value != 0:
```
arostamianfar: I think just doing `if value is None` is easier to understand here (everything else should be taken care of by `str(value)` below).

gcp_variant_transforms/libs/bigquery_vcf_data_converter_test.py
```diff
@@ -701,6 +701,58 @@ def test_get_variant_info(self):
     self.assertEqual(expected_variant_info,
                      self._variant_generator._get_variant_info(row))
 
+  def test_get_variant_info_annotation(self):
+    variant_generator = bigquery_vcf_data_converter.VariantGenerator({
+        'CSQ': ['Consequence', 'AF', 'IMPACT']
+    })
+    row = {
+        unicode(ColumnKeyConstants.ALTERNATE_BASES): [
```
arostamianfar: Is there a way to avoid having these `unicode` casts? The reason is that `unicode` is not supported in python3, so it makes the transition to py3 easier if we don't have these everywhere here.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -138,6 +142,39 @@ def _bigquery_to_vcf_shards(
...
+  table = bigquery_client.tables.Get(bigquery_v2.BigqueryTablesGetRequest(
+      projectId=project_id, datasetId=dataset_id, tableId=table_id))
+  return table.schema
+
+
+def _extract_annotation_names(schema):
+  # type: (bigquery_v2.TableSchema) -> Dict[str, List[str]]
+  """Returns a mapping of annotation id to the corresponding annotation names.
+
+  The annotation names are useful for reconstructing the annotation str (e.g.,
```
arostamianfar: Please add a comment about the assumption of annotation names (i.e. they're a record within alternate bases; any such record is considered an annotation field).

allieychen:
@bashir2 I am planning to add an `Unparser` class in `annotation_parser.py` (may change the name to `annotation_parser_and_unparser`), as the reversed functionality of `Parser`. I will let you know when it is ready for review. :)

gcp_variant_transforms/libs/bigquery_vcf_data_converter_test.py
allieychen: Well, we don't have to use unicode for the unit tests, we can definitely make it to string. I used them here to emphasize that these keys and values are unicode when we loaded from BigQuery, we may have issues if we use `str()` to convert it. Do you prefer to use string instead?

gcp_variant_transforms/libs/bigquery_vcf_data_converter.py
```diff
@@ -279,7 +295,12 @@ def _get_variant_info(self, row):
             not self._is_null_or_empty(value)):
           if key not in info:
             info[key] = []
-          info[key].append(value)
+          if self._annotation_unparser.is_valid_annotation_id(key):
+            info[key].extend(
+                self._annotation_unparser.reconstruct_annotation_str(
+                    key, value))
```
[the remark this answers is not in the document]
allieychen: Done. Emphasized it in `reconstruct_annotation_str`.

gcp_variant_transforms/libs/annotation/annotation_parser.py
```diff
@@ -301,6 +301,79 @@ def _find_alt_index_by_allele_num(self, annotation_map):
...
+    for annotation_map in annotation_maps:
+      annotation_values = []
+      for annotation_name in self._annotation_id_to_annotation_names.get(
+          annotation_id):
+        value = annotation_map.get(annotation_name)
+        if value is None:
+          annotation_values.append(_MISSING_ANNOTATION_FIELD_VALUE)
+        else:
+          annotation_values.append(str(value))
+      annotation_strs.append('|'.join(annotation_values))
```
bashir2: I am pretty sure I wrote this comment in my previous review too but not sure where it was dropped in the process:

Have you considered making this into a generator and `yield` here instead of returning the list? While in general I prefer lists as returned value (as it makes code more readable) when the returned value is big enough, then it's worth considering generators as it avoids a big memory copy of the returned value and may have non-insignificant performance consequences.

allieychen: @bashir2 Thanks for the comment! The returned/yielded value is extended to another list, which is saved as the value of the map `info[KEY]` (e.g. KEY = 'CSQ'). Do you mean use info[KEY]=generator rather than info[KEY] = list[str]? I feel it is not feasible to use a generator as the value here.

allieychen: Thanks Bashir for sharing your thoughts offline. PTAL :)

pgrosu: @allieychen Under what situation do you think can generators be an issue?  What do you think the approach should be in that situation?

gcp_variant_transforms/libs/annotation/annotation_parser.py
```diff
@@ -341,7 +341,7 @@ def reconstruct_annotation_str(self, annotation_id, annotation_maps):
         ]
 
     Returns:
```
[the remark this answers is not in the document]
allieychen: Done. Thanks.

Issue #184, opened 2018-04-12
Document all settings/flags for robust import into BQ


nmousavi:
This includes:

- allow_incompatible_records flag:
   This flag was initially added to allow casting VCF record values to BQ schema in case there is mismatch. We later extended its use to allow allow resolving conflict in header definitions either defined or inferred (see PR #181 ). 

- infer-undefined-header-fields

PR #357, opened 2018-09-05
Document flags for robust Variant Transforms.

nmousavi:
Fixes issue #184

### DIFF
--- a/docs/malformed_files.md
+++ b/docs/malformed_files.md
@@ -75,5 +75,5 @@ and actual values or if a field has two inconsistent definitions in two
 different VCF files.
 By specifying `--allow_incompatible_records`, pipeline will resolve conflicts
 in header definitons. It will also cast field values to match BigQuery schema if
-there is a mismatch between field definition and field value (e.g. Integer field
-value is casted to String to match a field schema of type String).
+there is a mismatch between field definition and field value (e.g. `Integer` field
+value is casted to `String` to match a field schema of type `String`).
--- a/docs/malformed_files.md
+++ b/docs/malformed_files.md
@@ -65,7 +65,9 @@ for:
 If this flag is set, pipeline will infer `TYPE` and `NUMBER` for undefined
 fields based on field values seen in VCF files. It will also output a
 representive header that contains infered definitions as well as definitions
-from headers. Use this flag if there are fields with missing definition.
+from headers. Use this flag if there are fields with missing definition or if
+pipeline should ignore header definitions that are incompatible with field values,
+and instead should infer the correct header definitions for the corresponding fields.
 
 
 ## Specifying `--allow_incompatible_records`
--- a/docs/malformed_files.md
+++ b/docs/malformed_files.md
@@ -1,5 +1,6 @@
 # Dealing with malformed files
 
+
 ## Field compatibility
 
 When loading multiple files, the `INFO` and `FORMAT` fields
@@ -15,9 +16,10 @@ The compatibility rules are as follows:
 * Fields with different `Type` values are compatible in the following cases:
 
   * `Integer` and `Float` fields are compatible and are converted to `Float`.
-  * We are adding more compatibility options (e.g. resolving `String` and
-  `Integer` to `String`). See
-  [below](#specifying---allow-incompatible-records) for more details.
+  * You must run the pipeline with --allow-incompatible-records to
+    automatically resolve conflicts between incompatible fields (e.g. String
+    and Integer). This is to ensure incompatible types are not silently ignored.
+    See [below](#specifying---allow-incompatible-records) for more details.
 
 * Fields with different `Number` values are compatible in the following cases:
 
@@ -29,10 +31,15 @@ The compatibility rules are as follows:
     * `Number=A` (one value for each alternate) only if running with
       `--split_alternate_allele_info_fields False`.
 
-  * We are adding more compatibility options (e.g. resolving `Number=1` and
-    `Number=.` with `Number=.`). See
-    [below](#specifying---allow-incompatible-records) for more details.
-
+  * You must run the pipeline with --allow-incompatible-records to
+    automatically resolve conflicts between incompatible fields (e.g.
+    `Number=1` and `Number=.`). This is to ensure incompatible types
+    are not silently ignored.
+    See [below](#specifying---allow-incompatible-records) for more details.
+    
+You can run preprossing tool to get a summary of malformed/incompatible
+records. Please refer to 
+[VCF files preprocessor](docs/vcf_files_preprocessor.md) for more details.
 
 ## Specifying `--representative_header_file`
 
@@ -50,7 +57,8 @@ for:
   files have identical VCF headers.
 * Providing definitions for missing header fields. See the
   [troubleshooting page](./troubleshooting.md) for more details.
-* Coming soon: resolving incompatible fields across files.
+* Resolving incompatible field definition across files. See
+  [below](#specifying---allow-incompatible-records) for an alternative.
 
 ## Specifying `--infer-undefined-header-fields`
 
@@ -67,6 +75,5 @@ and actual values or if a field has two inconsistent definitions in two
 different VCF files.
 By specifying `--allow-incompatible-records`, pipeline will resolve conflicts
 in header definitons. It will also cast field values to match BigQuery schema if
-there is a mismatch between field definition and field value. Please see [this
-doc](https://docs.google.com/document/d/1JusaPllbpXynNziNzdyOdSRTM8i7UXbNvtkbrWNMq3Q)
-for more details.
+there is a mismatch between field definition and field value (e.g. Integer field
+value is casted to String to match a field schema of type String).
--- a/docs/malformed_files.md
+++ b/docs/malformed_files.md
@@ -1,0 +1,72 @@
+# Dealing with malformed files
+
+## Field compatibility
+
+When loading multiple files, the `INFO` and `FORMAT` fields
+from all VCF files are merged together to generate a "representative header",
+which is then used to generate a single BigQuery schema. If the same key is
+defined in multiple files, then its definition must be compatible across files.
+The compatibility rules are as follows:
+
+* Fields are compatible if they have the same `Number` and `Type` fields.
+  Annotation fields (i.e. those specified by `--annotation_fields`) must also
+  have the same `Description`.
+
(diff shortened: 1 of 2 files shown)


docs/malformed_files.md
```diff
@@ -0,0 +1,72 @@
...
+The compatibility rules are as follows:
+
+* Fields are compatible if they have the same `Number` and `Type` fields.
+  Annotation fields (i.e. those specified by `--annotation_fields`) must also
+  have the same `Description`.
+
+* Fields with different `Type` values are compatible in the following cases:
+
+  * `Integer` and `Float` fields are compatible and are converted to `Float`.
+  * We are adding more compatibility options (e.g. resolving `String` and
```
arostamianfar: Please update. "We have added...." or add something like: "You must run the pipeline with --allow-incompatible-records to automatically resolve conflicts between incompatible fields (e.g. String and Integer). This is to ensure incompatible types are not silently ignored.".

docs/malformed_files.md
```diff
@@ -0,0 +1,72 @@
...
+
+  * "Repeated" numbers are compatible with each other. They include:
+    * `Number=.` (unknown number)
+    * Any `Number` greater than 1.
+    * `Number=G` (one value per genotype) and `Number=R` (one value for each
+      alternate and reference).
+    * `Number=A` (one value for each alternate) only if running with
+      `--split_alternate_allele_info_fields False`.
+
+  * We are adding more compatibility options (e.g. resolving `Number=1` and
```
arostamianfar: Please update here as well.

docs/malformed_files.md
```diff
@@ -0,0 +1,72 @@
...
+fields or can point to an actual VCF file. Providing this file can be useful
+for:
+
+* Speeding up the pipeline especially if a large number of files are provided.
+  The pipeline will use the provided file to generate the BigQuery schema and
+  will skip merging headers across files. This is particularly useful if all
+  files have identical VCF headers.
+* Providing definitions for missing header fields. See the
+  [troubleshooting page](./troubleshooting.md) for more details.
+* Coming soon: resolving incompatible fields across files.
```
arostamianfar: Please update here too. We already do this and treat the number/type from the header fields regardless of what is provided in the file.

docs/malformed_files.md
```diff
@@ -0,0 +1,72 @@
...
+
+
+## Specifying `--allow-incompatible-records`
+
+Pipeline will fail by defualt if there is a mismatch between field definition
+and actual values or if a field has two inconsistent definitions in two
+different VCF files.
+By specifying `--allow-incompatible-records`, pipeline will resolve conflicts
+in header definitons. It will also cast field values to match BigQuery schema if
+there is a mismatch between field definition and field value. Please see [this
```
arostamianfar: Please provide an example (e.g. String and Integer -> String).

arostamianfar: Let's not link to that doc. It's too confusing for users. You can summarize any other interesting/important notes from there to here.

docs/malformed_files.md
```diff
@@ -0,0 +1,72 @@
+# Dealing with malformed files
+
+## Field compatibility
```
arostamianfar: Please add a note about running the preprocessor tool here. It will give a summary of malformed/incompatible records.

nmousavi: Done

docs/malformed_files.md
nmousavi: Done

docs/malformed_files.md
nmousavi: Done

docs/malformed_files.md
nmousavi: Done

docs/malformed_files.md
nmousavi: Done

nmousavi: Done

docs/malformed_files.md
```diff
@@ -0,0 +1,79 @@
...
+
+  * "Repeated" numbers are compatible with each other. They include:
+    * `Number=.` (unknown number)
+    * Any `Number` greater than 1.
+    * `Number=G` (one value per genotype) and `Number=R` (one value for each
+      alternate and reference).
+    * `Number=A` (one value for each alternate) only if running with
+      `--split_alternate_allele_info_fields False`.
+
+  * You must run the pipeline with --allow-incompatible-records to
```
allieychen: I find there are multiple places for the flags that are using `-` instead of the underline. This is one example of that.

allieychen: Also please add `` for the flag here.

docs/malformed_files.md
```diff
@@ -0,0 +1,79 @@
...
+    automatically resolve conflicts between incompatible fields (e.g.
+    `Number=1` and `Number=.`). This is to ensure incompatible types
+    are not silently ignored.
+    See [below](#specifying---allow-incompatible-records) for more details.
+    
+You can run preprossing tool to get a summary of malformed/incompatible
+records. Please refer to 
+[VCF files preprocessor](docs/vcf_files_preprocessor.md) for more details.
```
allieychen: The link is not working.

docs/malformed_files.md
```diff
@@ -0,0 +1,79 @@
...
+* Speeding up the pipeline especially if a large number of files are provided.
+  The pipeline will use the provided file to generate the BigQuery schema and
+  will skip merging headers across files. This is particularly useful if all
+  files have identical VCF headers.
+* Providing definitions for missing header fields. See the
+  [troubleshooting page](./troubleshooting.md) for more details.
+* Resolving incompatible field definition across files. See
+  [below](#specifying---allow-incompatible-records) for an alternative.
+
+## Specifying `--infer-undefined-header-fields`
```
allieychen: s/--infer-undefined-header-fields/--infer_headers or --infer_undefined_headers.
I would prefer to use `infer_headers` here since the headers are inferred not just for undefined headers, but also when there are mismatches between the header and the field value. `infer_undefined_headers` is still supported for compatibility.

docs/malformed_files.md
```diff
@@ -0,0 +1,79 @@
...
+
+If this flag is set, pipeline will infer `TYPE` and `NUMBER` for undefined
```
allieychen: Please also mention the mismatches between the header and the field value when suing this flag.

docs/malformed_files.md
nmousavi: I was not aware of changing this flag to `infer_headers` and `infer_undefined_headers`. Done!

docs/malformed_files.md
nmousavi: Can you elaborate on this? Not sure if I understand what you mean.

allieychen: Sorry for the confusion.

The `infer_headers` has been updated recently ([PR 283](https://github.com/googlegenomics/gcp-variant-transforms/pull/283)), and now two types of fields are inferred (for INFO, for instance),
    - The info fields are undefined in the headers.
    - The info fields' definitions provided by the header does not match the field value.

Most related [method](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/transforms/infer_headers.py#L222). 

For instance, if the header is defined as integer, but a float value is provided, we will infer/correct the header to be float.

nmousavi: Done, PTAL

docs/malformed_files.md
```diff
@@ -0,0 +1,81 @@
...
+
+* Fields are compatible if they have the same `Number` and `Type` fields.
+  Annotation fields (i.e. those specified by `--annotation_fields`) must also
+  have the same `Description`.
+
+* Fields with different `Type` values are compatible in the following cases:
+
+  * `Integer` and `Float` fields are compatible and are converted to `Float`.
+  * You must run the pipeline with `--allow_incompatible_records` to
+    automatically resolve conflicts between incompatible fields (e.g. String
```
arostamianfar: nit: please put String and Integer in backquotes to match the style of other types.

docs/malformed_files.md
```diff
@@ -0,0 +1,81 @@
...
+
+  * "Repeated" numbers are compatible with each other. They include:
+    * `Number=.` (unknown number)
+    * Any `Number` greater than 1.
+    * `Number=G` (one value per genotype) and `Number=R` (one value for each
+      alternate and reference).
+    * `Number=A` (one value for each alternate) only if running with
+      `--split_alternate_allele_info_fields False`.
+
+  * You must run the pipeline with --allow_incompatible-records to
```
arostamianfar: Please put backquotes around the flag too.

docs/malformed_files.md
```diff
@@ -0,0 +1,81 @@
...
+* Providing definitions for missing header fields. See the
+  [troubleshooting page](./troubleshooting.md) for more details.
+* Resolving incompatible field definition across files. See
+  [below](#specifying---allow-incompatible-records) for an alternative.
+
+## Specifying `--infer_headers`
+
+If this flag is set, pipeline will infer `TYPE` and `NUMBER` for undefined
+fields based on field values seen in VCF files. It will also output a
+representive header that contains infered definitions as well as definitions
```
arostamianfar: typo in inferred

docs/malformed_files.md
```diff
@@ -0,0 +1,81 @@
...
+from headers. Use this flag if there are fields with missing definition or if
+pipeline should ignore header definitions that are incompatible with field values,
+and instead should infer the correct header definitions for the corresponding fields.
```
arostamianfar: nit: >80 chars.

arostamianfar: Please also add a note that this will do two passes on the data, which results in ~30% more compute.

docs/malformed_files.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+## Specifying `--allow_incompatible_records`
+
+Pipeline will fail by defualt if there is a mismatch between field definition
```
arostamianfar: typo in default.

docs/malformed_files.md
nmousavi: Done

docs/malformed_files.md
nmousavi: Done

nmousavi: Done!

docs/malformed_files.md
nmousavi: Done

PR #359, opened 2018-09-13
direct runner support for BQ To VCF #85

allieychen:
Add the support for BQ to VCF with direct runner.

### DIFF
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -89,7 +89,7 @@ def run(argv=None):
     raise ValueError('Please set the output file {} to GCS when running with '
                      'DataflowRunner.'.format(known_args.output_file))
   if is_direct_runner:
-    known_args.number_of_bases_per_shard = -1
+    known_args.number_of_bases_per_shard = sys.maxsize
 
   temp_folder = google_cloud_options.temp_location or tempfile.mkdtemp()
   timestamp_str = datetime.now().strftime('%Y%m%d_%H%M%S')
@@ -262,11 +262,11 @@ def _pair_variant_with_key(variant, number_of_variants_per_shard):
   # type: (vcfio.Variant, int) -> Tuple[str, vcfio.Variant]
   """Pairs the variant with a key.
 
-  If `number_of_variants_per_shard` is -1, the variants are paired with the
-  `reference_name`. Otherwise, the key is determined by the variant's reference
-  name and start position.
+  If `number_of_variants_per_shard` is sys.maxsize, the variants are paired
+  with the `reference_name`. Otherwise, the key is determined by the variant's
+  reference name and start position.
   """
-  if number_of_variants_per_shard == -1:
+  if number_of_variants_per_shard == sys.maxsize:
     return (variant.reference_name, variant)
   return ('%s_%011d' % (variant.reference_name,
                         variant.start // number_of_variants_per_shard *
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -39,6 +39,8 @@ python -m gcp_variant_transforms.bq_to_vcf \
   --runner DataflowRunner
 """
 
+from __future__ import division
+
 import logging
 import sys
 import tempfile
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -71,21 +81,20 @@ def run(argv=None):
...
-  if not google_cloud_options.temp_location or not google_cloud_options.project:
-    raise ValueError('temp_location and project must be set.')
 
+  temp_folder = google_cloud_options.temp_location or tempfile.mkdtemp()
   timestamp_str = datetime.now().strftime('%Y%m%d_%H%M%S')
   vcf_data_temp_folder = filesystems.FileSystems.join(
-      google_cloud_options.temp_location,
+      temp_folder,
       'bq_to_vcf_data_temp_files_{}'.format(timestamp_str))
+  filesystems.FileSystems.mkdirs(vcf_data_temp_folder)
```
allieychen: This line is needed when using direct runner. Otherwise, It won't be able to create the file if the directory is not exist.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -228,7 +255,27 @@ def _get_file_path_and_sorted_variants((file_name, variants), file_path_prefix):
...
+                           number_of_variants_per_shard):
+  # type: (vcfio.Variant, bool, int) -> Tuple[str, vcfio.Variant]
+  """Pairs the variant with a key.
+
+  For direct runner, all variants are paired with the same key. Otherwise, the
+  key is determined by the variant's reference name and start position.
+  """
+  if is_direct_runner:
+    return (_LOCAL_TEMP_DATA_FILE_NAME, variant)
   return ('%s_%011d' % (variant.reference_name,
```
arostamianfar: I just realized that we should do `from __future__ import division` at the top so that we get python3 division behavior. You should then do `//` here to get an integer. See https://www.python.org/dev/peps/pep-0238/ for details.

gcp_variant_transforms/libs/vcf_file_composer.py
```diff
@@ -67,6 +67,33 @@ def compose_vcf_shards(project,  # type: str
...
+    vcf_header_file_path: The path of the VCF header file, it contains the meta
+      information, as well as the data header line with the call names.
+    vcf_data_files_folder: The folder that contains all VCF data files.
+    output_file: The final VCF file path.
+  """
+  with filesystems.FileSystems.create(output_file) as file_to_write:
+    file_to_write.write(
+        filesystems.FileSystems.open(vcf_header_file_path).read())
+    data_pattern = filesystems.FileSystems.join(vcf_data_files_folder, '*')
+    data_file_paths = [file_metadata.path
```
arostamianfar: you can add `sorted([...])` to make sure the list is sorted.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -70,34 +81,47 @@ def run(argv=None):
...
-  if not google_cloud_options.temp_location or not google_cloud_options.project:
-    raise ValueError('temp_location and project must be set.')
-
+  if not google_cloud_options.project:
+    raise ValueError('project must be set.')
+  if not is_direct_runner and not known_args.output_file.startswith('gs://'):
+    raise ValueError('Please set the output file {} to GCS when running with '
+                     'DataflowRunner.'.format(known_args.output_file))
+  if is_direct_runner:
+    known_args.number_of_bases_per_shard = -1
```
arostamianfar: Sorry Allie: I now think it's better to use `sys.maxsize` instead of `-1` here. I was originally thinking that we'd want to avoid overflow, but given that we're special-casing that value anyway, then we'd need to create a proper constant for `-1` too, which is strange. We can use `maxsize` as a defined constant representing "infinite" range.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -70,34 +81,47 @@ def run(argv=None):
   known_args, pipeline_args = vcf_to_bq_common.parse_args(argv,
                                                           _COMMAND_LINE_OPTIONS)
   options = pipeline_options.PipelineOptions(pipeline_args)
+  is_direct_runner = _is_direct_runner(beam.Pipeline(options=options))
   google_cloud_options = options.view_as(pipeline_options.GoogleCloudOptions)
-  # TODO(allieychen): Add support for local location.
-  if not google_cloud_options.temp_location or not google_cloud_options.project:
-    raise ValueError('temp_location and project must be set.')
-
+  if not google_cloud_options.project:
```
arostamianfar: Actually, why is project required for direct runner?

PR #361, opened 2018-09-18
Initial commit of annotation service.

Jessime:
The pipeline has had a successful end-to-end run of annotating variants
and inserting them into BQ. Many components are inflexible, ignore edge
cases, or are otherwise brittle. Efforts have been made to document
current issues with TODO comments.


bashir2:
Thanks Jessime for your efforts on making this prototype work, I am really glad with the speed-up we have got. I think given the circumstances, the best way going forward is to check in the server side code (on the `variant-annotation` repo) and leave this repo for now (without checking in) until we prioritize the continuation/productionization of this work for Q4 or beyond. Adding Asha as FYI.

arostamianfar:
Closing this PR as we need to revisit the design details and some of the TODOs. Github will keep the history, so we can always revisit the diffs from this PR.

PR #363, opened 2018-09-18
Add a new flag for vcf_parser_type

samanvp:
and propagate its value all the way from:
vcf_to_bq._read_variants() -> vcfio.ReadFromVcf() -> vcfio._VcfSource()

We will pick the corresponding parser class (either vcf_parser.PyVcfParser
or vcf_parser.NucleusParser) based on this flag's value.


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -94,13 +94,24 @@ def add_arguments(self, parser):
               'of the the header fields do not match the field values. Note: '
               'setting this flag or `--infer_annotation_types` incurs a '
               'performance penalty of an extra pass over all variants.'))
+    parser.add_argument(
+        '--vcf_parser_type',
```
arostamianfar: nit: I think just `--vcf_parser` is a better name for this option. We added `type` for the enum to distinguish an actual class from class choices, but in here, it's clear that it's not really an actual class :)

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -94,13 +94,24 @@ def add_arguments(self, parser):
...
+        default='PyVcf',
+        help=('Choose the underlying parser for reading VCF files. Currently '
+              'we only support `PyVcf` (default) and `Nucleus`. Note: Nucleus '
+              'parser is still in experimental stage and we do not recommend '
```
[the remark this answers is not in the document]
samanvp: done.

samanvp:
All done, thanks for good comments.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -251,6 +251,7 @@ def __init__(
       compression_type=CompressionTypes.AUTO,  # type: str
       validate=True,  # type: bool
       allow_malformed_records=False,  # type: bool
+      vcf_parser_type=VcfParserType.PYVCF, # type: int
```
arostamianfar: nit: please use two spaces before `#`

samanvp: done.

samanvp:
Integration tests passed. I also confirmed the correct propagation of flag value using DirectRunner tests.

PR #366, opened 2018-09-19
remove the utf-8 encoded data from nucleus

allieychen:
A quick fix on the unit tests failure for nucleus with utf-8 encoded data.

- May revisit this test case when nucleus related functions are ready.

Tested: unit tests


samanvp:
Sorry, I saw this a bit late.
Thanks Asha for quick approve.

PR #367, opened 2018-09-20
Write VCF header generated from schema #85

allieychen:
- Writes the header when `representative_header_file` is not present.
- Updated the `TableFieldConstants` to be consistent with BigQuery schema.

Tested: unit tests, integration tests & manually ran bq_to_vcf.

### DIFF
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -60,6 +60,7 @@ from gcp_variant_transforms import vcf_to_bq_common
 from gcp_variant_transforms.beam_io import vcf_header_io
 from gcp_variant_transforms.beam_io import vcfio
 from gcp_variant_transforms.libs import bigquery_util
+from gcp_variant_transforms.libs import bigquery_vcf_schema_converter
 from gcp_variant_transforms.libs import genomic_region_parser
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -127,10 +129,11 @@ def run(argv=None):
...
 
 
-def _write_vcf_meta_info(representative_header_file):
-  # TODO(allieychen): infer the meta information from the schema and replace the
-  # empty `VcfHeader` below.
-  header_fields = vcf_header_io.VcfHeader()
+def _write_vcf_meta_info(input_table, representative_header_file):
+  # type: (str, str) -> None
+  """Writes the meta information generated from BigQuery schema."""
+  header_fields = (bigquery_vcf_schema_converter.
```
arostamianfar: nit: please try to avoid breaking on the `.` as it looks like it's a tuple of multiple objects rather than one object....does this work?
```
  header_fields = (
      bigquery_vcf_schema_converter. generate_header_fields_from_schema(
          _get_schema(input_table)))
```

allieychen: Thank you Asha!

PR #368, opened 2018-09-21
create integration tests for BQ to VCF

allieychen:
Add integration tests for BQ to VCF pipeline. More test cases (json file) will be added in the folder `bq_to_vcf_tests`.

Tested: manually ran run_bq_to_vcf_tests.


allieychen:
Thanks Asha for the review! I agree that checking the contents will be more straightforward (and also informative). PTAL :)
P.S.: More test cases will be added in another PR.

gcp_variant_transforms/testing/integration/bq_to_vcf_tests/4_0_with_representative_header_file.json
```diff
@@ -0,0 +1,11 @@
+[
+  {
+    "test_name": "bq-to-vcf-4-0-with-representative-header-file",
+    "input_table": "gcp-variant-transforms-test:bq_to_vcf_integration_tests.4_0",
+    "representative_header_file": "gs://gcp-variant-transforms-testfiles/small_tests/valid-4.0.vcf",
+    "output_file_name": "bq_to_vcf_4_0_with_representative_header_file.vcf",
+    "runner": "DirectRunner",
+    "zones": ["us-west1-b"],
+    "expected_output_file": "4_0_with_representative_header_file.vcf"
```
arostamianfar: Why not just specify the full path here? That way, we'll have more control over how we want to organize the expected output files and can even store it in GCS (e.g. for larger files).

gcp_variant_transforms/testing/data/vcf/bq_to_vcf/4_0_with_representative_header_file.vcf
```diff
@@ -0,0 +1,23 @@
+##fileformat=VCFv4.0
```
arostamianfar: nit: should this be under `...vcf/bq_to_vcf/expected_output/...` instead? We can have different representative header files or other arbitrary files here, so I think it's better to be explicit about which ones are about 'expected' output and which ones specify the input params.

PR #369, opened 2018-09-26
add sample filter

allieychen:
In BQ to VCF pipeline, add one option for the user to load the selected samples.

### DIFF
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -49,6 +49,7 @@ from datetime import datetime
 from typing import Dict, Iterable, List, Tuple  # pylint: disable=unused-import
 
 import apache_beam as beam
+from apache_beam import transforms
 from apache_beam.io import filesystems
 from apache_beam.io.gcp import bigquery
 from apache_beam.io.gcp.internal.clients import bigquery as bigquery_v2
@@ -170,9 +171,14 @@ def _bigquery_to_vcf_shards(
     variants = (p
                 | 'ReadFromBigQuery ' >> beam.io.Read(bq_source)
                 | bigquery_to_variant.BigQueryToVariant(annotation_names))
-    call_names = (variants
-                  | 'CombineCallNames' >>
-                  combine_call_names.CallNamesCombiner())
+    if known_args.sample_names:
+      call_names = (p
+                    | transforms.Create(known_args.sample_names)
+                    | beam.combiners.ToList())
+    else:
+      call_names = (variants
+                    | 'CombineCallNames' >>
+                    combine_call_names.CallNamesCombiner())
 
     _ = (call_names
          | 'GenerateVcfDataHeader' >>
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -496,3 +496,9 @@ def add_arguments(self, parser):
               'chromosome is requested. Only variants matching at least one '
               'of these regions will be loaded. If this parameter is not '
               'specified, all variants will be kept.'))
+    parser.add_argument(
+        '--sample_names',
```
arostamianfar: nit: s/sample_names/call_names to be consistent with naming.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -169,9 +170,14 @@ def _bigquery_to_vcf_shards(
     variants = (p
                 | 'ReadFromBigQuery ' >> beam.io.Read(bq_source)
                 | bigquery_to_variant.BigQueryToVariant(annotation_names))
-    call_names = (variants
-                  | 'CombineCallNames' >>
-                  combine_call_names.CallNamesCombiner())
+    if known_args.sample_names:
```
arostamianfar: Please add a TODO to modify the SQL query to only return the specified call names to reduce the amount of data being processed from the beginning (I think the query would be more complex, so it's fine to leave as is for now).

allieychen: That's a good point! Add a TODO for now.

gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -158,7 +159,7 @@ def _bigquery_to_vcf_shards(
   Also, it writes the meta info and data header with the call names to
   `vcf_header_file_path`.
   """
-
+  # TODO (allieychen): Modify the SQL query with the specified call_names.
```
arostamianfar: nit: remove space before `(`

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -496,3 +496,9 @@ def add_arguments(self, parser):
               'chromosome is requested. Only variants matching at least one '
               'of these regions will be loaded. If this parameter is not '
               'specified, all variants will be kept.'))
+    parser.add_argument(
+        '--call_names',
+        default=None, nargs='+',
+        help=('A list of call names (separated by a space). Only variant '
```
arostamianfar: nit: `Only variants for these calls will be ...` and also, s/samples/calls at the end.

gcp_variant_transforms/libs/bigquery_vcf_schema_converter.py
```diff
@@ -264,9 +264,13 @@ def _add_info_fields_from_alternate_bases(schema, infos):
   Notice that the validation of field mode is skipped for reserved fields since
   the mode (NULLABLE) of field in alternate bases is expected to be different
   from the mode (REPEATED) in reserved field definition.
+
+  Any `Record` field within alternate bases is considered as an annotation
+  field, and the annotation fields are skipped.
   """
   for field in schema.fields:
-    if field.name in _CONSTANT_ALTERNATE_BASES_FIELDS:
+    if (field.name in _CONSTANT_ALTERNATE_BASES_FIELDS or
```
arostamianfar: This seems unrelated to this PR; usually it's better to do it in another PR, but I leave it up to you as it's a small change anyway.

PR #373, opened 2018-10-03
load public BQ table to VCF

allieychen:
Update the code to support the public genomic tables:
- Ignore the schema field if it has type DATE.
- Support the MODE=None fields. When trying to retrieve the mode for NULLABLE fields, it is  actually `None` instead of `NULLABLE` for the public tables (for instance, 
`bigquery-public-data:human_genome_variants.1000_genomes_phase_3_variants_20150220`).

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_util.py
+++ b/gcp_variant_transforms/libs/bigquery_util.py
@@ -59,11 +59,11 @@ class _SupportedTableFieldType(enum.Enum):
 
   Only schema fields with these types are interchangeable with VCF.
   """
-  TYPE_STRING = 'STRING'
-  TYPE_INTEGER = 'INTEGER'
-  TYPE_RECORD = 'RECORD'
-  TYPE_FLOAT = 'FLOAT'
-  TYPE_BOOLEAN = 'BOOLEAN'
+  TYPE_STRING = TableFieldConstants.TYPE_STRING
+  TYPE_INTEGER = TableFieldConstants.TYPE_INTEGER
+  TYPE_RECORD = TableFieldConstants.TYPE_RECORD
+  TYPE_FLOAT = TableFieldConstants.TYPE_FLOAT
+  TYPE_BOOLEAN = TableFieldConstants.TYPE_BOOLEAN
 
 
 # A map to convert from VCF types to their equivalent BigQuery types.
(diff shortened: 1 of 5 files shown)


gcp_variant_transforms/libs/bigquery_vcf_schema_converter.py
```diff
@@ -325,6 +326,8 @@ def _validate_reserved_field_type(field_schema, reserved_definition):
 
 def _validate_reserved_field_mode(field_schema, reserved_definition):
   schema_mode = field_schema.mode
+  if not schema_mode:
```
arostamianfar: nit: more pythonic to do `schema_mode = field_schema.mode or bigquery_util.TableFieldConstants.MODE_NULLABLE`

gcp_variant_transforms/libs/bigquery_vcf_schema_converter.py
```diff
@@ -203,7 +203,8 @@ def generate_header_fields_from_schema(schema, allow_incompatible_schema=False):
   infos = OrderedDict()  # type: OrderedDict[str, _Info]
   formats = OrderedDict()  # type: OrderedDict[str, _Format]
   for field in schema.fields:
-    if field.name in _NON_INFO_OR_FORMAT_CONSTANT_FIELDS:
+    if (field.type == bigquery_util.TableFieldConstants.TYPE_DATE or
```
arostamianfar: Consider doing a "whitelist" rather than a "blacklist". In the current way ("blacklist" approach) you'd also need to include other types like TIMESTAMP (or any new field that gets added later). I think you can create a helper method like `_get_supported_bigquery_schema_types` that returns all supported types (ideally, we'd use an enum to return this programmatically, but ok to hard code it for now).

allieychen:
Thanks Asha for the review! I realized that we also need to remove the date (e.g., `partition_date_please_ignore=2018-08-23`) from the variant info. I will submit another PR to include only columns that have the supported types when forming the SQL query. But maybe in a long term we can provide an option for the user to select/exclude some columns as one of the customized export functions.

allieychen:
I also added the enum class as you mentioned, PTAL :-)

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -53,6 +54,18 @@ class TableFieldConstants(object):
   MODE_REPEATED = 'REPEATED'
 
 
+class _SupportedTableFieldType(enum.Enum):
+  """The supported BigQuery field types.
+
+  Only schema fields with these types are interchangeable with VCF.
+  """
+  TYPE_STRING = 'STRING'
```
arostamianfar: Please use `TYPE_STRING = TableFieldConstants.TYPE_STRING` etc here.

PR #374, opened 2018-10-04
remove date info from variant

allieychen:
Follow up for [PR 373](https://github.com/googlegenomics/gcp-variant-transforms/pull/373), update the code to support the public genomic tables:

- remove the unsupported columns (e.g., Date column) when forming the query.

### DIFF
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -72,7 +72,7 @@ from gcp_variant_transforms.transforms import combine_call_names
 from gcp_variant_transforms.transforms import densify_variants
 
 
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/bq_to_vcf.py
```diff
@@ -162,13 +162,14 @@ def _bigquery_to_vcf_shards(
   Also, it writes the meta info and data header with the call names to
   `vcf_header_file_path`.
   """
+  schema = _get_schema(known_args.input_table)
+  columns = _get_query_columns(schema)
```
arostamianfar: nit: I think it's better to pass the `schema` to `_get_bigquery_query` and get the columns inside that method (unless you're planning to reuse columns somewhere else later). The main reason is to make the API of `_get_bigquery_query` more generic and also limiting the scope of `columns` to the only method that actually needs it.

PR #381, opened 2018-10-16
retry on failure

allieychen:
When there are a large number of blobs, the composing may fail. Retry when this happens. 

Tested: manually ran BQ to VCF.

### DIFF
--- a/gcp_variant_transforms/libs/vcf_file_composer.py
+++ b/gcp_variant_transforms/libs/vcf_file_composer.py
@@ -14,6 +14,7 @@
 
 """Composes multiple files in GCS to one VCF file."""
 
+import logging
 import multiprocessing
 from typing import Iterable, List  # pylint: disable=unused-import
 
@@ -24,6 +25,8 @@ from google.cloud import storage
 
 # Cloud Storage allows to compose up to 32 objects.
 _MAX_NUM_OF_BLOBS_PER_COMPOSE = 32
+_NUMBER_OF_API_CALL_RETRIES = 5
+_TIMEOUT = 30
 
 
 def compose_gcs_vcf_shards(project,  # type: str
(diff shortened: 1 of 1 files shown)


gcp_variant_transforms/libs/vcf_file_composer.py
```diff
@@ -175,20 +178,45 @@ def _compose_blobs_to_one(self, blob_prefix):
...
+      arguments = failed_composing_arguments
+      retry += 1
+      if arguments:
+        if retry > _NUMBER_OF_API_CALL_RETRIES:
+          raise RuntimeError('Composing of blobs fails after {} '
+                             'retries.'.format(_NUMBER_OF_API_CALL_RETRIES))
+        else:
+          logging.warning(
+              '%d jobs of composing of blobs failed due to timeout. Retry for '
+              'the %d time.', len(arguments), retry)
```
arostamianfar: nit: s/Retry for ... / Retrying %d of %d times. ... so it becomes, "Retrying 1 of 3 times."

gcp_variant_transforms/libs/vcf_file_composer.py
```diff
@@ -175,20 +178,47 @@ def _compose_blobs_to_one(self, blob_prefix):
...
+        num_retries += 1
+        blobs_to_compose_args = failed_blobs_to_compose_args
+        logging.warning(
+            '%d jobs of composing of blobs failed due to timeout. Retrying for '
+            '%d of %d time.', len(blobs_to_compose_args), num_retries,
+            _MAX_NUM_OF_COMPOSE_RETRIES)
+      else:
+        break
+    else:
+      raise RuntimeError('Composing of blobs fails after {} '
```
arostamianfar: nit: s/fails/failed

gcp_variant_transforms/libs/vcf_file_composer.py
```diff
@@ -175,20 +178,47 @@ def _compose_blobs_to_one(self, blob_prefix):
...
+        except multiprocessing.TimeoutError:
+          logging.warning('Aborting the composing of blobs (%s to %s) due to '
+                          'timeout.', argument[2][0], argument[2][-1])
+          failed_blobs_to_compose_args.append(argument)
+
+      if failed_blobs_to_compose_args:
+        num_retries += 1
+        blobs_to_compose_args = failed_blobs_to_compose_args
+        logging.warning(
+            '%d jobs of composing of blobs failed due to timeout. Retrying for '
```
arostamianfar: nit: just "Retrying %d of %d." is cleaner (I think it should be "times" as "time" has a different meaning, but we can just remove it :) ).

PR #383, opened 2018-10-16
Remove the new lines in description

allieychen:
In BigQuery public datasets, there are some field descriptions contain `\n` (For instance, in `call.quality` of `bigquery-public-data:human_genome_variants.platinum_genomes_deepvariant_variants_20180823` ).  If adding those descriptions with `\n` into the VCF header, loading it back to BQ will raise an error. This PR:
- replacing all `\n` to ` `.

Tested: manually tested bq to vcf.

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_vcf_schema_converter.py
+++ b/gcp_variant_transforms/libs/bigquery_vcf_schema_converter.py
@@ -227,14 +227,15 @@ def _add_format_fields(schema, formats, allow_incompatible_schema=False):
           id=field.name,
           num=reserved_definition.num,
           type=reserved_definition.type,
-          desc=field.description or reserved_definition.desc)})
+          desc=(_remove_new_line_character(field.description) or
+                reserved_definition.desc))})
     else:
       formats.update({field.name: _Format(
           id=field.name,
           num=bigquery_util.get_vcf_num_from_bigquery_schema(field.mode,
                                                              field.type),
           type=bigquery_util.get_vcf_type_from_bigquery_type(field.type),
-          desc=field.description)})
+          desc=_remove_new_line_character(field.description))})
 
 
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/libs/bigquery_vcf_schema_converter.py
```diff
@@ -227,14 +227,15 @@ def _add_format_fields(schema, formats, allow_incompatible_schema=False):
           id=field.name,
           num=reserved_definition.num,
           type=reserved_definition.type,
-          desc=field.description or reserved_definition.desc)})
+          desc=(_remove_new_line_character(field.description) or
```
arostamianfar: Actually, you can just call `.strip()` method, which removes any whitespace character (including `\n`, `\t` etc).

gcp_variant_transforms/libs/bigquery_vcf_schema_converter.py
```diff
@@ -333,3 +336,7 @@ def _validate_reserved_field_mode(field_schema, reserved_definition):
     raise ValueError(
         'The mode of field {} is different from the VCF spec: {} vs {}.'
         .format(field_schema.name, schema_mode, reserved_mode))
+
+
+def _remove_new_line_character(description):
```
arostamianfar: s/_remove_new_line_character/_remove_special_characters
and also remove `"`.

gcp_variant_transforms/libs/bigquery_vcf_schema_converter.py
arostamianfar: nit: it's a bit cleaner to just do `desc=_remove_special_characters(field.description or reserved_definition.desc)`...we know that our reserved defs dont include special characters so it's a no-op.

allieychen:
Thank you, Asha! PTAL :)

PR #392, opened 2018-10-18
Add prefix to integreation tests of cloudbuild_CI

samanvp:
In order to be able to filter the job names in Stackdriver, we add a
specific prefix to all integration tests run by cloudbuild_CI.
This way we will be able to distinguish between our manual runs of integration tests versus cloudbuild_CI runs.


gcp_variant_transforms/testing/integration/run_vcf_to_bq_tests.py
```diff
@@ -288,6 +293,8 @@ def main():
     for test_case_configs in test_configs:
       test_cases = []
       for config in test_case_configs:
+        if args.test_name_prefix != '':
```
arostamianfar: nit: just `if args.test_name_prefix` is more pythonic :)

samanvp: done.

PR #396, opened 2018-10-22
Update integration tests to use piplines v2alpha1.

### LINKED ISSUE #296: Update integration tests to use Pipelines v2alpha1 API
When issue #222 is fixed, our documentation should all refer to the Pipelines v2alpha1 API, but our integration tests are still running the v1 API. We should update that to use v2 as well.

arostamianfar:
Tested:
Manual run of `./deploy_and_run_tests.sh --run_preprocessor_tests --run_bq_to_vcf_tests --run_presubmit_tests` both with success and failure.

Fixes #296

### DIFF
--- a/gcp_variant_transforms/testing/integration/run_bq_to_vcf_tests.py
+++ b/gcp_variant_transforms/testing/integration/run_bq_to_vcf_tests.py
@@ -80,7 +80,8 @@ class BqToVcfTestCase(run_tests_common.TestCaseInterface):
 
     self.pipelines_api_request = run_tests_common.form_pipelines_api_request(
         parsed_args.project,
-        '/'.join([parsed_args.logging_location, self._output_file]),
+        filesystems.FileSystems.join(parsed_args.logging_location,
+                                     '_'.join([test_name, timestamp])),
         parsed_args.image, _PIPELINE_NAME, _SCRIPT_PATH, zones, args)
 
   def validate_result(self):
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/testing/integration/run_bq_to_vcf_tests.py
```diff
@@ -79,9 +78,10 @@ def __init__(self,
     for k, v in kwargs.iteritems():
       args.append('--{} {}'.format(k, v))
 
-    self.pipeline_api_request = run_tests_common.form_pipeline_api_request(
-        parsed_args.project, parsed_args.logging_location, parsed_args.image,
-        _SCOPES, _PIPELINE_NAME, _SCRIPT_PATH, zones, args)
+    self.pipelines_api_request = run_tests_common.form_pipelines_api_request(
+        parsed_args.project,
+        '/'.join([parsed_args.logging_location, self._output_file]),
```
allieychen: Why we are changing `logging_location`? Will it writes the logging to output_file?

arostamianfar: So, logging_location used to just specify the folder for logging. And then the v1 API created some 'random' files under that folder (corresponded to operation IDs). With the v2 API, we actually have to specify the full log path, so in this case, it will still be under the same folder, but just with the unique file name of the output file or BQ table.

allieychen: I see, thanks for clarifying. It may come strange to join two paths since `self._output_file` also has `gs://../temp`. It is better to use the name with the time stamp.

arostamianfar: Great point! I updated to use timestamp and test_name (that should be unique). Also updated it to use filesystems.join.

PR #397, opened 2018-10-22
Validate input_pattern.

### LINKED ISSUE #82: Add better argument validation

arostamianfar:
Tested:
- Unit tests
- Manual verification with a nonexistent path, existing path, and path without access

Fixes #82.

### DIFF
--- a/gcp_variant_transforms/options/variant_transform_options.py
+++ b/gcp_variant_transforms/options/variant_transform_options.py
@@ -17,6 +17,8 @@ from __future__ import absolute_import
 import argparse  # pylint: disable=unused-import
 import re
 
+from apache_beam.io import filesystem
+from apache_beam.io import filesystems
 from apache_beam.io.gcp.internal.clients import bigquery
 from apitools.base.py import exceptions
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -111,6 +113,16 @@ def validate(self, parsed_args):
       raise ValueError('Both --infer_headers and --representative_header_file '
                        'are passed! Please double check and choose at most one '
                        'of them.')
+    try:
+      # Gets at most one pattern match result of type `filesystems.MatchResult`.
+      first_match = filesystems.FileSystems.match(
```
bashir2: nit: This functionality also exists in `vep_runner` module, please consider removing that as it is not needed anymore.

PR #406, opened 2018-10-31
Add an option to allow records with malformed Number=A fields.

### LINKED ISSUE #129: option to allow number=A fields with too few values
we currently have a couple `number=A` INFO fields in our VCFs that often only have one value. right now, VT's validation is strict and doesn't allow these. it's right, of course, and we should either fix them or remove them from our VCFs...

...but in the meantime, i've been using a small hack to allow these, and put the value in just the first allele (again, incorrect, i know :P): https://github.com/snarfed/gcp-variant-transforms/commit/1aa4f4fcf0ad7869410f0c7aa265133ffbecfb17 . it's from before parts of `bigquery_vcf_schema.py` were refactored into `processed_variant.py` (cc @bashir2).

would you all consider maybe allowing this, or doing something similar, with a flag? maybe with `--allow_malformed_records`? i'm open to alternatives.

arostamianfar:
This option gracefully handles both excess and insufficient values. For now, the flag is set via `allow_malformed_records` in order to avoid having too many flags. We can consider adding a separate flag if users want to control this behavior separate from other malformed conditions.

Tested:
unit + integration tests

Fixes #129

### DIFF
--- a/gcp_variant_transforms/libs/processed_variant.py
+++ b/gcp_variant_transforms/libs/processed_variant.py
@@ -54,6 +54,7 @@ VCF_TYPE_TO_PY = {vcf_header_io.VcfHeaderFieldTypeConstants.STRING: str,
 # Counter names
 class _CounterEnum(enum.Enum):
   VARIANT = 'variant_counter'
+  ALTERNATE_ALLELE_INFO_MISMATCH = 'alternate_allele_info_mismatch_counter'
   ANNOTATION_ALT_MATCH = 'annotation_alt_match_counter'
   ANNOTATION_ALT_MINIMAL_AMBIGUOUS = 'annotation_alt_minimal_ambiguous_counter'
   ANNOTATION_ALT_MISMATCH = 'annotation_alt_mismatch_counter'
@@ -200,7 +201,8 @@ class ProcessedVariantFactory(object):
       use_allele_num=False,  # type: bool
       minimal_match=False,  # type: bool
       infer_annotation_types=False,  # type: bool
-      counter_factory=None  # type: metrics_util.CounterFactoryInterface
+      counter_factory=None,  # type: metrics_util.CounterFactoryInterface,
+      allow_alternate_allele_info_mismatch=False  # type: bool
   ):
     # type: (...) -> None
     """Sets the internal state of the factory class.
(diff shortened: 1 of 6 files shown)


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -224,7 +224,8 @@ def run(argv=None):
       known_args.use_allele_num,
       known_args.minimal_vep_alt_matching,
       known_args.infer_annotation_types,
-      counter_factory)
+      counter_factory,
```
allieychen: It may not related with this PR, but is there any reason why not create the counter inside `ProcessedVariantFactory`?

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -209,24 +211,36 @@ def __init__(
...
     self._split_alternate_allele_info_fields = (
         split_alternate_allele_info_fields)
+    self._allow_alternate_allele_info_mismatch = (
+        allow_alternate_allele_info_mismatch)
     self._annotation_field_set = set(annotation_fields or [])
     cfactory = counter_factory or metrics_util.NoOpCounterFactory()
     self._variant_counter = cfactory.create_counter(
         _CounterEnum.VARIANT.value)
+    self._alternate_allele_info_mismatche_counter = cfactory.create_counter(
+        _CounterEnum.ALTERNATE_ALLELE_INFO_MISMATCH)
```
allieychen: nit: s/ALTERNATE_ALLELE_INFO_MISMATCH/ALTERNATE_ALLELE_INFO_MISMATCH.value, as the above counter `_variant_counter` used `VARIANT.value`

gcp_variant_transforms/libs/processed_variant.py
```diff
@@ -200,7 +201,8 @@ def __init__(
       use_allele_num=False,  # type: bool
       minimal_match=False,  # type: bool
       infer_annotation_types=False,  # type: bool
-      counter_factory=None  # type: metrics_util.CounterFactoryInterface
+      counter_factory=None,  # type: metrics_util.CounterFactoryInterface,
```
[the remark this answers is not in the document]
allieychen: nit: I prefer to switch the order of `counter_factory` and `allow_alternate_allele_info_mismatch`, so that all bool flags can stay together to improve the readability.

gcp_variant_transforms/libs/processed_variant.py
arostamianfar: Nice catch! Done.

gcp_variant_transforms/libs/processed_variant.py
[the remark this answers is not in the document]
arostamianfar: Great suggestion! I actually moved it all the way up to be close to `split_alternate_allele_info_fields` as they're related.

gcp_variant_transforms/vcf_to_bq.py
arostamianfar: Good question. The main reason is to share the factory across all classes (currently only one classes uses counters, but we have some open issues to add counters to other classes too) and also that we want to override it in tests to use the `NoOpCounterFactory` (or any other custom metric factory).

allieychen: Good point! Thanks!

Issue #291, opened 2018-07-03
Remove the constrains (<=2.4) on beam version


allieychen:
The latest Apache Beam (2.5) has some issues on directRunner, which fails our integration tests. A temporary fix was made through  [PR 284](https://github.com/googlegenomics/gcp-variant-transforms/pull/284/files). Once they release the next version, revert the PR and make sure the build works fine.

arostamianfar:
Beam 2.6.0 is released! yay! Allie, could you please run our integration tests with 2.6.0 and see whether we can remove the constraint on 2.4.0?

allieychen:
The pipeline still fails after updating to 2.6.0. It shows the same error message as in [BEAM 4755](https://issues.apache.org/jira/browse/BEAM-4755?jql=project%20%3D%20BEAM%20AND%20text%20~%20%22Python%20cast%22).

pgrosu:
@allieychen Don't give up so quicky - it's easier than you think.  In `Receiver` what do you see the issue to be - here's the link and below is the code:

https://github.com/apache/beam/blob/master/sdks/python/apache_beam/runners/common.py#L121-L131

```Python
class Receiver(object):
  """For internal use only; no backwards-compatibility guarantees.
  An object that consumes a WindowedValue.
  This class can be efficiently used to pass values between the
  sdk and worker harnesses.
  """

  def receive(self, windowed_value):
    raise NotImplementedError
```

Basically what do you see missing that you would usually like to have when creating a class in Python?  The `TypeError: object() takes no parameters` is a big hint :)

~p

allieychen:
It seems that [BEAM 4755](https://issues.apache.org/jira/browse/BEAM-4755) has been resolved recently. I think we can just wait till the next version is released.

allieychen:
The issue is filed in [BEAM 5927](https://issues.apache.org/jira/browse/BEAM-5927).

allieychen:
Track the issue in [BEAM 6404](https://issues.apache.org/jira/browse/BEAM-6404).

PR #407, opened 2018-10-31
small fixes

allieychen:
- Use test name in integration tests error message instead of the operation name to be more informative.
- When the data is small we always run pipeline locally, unless `--infer_headers` is provided. Add `--infer_annotation_types` as well since it also needs to read the variants. (The `valid_4_2_VEP.json` test case now runs correctly when using `BundleBasedDirectRunner` with a higher version of apache beam).

Related issues: https://github.com/googlegenomics/gcp-variant-transforms/issues/291

### DIFF
--- a/gcp_variant_transforms/testing/integration/run_tests_common.py
+++ b/gcp_variant_transforms/testing/integration/run_tests_common.py
@@ -70,6 +70,7 @@ class TestRunner(object):
         credentials=GoogleCredentials.get_application_default())
     self._revalidate = revalidate
     self._operation_names_to_test_states = {}  # type: Dict[str, TestCaseState]
+    self._operation_names_to_test_case_names = {}  # type: Dict[str, str]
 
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/testing/integration/run_tests_common.py
```diff
@@ -113,7 +116,9 @@ def _wait_for_all_operations_done(self):
         request = operations.get(name=operation_name)
         response = request.execute()
         if response['done']:
-          self._handle_failure(operation_name, response)
+          self._handle_failure(
+              self._operation_names_to_test_case_names[operation_name],
```
arostamianfar: I think it's important to have _both_ the operation name and the test case name. The operation name is the main way in which we can get the logs and details of the run.

allieychen:
Thanks Asha! PTAL.

PR #408, opened 2018-11-02
move common funs to common file

allieychen:
Moved some common functions into a common file.
- generate a unique job name/file name.
- check whether a pipeline is using direct runner.

Tested: integration tests.

### DIFF
--- a/gcp_variant_transforms/bq_to_vcf.py
+++ b/gcp_variant_transforms/bq_to_vcf.py
@@ -45,8 +45,6 @@ from __future__ import division
 import logging
 import sys
 import tempfile
-import uuid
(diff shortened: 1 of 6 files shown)


arostamianfar:
Thanks, Allie! This is great. What do you think of renaming `vcf_to_bq_common.py` to something more generic like `pipeline_common.py`? (lemme know if you have better ideas about the name!) the main reason is that it's no longer just about `vcf_to_bq` and can be more about `bq_to_bq`, `vcf_to_vcf` etc.

allieychen:
Thanks Asha! I agree we should change the name. PTAL :)

Issue #229, opened 2018-05-04
Add ability to run a single integration test


arostamianfar:
Currently, the only options are: run small tests, run presubmit tests, or run all tests. Would be nice to add the ability to run a particular test when we want to debug one failing test.

bashir2:
+1 to the feature. In the meantime, if anyone needs to do this, here is [the line to edit](https://github.com/googlegenomics/gcp-variant-transforms/blob/3dad134166b099a8606d93e966c5ebf77daec16d/gcp_variant_transforms/testing/integration/run_tests.py#L314) and limit the files to the ones you want. Please also note the [--keep_tables](https://github.com/googlegenomics/gcp-variant-transforms/blob/3dad134166b099a8606d93e966c5ebf77daec16d/gcp_variant_transforms/testing/integration/run_tests.py#L294) and [--revalidation_dataset_id](https://github.com/googlegenomics/gcp-variant-transforms/blob/3dad134166b099a8606d93e966c5ebf77daec16d/gcp_variant_transforms/testing/integration/run_tests.py#L297) options which can speed up testing if you are changing queries only.

PR #409, opened 2018-11-02
support running one integration test

allieychen:
Add one flag `--test_file_suffix` in vcf_to_bq_tests. 
- Only the test files that end with the suffix will run.
- It provides the ability to run only one test.

Issues: https://github.com/googlegenomics/gcp-variant-transforms/issues/229
Tested: manually ran one test and integration tests.


arostamianfar:
I think it's ok to submit this PR as is. The better long term solution is to create a single binary and run all tests through that (rather than having 3 binaries).

Issue #332, opened 2018-08-15
Integration CloudBuild into Github


arostamianfar:
CloudBuild can now be natively integrated into GitHub: https://cloud.google.com/cloud-build/docs/run-builds-on-github

We can take advantage of this and run small integration tests as part of the presubmit, which would be pretty cool! I think we can also remove travis from our dependencies.

allieychen:
The automatic running of CloudBuild is not supported for the forked repository yet, but it will be ready for dogfood soon.

PR #410, opened 2018-11-05
cloud build integration

allieychen:
Update `cloudbuild.yaml`. It will be used by Google Cloud Build GitHub app, which automatically builds the code each time you push code to Github. This is what it does:
- Builds an image using the GitHub commit hash as the tag.
- Pushes that image to Container Registry.
- Runs small integration tests (only one test included for now) from that very image.

issues: https://github.com/googlegenomics/gcp-variant-transforms/issues/332

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -16,8 +16,7 @@
 # code each time you push code to Github. This is what it does:
 # * Builds an image using the GitHub commit hash as the tag.
 # * Pushes that image to Container Registry.
-# * Runs small integration tests (only one test included for now) from that very
-#   image.
+# * Runs small integration tests (only one test included for now) from that image.
 steps:
   - name: 'gcr.io/cloud-builders/docker'
     args:
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -18,7 +18,6 @@
 # * Pushes that image to Container Registry.
 # * Runs small integration tests (only one test included for now) from that very
 #   image.
-
 steps:
   - name: 'gcr.io/cloud-builders/docker'
     args:
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -12,17 +12,45 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Builds docker image for gcp-variant-transforms:
-# Run using:
-# $ gcloud container builds submit --config cloudbuild.yaml --timeout 1h .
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -12,17 +12,43 @@
...
+# code each time you push code to Github. This is what it does:
+# * Builds an image using the GitHub commit hash as the tag.
+# * Pushes that image to Container Registry.
+# * Runs small integration tests (only one test included for now) from that image.
 steps:
   - name: 'gcr.io/cloud-builders/docker'
     args:
       - 'build'
-      - '--tag=gcr.io/$PROJECT_ID/gcp-variant-transforms:${_CUSTOM_TAG_NAME}'
+      - '--tag=gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
```
kemp-google: How does PROJECT_ID get set here?

cloudbuild.yaml
```diff
@@ -12,17 +12,43 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Builds docker image for gcp-variant-transforms:
-# Run using:
-# $ gcloud container builds submit --config cloudbuild.yaml --timeout 1h .
-substitutions:
-  _CUSTOM_TAG_NAME: 'latest'
+# This is used by Google Cloud Build GitHub app, which automatically builds the
+# code each time you push code to Github. This is what it does:
```
kemp-google: This is actually post commit (eg, after pushed to master) or is it when a PR is created?

allieychen: It is supposed to work when we create a PR, but the automatic running of CloudBuild is not supported for the forked repository yet.

cloudbuild.yaml
allieychen: I think Cloud Build provides some [default substitutions](https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values), and `PROJECT_ID` is one of them. During the [setup](https://cloud.google.com/cloud-build/docs/run-builds-on-github), it will connect Google Cloud Build GitHub app to one GCP project.

allieychen:
Thank you Aaron for the comments. I think I should close this PR since the automatic running of CloudBuild is not supported for the forked repository yet. It does not make sense to run another post commit testing since we already have one. We can revisit it later when forked repository is supported.

PR #416, opened 2018-11-22
Adding google-nucleus to the REQUIRED_PACKAGES list

samanvp:
In PR #346 we added a temporary solution to install Nucleus using its
wheel file. Nucleus now has pip install so we can remove that hack.

Also when running integration tests I was prompted to replace `gcloud container builds submit` with `gcloud builds submit`. My first attempt to run the integration tests after that change failed due to build timeout, so I added the `--timeout "30"` flag.

### DIFF
--- a/deploy_and_run_tests.sh
+++ b/deploy_and_run_tests.sh
@@ -207,8 +207,9 @@ if [[ -z "${skip_build}" ]]; then
   # TODO(bashir2): This will pick and include all directories in the image,
   # including local build and library dirs that do not need to be included.
   # Update this to include only the required files/directories.
-  gcloud container builds submit --config "${build_file}" \
+  gcloud builds submit --config "${build_file}" \
       --project "${project}" \
+      --timeout '30m' \
       --substitutions _CUSTOM_TAG_NAME="${image_tag}" .
 fi
 
--- a/gcp_variant_transforms/beam_io/vcfio_test.py
+++ b/gcp_variant_transforms/beam_io/vcfio_test.py
@@ -101,11 +101,14 @@ def _get_sample_variant_1(is_for_nucleus=False):
         vcfio.VariantCall(name='Sample2', genotype=[1, 0], info={'GQ': 20}))
   else:
     # 0.1 -> 0.25 float precision loss due to binary floating point conversion.
-    vcf_line = ('20	1234	rs123;rs2	C	A,T	50	'
+    # rs123;rs2 -> rs123 it seems nuclues does not parse IDs correctly.
+    # quality=50 -> 50.0 nucleus converts quality values to float.
+    # TODO(samanvp): convert all quality values to float.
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -28,6 +26,7 @@
     'google-api-python-client>=1.6',
     'intervaltree>=2.1.0,<2.2.0',
     'pyvcf<0.7.0',
+    'google-nucleus>=0.2.0',
```
arostamianfar: For now, let's actually pin this to `0.2.0` as they may release a new version, which may break our unit tests.

gcp_variant_transforms/beam_io/vcfio_test.py
```diff
@@ -101,11 +101,14 @@ def _get_sample_variant_1(is_for_nucleus=False):
         vcfio.VariantCall(name='Sample2', genotype=[1, 0], info={'GQ': 20}))
   else:
     # 0.1 -> 0.25 float precision loss due to binary floating point conversion.
-    vcf_line = ('20	1234	rs123;rs2	C	A,T	50	'
+    # rs123;rs2 -> rs123 it seems nuclues does not parse IDs correctly.
+    # quality=50 -> 50.0 nucleus converts quality values to float.
+    # TODO(samanvp): convert all quality values to float.
```
arostamianfar: @allieychen FYI
I actually noticed that we have the same problem for the BQ -> VCF pipeline. The VCF quality values are typically integers, but the spec has float. So, while technically correct, our output VCF files have `#.0` for the quality field.
I think a better solution is to add some logic to our parser to output these as ints if the value is actually an integer (an easy way to do this is: `return quality if int(quality) != quality else int(quality)` (with some special casing when quality is None). Let's do this in another PR though.

deploy_and_run_tests.sh
```diff
@@ -207,8 +207,9 @@ if [[ -z "${skip_build}" ]]; then
   # TODO(bashir2): This will pick and include all directories in the image,
   # including local build and library dirs that do not need to be included.
   # Update this to include only the required files/directories.
-  gcloud container builds submit --config "${build_file}" \
+  gcloud builds submit --config "${build_file}" \
```
arostamianfar: to clarify: is this the new way of using cloud build?

deploy_and_run_tests.sh
```diff
@@ -207,8 +207,9 @@ if [[ -z "${skip_build}" ]]; then
...
       --project "${project}" \
+      --timeout '30m' \
```
arostamianfar: How much does adding nucleus increase the build time?

gcp_variant_transforms/beam_io/vcfio_test.py
allieychen: Got it! Thanks!

PR #417, opened 2018-11-23
Update BQ row limit to 100MB and make it more efficient.

arostamianfar:
BQ row limit has changed from 10MB to 100MB. I also made this method more efficient as it turns out jsonifying every single call/variant is a very expensive operation! This PTransform is now **50% faster!** The sampling logic can be improved (e.g. randomizing the samples rather than picking particular locations, or dynamically choosing sample size based on input), but I realistically don't think anyone is going to hit the 100MB limit anytime soon and this method can be good enough even then.

Tested:
- unit + integration tests
- Also ran on a synthetic dataset with 500k+ files to force row split. Can consider creating this dataset as part of `huge_tests` later.

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_vcf_data_converter.py
+++ b/gcp_variant_transforms/libs/bigquery_vcf_data_converter.py
@@ -15,6 +15,7 @@
 """Handles the conversion between BigQuery row and variant."""
 
 from __future__ import absolute_import
+from __future__ import division
 
 import copy
 import json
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/libs/bigquery_vcf_data_converter.py
```diff
@@ -56,6 +49,19 @@
...
 ]
 
+# Constants for calculating row size. This is needed because the maximum size of
+# a BigQuery row is 100MB. See
+# https://cloud.google.com/bigquery/quotas#import for details.
+# We set it to 90MB to leave some room for error as our row estimate is based
+# on sampling rather than exact byte size.
+_MAX_BIGQUERY_ROW_SIZE_BYTES = 90 * 1024 * 1024
+# Maximum number of calls to sample for BigQuery row size estimate.
+_MAX_NUM_CALL_SAMPLES = 5
```
allieychen: If I understand correctly, this is actually the min number of calls (we always sample 5 calls or 5+1 calls). Maybe `NUM_CALL_SAMPLES` is good enough.

arostamianfar: Good point! Done!

PR #427, opened 2018-12-06
Minor fix to the test script after gcloud update.

bashir2:
Tested: Ran `deploy_and_run_tests.sh`.


allieychen:
Thank you Bashir! But I think Saman already has a [PR](https://github.com/googlegenomics/gcp-variant-transforms/pull/416) that includes this change. :)

samanvp:
I just submitted #416 
Thank you both

PR #430, opened 2019-01-02
Scalable vep annotation in VT

allieychen:
This PR improves the current annotation pipeline by:
- Add option '--shard_input_files' to shard the input files before annotations.
- Add option '--run_with_garbage_collection' to clean up the VMs in case of failure or cancellation.
- Distribute the annotation workload to VMs more evenly based on the number of variants rather than the number of files.
- Add Integration tests and published a new VEP image.

TODO: update the doc.
Tested: unit tests  & integration tests

### DIFF
--- a/gcp_variant_transforms/libs/annotation/vep/vep_runner.py
+++ b/gcp_variant_transforms/libs/annotation/vep/vep_runner.py
@@ -51,11 +51,11 @@ _GCS_UTIL_IMAGE = 'gcr.io/cloud-genomics-pipelines/io'
 # The expected path of the run_vep.sh script in the docker container.
 _VEP_RUN_SCRIPT = '/opt/variant_effect_predictor/run_vep.sh'
 
-# The expected path of run_vep_in_variant_transforms.sh script in the
-# docker container. The difference between this script and run_vep.sh is that
-# it kills the VEP process in case of failure or cancellation.
+# The expected path of `run_script_with_watchdog.sh` script in the
+# docker container. It runs `run_vep.sh` in background and kills this background
+# process in case of failure or cancellation.
 _VEP_RUN_IN_VT_SCRIPT = (
-    '/opt/variant_effect_predictor/run_vep_in_variant_transforms.sh')
+    '/opt/variant_effect_predictor/run_script_with_watchdog.sh')
 # The local name of the output file and directory for VEP runs.
 _LOCAL_OUTPUT_DIR = '/mnt/vep/output_files'
 _LOCAL_OUTPUT_FILE = _LOCAL_OUTPUT_DIR + '/output.vcf'
@@ -69,13 +69,13 @@ _NUMBER_OF_API_CALL_RETRIES = 5
 
 def create_runner(known_args, pipeline_args, input_pattern, watchdog_file):
   # type: (argparse.Namespace, List[str], str, Optional[str]) -> VepRunner
-  """Creates an instance of VepRunner using the provided args.
+  """Returns an instance of VepRunner using the provided args.
 
   Args:
     known_args: The list of arguments defined in `variant_transform_options`.
     pipeline_args: The list of remaining arguments meant to be used to
       determine resources like number of workers, machine type, etc.
-    input_pattern: The file pattern that is to be annotated.
+    input_pattern: The VCF files to be annotated.
     watchdog_file: The file that will be updated by the Dataflow worker every
       `_POLLING_INTERVAL_SECONDS`. Once the file is found to be stale, the VEP
       process running in the VM will be killed.
@@ -144,7 +144,7 @@ class VepRunner(object):
     self._process_pipeline_args(pipeline_args)
     self._watchdog_file = watchdog_file
     self._running_operation_ids = []  # type: List[str]
-    self._operation_name_to_actions = {}
+    self._operation_name_to_io_infos = {}
     self._operation_name_to_logs = {}
 
   def _make_vep_cache_path(self, vep_cache_path):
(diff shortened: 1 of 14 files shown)


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -286,15 +286,23 @@ def add_arguments(self, parser):
...
         help=('The path on Google Cloud Storage to store annotated outputs. '
               'The output files are VCF and follow the same directory '
-              'structure as input files with a suffix added to them.'))
+              'structure as input files with a suffix added to them. Note that '
+              'this is expected not to exist and is created in the process of '
+              'running VEP pipelines.'))
     parser.add_argument(
         '--' + AnnotationOptions._VEP_IMAGE_FLAG,
-        default='gcr.io/gcp-variant-annotation/vep_91',
+        default='gcr.io/gcp-variant-annotation/vep_91_in_vt',
```
nmousavi: I would remove _in_vt suffix. The fact that it has watchdog logic should not make it specific to VT.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -342,6 +350,16 @@ def add_arguments(self, parser):
               'pass over all variants. Additionally, this flag will resolve '
               'conflicts for all headers as if `allow_incompatible_types` was '
               'true.'))
+    parser.add_argument(
+        '--run_with_garbage_collection',
```
nmousavi: Why not always do it?

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -342,6 +350,16 @@ def add_arguments(self, parser):
...
+        type='bool', default=True, nargs='?', const=True,
+        help=('If set, in case of failure or cancellation, the VMs running '
+              'VEP annotation will be cleaned up automatically.'))
+    parser.add_argument(
+        '--number_of_variants_per_shard',
+        type=int, default=20000,
+        help=('The maximum number of variants written to each shard if '
```
nmousavi: It is not clear why shard refers to here, therefore the description is not very helpful. Maybe explain how this number affect performance, and why someone should change it.

allieychen: Updated.

gcp_variant_transforms/options/variant_transform_options.py
allieychen: Yes we should always do it. I keep it for the first version just in case when something unexpected happens, we can still run previous annotation version correctly (like use the previous image, no sharding at all, etc.). Does it make sense to you?

gcp_variant_transforms/transforms/write_variants_to_shards.py
```diff
@@ -0,0 +1,119 @@
...
+      self._counter = 0
+      self._variant_lines[:] = []
+
+  def _write_variant_lines_to_vcf_shard(self, variant_lines):
+    # type: (List[str]) -> None
+    """Writes variant lines to one VCF shard."""
+    vcf_fixed_columns = ['#CHROM', 'POS', 'ID', 'REF', 'ALT', 'QUAL', 'FILTER',
+                         'INFO', 'FORMAT']
+    str_call_names = [str(call_name) for call_name in self._call_names]
+    vcf_header = str('\t'.join(vcf_fixed_columns + str_call_names))
```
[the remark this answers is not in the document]
nmousavi: in that case, we have to rerun merge_headers, don't we? If not, we should definitely avoid it.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -142,15 +150,80 @@ def _add_inferred_headers(pipeline,  # type: beam.Pipeline
...
+      | 'FlattenHeaders' >> beam.Flatten()
       | 'MergeHeadersFromVcfAndVariants' >> merge_headers.MergeHeaders(
           known_args.split_alternate_allele_info_fields,
           known_args.allow_incompatible_records))
   return merged_header
 
 
-def _merge_headers(known_args, pipeline_args, pipeline_mode):
-  # type: (argparse.Namespace, List[str], int) -> None
+def _shard_vcf_files(known_args, pipeline_args, pipeline_mode):
```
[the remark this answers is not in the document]
nmousavi: ?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -194,21 +271,31 @@ def run(argv=None):
...
-    runner.run_on_all_files()
-    runner.wait_until_done()
-    logging.info('Using VEP processed files: %s', known_args.input_pattern)
+    match_results = filesystems.FileSystems.match(
+        ['{}*'.format(known_args.annotation_output_dir)])
+    if match_results and match_results[0].metadata_list:
+      raise ValueError('Output directory {} already exists.'.format(
+          known_args.annotation_output_dir))
+    known_args.omit_empty_sample_calls = True
+    original_input_pattern = known_args.input_pattern
```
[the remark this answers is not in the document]
nmousavi: It's confusing to assign different value to a flag, specially that the flag value is used in the code. I would prefer having  'intermediate_vcf_files' variable over rewriting the flag.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -182,12 +255,32 @@ def _merge_headers(known_args, pipeline_args, pipeline_mode):
         headers,
         known_args.split_alternate_allele_info_fields,
         known_args.allow_incompatible_records)
+    if original_input_pattern:
```
nmousavi: I found this check confusing. 
By just looking at this, I have no idea why headers are not merged if original_input_pattern is None, nor why original_input_pattern can be even None. I have to go and read the piece of the code that sets `original_input_pattern` in order to understand it. Could you use a different conditional variable, something explaining the situation better, perhaps something indicating annotation step is undergoing.

gcp_variant_transforms/libs/annotation/vep/vep_runner.py
```diff
@@ -288,6 +319,16 @@ def _is_done(self, operation):
       logging.info('Operation %s is done.', operation)
     return is_done
 
+  def _get_error_message(self, operation):
+    request = self._pipeline_service.projects().operations().get(
+        name=operation)
+    try:
+      errors = request.execute(num_retries=_NUMBER_OF_API_CALL_RETRIES)['error']
+      return errors['message']
+    except KeyError:
```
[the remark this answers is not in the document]
allieychen: Confirmed with @anamano .

PR #431, opened 2019-01-14
Extract common logic out of merge_header_definitions.py

tneymanov:
Move `VcfHeaderDefinitions` and `DefinitionsMerger` out of `merge_header_definitions.py` and into `libs/` subfolder.

Tested via running presubmit checks and unit tests locally.

### DIFF
--- a/gcp_variant_transforms/libs/preprocess_reporter.py
+++ b/gcp_variant_transforms/libs/preprocess_reporter.py
@@ -50,9 +50,7 @@ from apache_beam.io import filesystems
 
 from gcp_variant_transforms.beam_io import vcfio  # pylint: disable=unused-import
 from gcp_variant_transforms.beam_io import vcf_header_io
-from gcp_variant_transforms.beam_io.vcf_header_io import VcfParserHeaderKeyConstants
-from gcp_variant_transforms.transforms import merge_header_definitions  # pylint: disable=unused-import
-from gcp_variant_transforms.transforms.merge_header_definitions import Definition  # pylint: disable=unused-import
+from gcp_variant_transforms.libs import vcf_header_definitions_merger  # pylint: disable=unused-import
 
 
 _NO_SOLUTION_MESSAGE = 'Not resolved.'
@@ -75,7 +73,8 @@ class _HeaderLine(object):
 
 
 def generate_report(
-    header_definitions,  # type: merge_header_definitions.VcfHeaderDefinitions
+    # type: vcf_header_definitions_merger.VcfHeaderDefinitions
+    header_definitions,
     file_path,  # type: str
     resolved_headers=None,  # type: vcf_header_io.VcfHeader
     inferred_headers=None,  # type: vcf_header_io.VcfHeader
@@ -92,7 +91,7 @@ def generate_report(
       fields that have conflicting definitions.
     inferred_headers: The `VcfHeader` that contains the inferred header
       definitions of the undefined/mismatched header fields.
-    malformed_records: A list of ``MalformedVcfRecord`` for which VCF parser
+    malformed_records: A list of `MalformedVcfRecord` for which VCF parser
       failed.
   """
   resolved_headers = resolved_headers or vcf_header_io.VcfHeader()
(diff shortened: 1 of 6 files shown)


gcp_variant_transforms/libs/vcf_header_definitions.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2019 Google Inc.  All Rights Reserved.
```
allieychen: This class is a pure data object, and it is only used for `HeaderMerger`. I prefer to still include it in module `definitions_merger.py`. For one thing, the `libs` package usually defines functions of how to do something, it might not be expected to have an object definition here (and I can't find a proper package for it). Secondly, the way we resolve the conflicts is tied to how we define `Definition`, if it changes in the future, `definition_merger` is the only place we need to update. WDYT?

gcp_variant_transforms/transforms/merge_header_definitions.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Google Inc.  All Rights Reserved.
+# Copyright 2019 Google Inc.  All Rights Reserved.
```
allieychen: nit: We usually update the year when we create a new module.

gcp_variant_transforms/libs/vcf_header_definitions.py
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+
+"""VCF file header definitions data object definitions."""
+
+import collections
+from collections import namedtuple
+from typing import Dict, List  # pylint: disable=unused-import
+
+from gcp_variant_transforms.beam_io import vcf_header_io
+from gcp_variant_transforms.beam_io.vcf_header_io import VcfHeader  # pylint: disable=unused-import
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/transforms/merge_header_definitions.py
```diff
@@ -14,127 +14,59 @@
...
-      ):
-    # type: (...) -> None
-    """Updates ``first`` by merging values from ``first`` and ``second``."""
-    for key, definitions_to_files_map in second.iteritems():
-      for definition, file_names in definitions_to_files_map.iteritems():
-        first[key].setdefault(definition, [])
-        first[key][definition].extend(str(s) for s in file_names)
-        first[key][definition] = (
-            first[key][definition][:self._MAX_NUM_FILE_NAMES])
+from gcp_variant_transforms.beam_io import vcf_header_io # pylint: disable=unused-import
```
allieychen: nit: Please leave one extra line between `apache_beam` and `gcp_variant_transforms`.

gcp_variant_transforms/transforms/merge_header_definitions.py
```diff
@@ -14,127 +14,59 @@
...
+
```
allieychen: nit: remove the extra line between the two imports. The general guide is that the imports are grouped in the order:
- standard library
- third party library
- local imports.
between each group we put one blank line.
read more: https://www.python.org/dev/peps/pep-0008/#imports

gcp_variant_transforms/transforms/merge_header_definitions.py
```diff
@@ -14,127 +14,59 @@
...
 
   def create_accumulator(self):
-    return VcfHeaderDefinitions()
+    return vcf_header_definitions_merger.VcfHeaderDefinitions()
 
   def add_input(self, source, to_merge):
-    # type: (VcfHeaderDefinitions, VcfHeader) -> VcfHeaderDefinitions
-    return self.merge_accumulators([source,
-                                    VcfHeaderDefinitions(vcf_header=to_merge)])
+    # type: (vcf_header_definitions_merger.VcfHeaderDefinitions,
```
allieychen: This type checking can not be understood by the third party tools. In general, we can add the type in two ways:
- One line type annotation: like https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/transforms/merge_headers.py#L47
- Multiple lines: example: https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/transforms/merge_headers.py#L39.

We usually go with the first one, if it can fit into one line. If it cannot, we will adopt the second one. If it is still too long, either define an alias or find a shorter (without sacrificing the readability) name.

tneymanov: sooo, I presume that your guys' philosophy is "the least amount of alias as possible", right? I'm complying with it here, but I would personally like to alias VcfHeader as well, for uniformity :/ .

Done.

gcp_variant_transforms/libs/preprocess_reporter.py
```diff
@@ -75,7 +73,8 @@ class _HeaderLine(object):
 
 
 def generate_report(
-    header_definitions,  # type: merge_header_definitions.VcfHeaderDefinitions
+    # type: vcf_header_definitions_merger.VcfHeaderDefinitions
```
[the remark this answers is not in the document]
tneymanov: Great to know!

Done.

PR #434, opened 2019-01-22
update the annotation document

allieychen:
- update the document
- add unit tests

Tested: unit tests

### DIFF
--- a/docs/variant_annotation.md
+++ b/docs/variant_annotation.md
@@ -101,6 +101,18 @@ followed by `_vep_output.vcf`. Note that if this directory already exists, then
 Variant Transforms fails. This is to prevent unintentional overwriting of old
 annotated VCFs.
 
+* [`--shard_variants`](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/options/variant_transform_options.py#L290)
+by default, the input files are sharded into smaller temporary VCF files before
+running VEP annotation. If the input files are small, i.e., each VCF file
+contains less than 50,000 variants, set this flag to false can be more time
+efficient.
+
+* [`--number_of_variants_per_shard`](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/options/variant_transform_options.py#L360)
+the maximum number of variants written to each shard if `shard_variants` is
+true. The default value should work for most cases. You may change this flag to
+a smaller value if you have a dataset with a lot of samples. Notice that it may
+take a longer time to run with a smaller value.
+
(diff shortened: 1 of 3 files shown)


docs/variant_annotation.md
```diff
@@ -101,6 +101,18 @@ followed by `_vep_output.vcf`. Note that if this directory already exists, then
 Variant Transforms fails. This is to prevent unintentional overwriting of old
 annotated VCFs.
 
+* [`--shard_variants`](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/options/variant_transform_options.py#L290)
+by default, the input files are sharded into smaller temporary VCF files before
+running VEP annotation. If the input files are small, i.e., each VCF file
+contains less than 50,000 variants, set this flag to false can be more time
```
nmousavi: nit: s/set/setting

I like the following slightly better, but leave it to you

", setting this flag can be computationally wasteful."

docs/variant_annotation.md
```diff
@@ -101,6 +101,18 @@ followed by `_vep_output.vcf`. Note that if this directory already exists, then
...
+efficient.
+
+* [`--number_of_variants_per_shard`](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/options/variant_transform_options.py#L360)
+the maximum number of variants written to each shard if `shard_variants` is
+true. The default value should work for most cases. You may change this flag to
+a smaller value if you have a dataset with a lot of samples. Notice that it may
+take a longer time to run with a smaller value.
```
nmousavi: S/.../Notice that pipeline may take longer to finish for smaller value of this flag.

docs/variant_annotation.md
```diff
@@ -121,19 +133,16 @@ annotations; use this field to override the field name.
 * [`--vep_num_fork`](https://github.com/googlegenomics/gcp-variant-transforms/blob/c4659bba2cf577d64f15db5cd9f477d9ea2b51b0/gcp_variant_transforms/options/variant_transform_options.py#L208)
 number of forks when running VEP, see [`--fork` option of VEP](ensembl.org/info/docs/tools/vep/script/vep_options.html#opt_fork).
 
+
 For other parameters, like how many VMs to use or where the VMs should be
 located, the same parameters for
 [Dataflow pipeline execution](https://cloud.google.com/dataflow/pipelines/specifying-exec-params)
-are reused, e.g., `--num_workers`.
+are reused, e.g., `--num_workers`. It is recommended to provide at least 100 
+workers for large inputs. Further increasing the number of workers can reduce
```
nmousavi: What is large inputs? If we cannot define it, this comment is not very helpful.

Alternatively we can give a rule of thumb to users, like 1 worker per 50k (or '--number_of_variants_per_shard') variants, can we?

PR #440, opened 2019-02-14
update the watchdog file

allieychen:
Fix one bug for annotation. The operations could get killed because of the watchdog file is stale even though no cancellation is made. The watchdog file was updated only when waiting for one operation to be done. This fix handles the following two failing cases:
- If there are a lot of PAPI calls, some operations might already start running VEP (after loading cache) while we are still submitting other operations, the stale watchdog file will get this operation to be killed.
- If there are a lot of consecutive done operations, the watchdog file won't be updated in time.

Tested: unit tests & integration tests & manually ran one annotation test.

### DIFF
--- a/gcp_variant_transforms/libs/annotation/vep/vep_runner.py
+++ b/gcp_variant_transforms/libs/annotation/vep/vep_runner.py
@@ -15,7 +15,6 @@
 from __future__ import absolute_import
 
 import argparse  # pylint: disable=unused-import
-import datetime
 import logging
 import time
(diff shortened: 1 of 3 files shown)


kemp-google:
@allieychen Can you put the commentary about *why* this change is being made in the commit itself?  It's much better to have information like this in the commit object rather than the PR since the PR is not a permanent part of the record.

allieychen:
Thanks for the review! I have moved updating the watchdog file in another thread. PTAL :)

Issue #170, opened 2018-04-05
Automate release process as much as possible


arostamianfar:
The existing release process requires a few manual steps. These can be easily automated through a bash script.

PR #443, opened 2019-02-26
Automate release process

allieychen:
- Create cloudbuild_release.yaml file to cover the release processes.
- Update the release doc.

Tested: created such trigger in private project and tested the release process.
Issue: [issue 170](https://github.com/googlegenomics/gcp-variant-transforms/issues/170)


cloudbuild_release.yaml
```diff
@@ -0,0 +1,57 @@
...
+# - Pushes that image to Container Registry.
+# - Runs all integration test from that very image.
+# - Pushes the verified image to the main project (gcp-variant-transforms).
+
+steps:
+  - name: 'gcr.io/cloud-builders/docker'
+    args:
+      - 'build'
+      - '--tag=gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
+      - '--file=docker/Dockerfile'
```
kemp-google: args: [ 'build', '-t', 'gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}', 'docker/' ]

you can break it out on lines if you want, but the `-t` and dropping `--file` should help make it more concise

cloudbuild_release.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - '.'
+    id: 'build-gcp-variant-transforms-docker'
+  - name: 'gcr.io/cloud-builders/docker'
```
kemp-google: Can you add a newline between steps?  I find it increases readability significantly

cloudbuild_release.yaml
```diff
@@ -0,0 +1,57 @@
...
+    args:
+      - 'push'
+      - 'gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
+    id: 'push-gcp-variant-transforms-docker'
+  - name: 'gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
+    args:
+      - 'bash'
+      - './deploy_and_run_tests.sh'
+      - '--keep_image'
```
kemp-google: Why keep_image?

cloudbuild_release.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - '--skip_build'
+      - '--project ${PROJECT_ID}'
+      - '--image_tag ${COMMIT_SHA}'
+      - '--run_unit_tests'
+      - '--run_preprocessor_tests'
+      - '--run_bq_to_vcf_tests'
+      - '--run_all_tests'
+      - '--test_name_prefix cloud-ci-'
+    id: 'test-gcp-variant-transforms-docker'
+  - name: 'gcr.io/cloud-builders/docker'
```
kemp-google: You shouldn't need this step: we will be running it in the gcp-variant-transforms project, and if we aren't we shouldn't cross tag

cloudbuild_release.yaml
```diff
@@ -0,0 +1,57 @@
...
+    args:
+      - 'tag'
+      - 'gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
+      - 'gcr.io/gcp-variant-transforms/gcp-variant-transforms:${COMMIT_SHA}'
+    id: 'tag-gcp-variant-transforms-docker'
+images:
```
kemp-google: Nor should you need this AFAIK since you've already pushed

allieychen:
Thanks Aaron! One thing to clarify, what I am thinking is that the build trigger will be added in `gcp-variant-transforms-test`, so all tests are actually running in `gcp-variant-transforms-test`, which I think it would be more convenient since we have the quotas there (also, accessible to some of the test files). If it succeeds, then we push the image to `gcp-variant-transforms`.

cloudbuild_release.yaml
kemp-google: OK you've convinced me to leave this as is :)

Let's change the step ids to something like

build-image
deploy-to-test-project
test-image
deploy-to-serving-project

so it's at least clear the difference?

Issue #447, opened 2019-03-08
[BQ to VCF] Unable to export chromosomes X, Y


mbookman:
I ran a test exported with:

```
--genomic_regions chrY
```

which failed to generate any variants (leading to #446 being filed).

It appears the problem is that `chrY` is turned into `chry`, which appears to be done in `parse_genomic_region`:

https://github.com/googlegenomics/gcp-variant-transforms/blob/82958e83f64e757f951e8bb023a34bdd5a5e6f8a/gcp_variant_transforms/libs/genomic_region_parser.py#L29

Issue #446, opened 2019-03-07
[BQ to VCF] handle failures when there is no shard to compose


allieychen:
The composing step fails (maximum recursion depth exceeded) when there is not shards to compose. Please handle this case properly and raise the error with proper message.

PR #448, opened 2019-03-08
Fix issues in bq to vcf

allieychen:
- Raise an error when there are no shards found.
- When parsing the genomic region filter, it should be case sensitive. For instance, chrY:1-1000, should have ref=chrY, rather than ref=chry, or the SELECT query will return the wrong results. This method is reused by partitioning, in which the matching of the reference should be case insensitive.

Issues: [447](https://github.com/googlegenomics/gcp-variant-transforms/issues/447),[446](https://github.com/googlegenomics/gcp-variant-transforms/issues/446)
Tested: unit tests

### DIFF
--- a/gcp_variant_transforms/libs/genomic_region_parser_test.py
+++ b/gcp_variant_transforms/libs/genomic_region_parser_test.py
@@ -29,8 +29,8 @@ class GenomicRegionParserTest(unittest.TestCase):
         ('chr1', 1000000, 2000000)
     )
     self.assertEqual(
-        genomic_region_parser.parse_genomic_region('chr1:1000000-2000000'),
-        ('chr1', 1000000, 2000000)
+        genomic_region_parser.parse_genomic_region('chrY:1000000-2000000'),
+        ('chrY', 1000000, 2000000)
     )
     self.assertEqual(
         genomic_region_parser.parse_genomic_region('chr'),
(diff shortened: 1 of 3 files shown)


gcp_variant_transforms/libs/vcf_file_composer.py
```diff
@@ -179,6 +179,8 @@ def _compose_blobs_to_one(self, blob_prefix):
     """
     blobs_to_be_composed = list(self._bucket.list_blobs(prefix=blob_prefix))
     logging.info('Total number of blobs is %d.', len(blobs_to_be_composed))
+    if not blobs_to_be_composed:
```
kemp-google: This would have been a good candidate for a separate commit.  This is part of the value of keeping your commit messages complete from the start: you can then rebase and push when you have multiple unrelated changes in a single PR.

gcp_variant_transforms/libs/genomic_region_parser_test.py
```diff
@@ -36,5 +36,9 @@ def test_parse_genomic_regions(self):
         genomic_region_parser.parse_genomic_region('chr'),
         ('chr', 0, genomic_region_parser._DEFAULT_END_POSITION)
     )
+    self.assertEqual(
```
kemp-google: This could be folded into the test above (chr1 -> chrY)

gcp_variant_transforms/libs/vcf_file_composer.py
allieychen: I see! I will drop this PR and create a new one. :)

Issue #453, opened 2019-03-11
SSL: CERTIFICATE_VERIFY_FAILED errors during local pipeline run.


tneymanov:
When running the tool locally, the following error pops up.

<pre>Traceback (most recent call last):
  File "gcp_variant_transforms/testing/integration/run_vcf_to_bq_tests.py", line 317, in <module>
    ret_code = main()
  File "gcp_variant_transforms/testing/integration/run_vcf_to_bq_tests.py", line 301, in main
    with TestContextManager(args) as context:
  File "gcp_variant_transforms/testing/integration/run_vcf_to_bq_tests.py", line 204, in __enter__
    _ = client.create_dataset(dataset)  # See #171, pylint: disable=no-member
  File "/tmp/tmp.ZfVM9FCvpG/local/lib/python2.7/site-packages/google/cloud/bigquery/client.py", line 348, in create_dataset
    api_response = self._call_api(retry, method="POST", path=path, data=data)
  File "/tmp/tmp.ZfVM9FCvpG/local/lib/python2.7/site-packages/google/cloud/bigquery/client.py", line 400, in _call_api
    return call()
  File "/tmp/tmp.ZfVM9FCvpG/local/lib/python2.7/site-packages/google/api_core/retry.py", line 270, in retry_wrapped_func
    on_error=on_error,
  File "/tmp/tmp.ZfVM9FCvpG/local/lib/python2.7/site-packages/google/api_core/retry.py", line 179, in retry_target
    return target()
  File "/tmp/tmp.ZfVM9FCvpG/local/lib/python2.7/site-packages/google/cloud/_http.py", line 299, in api_request
    headers=headers, target_object=_target_object)
  File "/tmp/tmp.ZfVM9FCvpG/local/lib/python2.7/site-packages/google/cloud/_http.py", line 193, in _make_request
    return self._do_request(method, url, headers, data, target_object)
  File "/tmp/tmp.ZfVM9FCvpG/local/lib/python2.7/site-packages/google/cloud/_http.py", line 223, in _do_request
    body=data)
  File "/tmp/tmp.ZfVM9FCvpG/local/lib/python2.7/site-packages/google_auth_httplib2.py", line 187, in request
    self._request, method, uri, request_headers)
  File "/tmp/tmp.ZfVM9FCvpG/local/lib/python2.7/site-packages/google/auth/credentials.py", line 122, in before_request
    self.refresh(request)
[...]

PR #454, opened 2019-03-11
Update deploy_and_run_tests with httplib2 update

tneymanov:
Update deploy_and_run_tests.sh to upgrade httplib2 to newer version.
Workaround for https://github.com/googlegenomics/gcp-variant-transforms/issues/453

### DIFF
--- a/deploy_and_run_tests.sh
+++ b/deploy_and_run_tests.sh
@@ -227,6 +227,11 @@ if [[ -n "${run_unit_tests}" ]]; then
 fi
 pip install --upgrade .[int_test]
 
+# Due to the dependency issue, we have to manually upgrade httplib2 in order to
+# circumvent the CERTIFICATE_VERIFY_FAILED error. For more info, visit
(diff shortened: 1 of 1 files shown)


allieychen:
Please checkout a new branch from the master, so that the commits for other issues won't be mixed together :)

deploy_and_run_tests.sh
```diff
@@ -227,6 +227,11 @@ if [[ -n "${run_unit_tests}" ]]; then
 fi
 pip install --upgrade .[int_test]
 
+# Due to the dependency issue, we have to manually upgrade httplib2 in order to
+# circumvent the CERTIFICATE_VERIFY_FAILED error. For more info, visit
+# https://github.com/googlegenomics/gcp-variant-transforms/issues/453.
+pip install --upgrade httplib2
```
kemp-google: # Force an upgrade to avoid SSL certificate verification errors (issue #453).

I think that will (a) be sufficient and (b) autolink to the issue?

tneymanov: Done.

PR #467, opened 2019-04-03
Update docker file

allieychen:
User can use docker run command to run the tool.
Update the integration tests to be based on the new image.

Tested: integration tests.


cloudbuild_CI.yaml
```diff
@@ -22,6 +22,8 @@ steps:
   - name: 'gcr.io/cloud-builders/docker'
     args:
       - 'build'
+      - '--build-arg'
+      - 'IMAGE_TAG=${COMMIT_SHA}'
```
kemp-google: The downside of this is that even our released versions will not use the release tag, they will still use the commit SHA1.

I'm not sure there is a better option, though.

cloudbuild_CI.yaml
```diff
@@ -34,32 +36,9 @@ steps:
...
-
-  # TODO(yifangchen): add the test for the image above.
-  # Run the test script from the first image we built and pushed.
-  #
-  # Note that ./deploy_and_run_tests.sh that is referenced here is coming from
-  # the source downloaded from GitHub and not the file in the docker image.
-  # The only reason that we use the built image in the 'name' argument is
-  # convenience, i.e., to avoid installing python, virtualenv, etc. on the
-  # image.
+  # Run the test script from the very image we built and pushed.
```
kemp-google: I don't know that this comment adds any value?

cloudbuild_CI.yaml
```diff
@@ -70,6 +49,8 @@ steps:
       - '--run_all_tests'
       - '--test_name_prefix cloud-ci-'
     id: 'test-gcp-variant-transforms-docker'
+    entrypoint: './deploy_and_run_tests.sh'
```
kemp-google: The `./` gives me pause.  I'd usually expect to see either the bare script (to allow exec to do PATH lookup) or a full path.

It's not clear to me why this works (eg, this suggests the working directory inside the container contains the script)

docker/Dockerfile
```diff
@@ -17,6 +17,9 @@
 
 FROM google/cloud-sdk
 
+ARG IMAGE_TAG
+ENV image_tag=${IMAGE_TAG}
```
kemp-google: I think, given that it's the COMMIT_SHA we should call it that.  Otherwise users that run

  `docker run gcr.io/gcp-variant-transforms/gcp-variant-transforms:1.0 echo $image_tag`

will see a commit SHA1 instead of the tag they actually pulled...

gcp_variant_transforms/testing/integration/run_bq_to_vcf_tests.py
```diff
@@ -41,9 +41,8 @@
 from gcp_variant_transforms.testing.integration import run_tests_common
 
 
-_PIPELINE_NAME = 'gcp-variant-transforms-bq-to-vcf-integration-test'
 _TEST_FOLDER = 'gcp_variant_transforms/testing/integration/bq_to_vcf_tests'
-_SCRIPT_PATH = '/opt/gcp_variant_transforms/bin/bq_to_vcf'
+_SCRIPT_PATH = 'bq_to_vcf'
```
kemp-google: _TOOL_NAME?

PR #470, opened 2019-04-08
Update apache beam to the latest version

allieychen:
Remove the constrains on apache beam version.
Update the input pattern and some unit tests since the match method in filesystems has changed in the newer version of beam.

Issues: #291 , #453 
Tested: integration tests.


kemp-google:
Usually `/*` means match files and `/**` means descend into directories too.  Where is this syntax documented for BEAM?

allieychen:
Yes, previously for GCS pattern, beam uses * to match into directories as well. The[ comment](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/filesystems.py#L152) explains how the match works.

PR #472, opened 2019-04-09
Remove the disk

allieychen:
The default disk size is 500, which is too large.

### DIFF
--- a/docker/pipelines_runner.sh
+++ b/docker/pipelines_runner.sh
@@ -83,7 +83,8 @@ function main {
     --zones "${zones}" \
     --image "${vt_docker_image}" \
     --pvm-attempts 0 \
-    --attempts 1
+    --attempts 1 \
+    --disk-size 10
 }
 
 main "$@"


allieychen:
@kemp-google Yeah, fixing the pipelines tool would be better. I checked the pipelines submitted using gcloud command, there is not disk added, so it should be fine. Do you prefer to set it to a smaller value (maybe 10 GB) or 0 as it is now?

kemp-google:
> @kemp-google Yeah, fixing the pipelines tool would be better. I checked the pipelines submitted using gcloud command, there is not disk added, so it should be fine. 

gcloud doesn't move `TMPDIR` I don't think.

> Do you prefer to set it to a smaller value (maybe 10 GB) or 0 as it is now?

If it works with 0, that's fine, but you will need to test it.  It would definitely be safer to just set it to a small value like 10GB until [we fix the pipelines tool](https://github.com/googlegenomics/pipelines-tools/issues/101).

allieychen:
Thanks! I have changed it to 10 GB to be safer.

PR #476, opened 2019-04-17
Revert "Update the doc for running with the new docker image"

allieychen:
Reverts googlegenomics/gcp-variant-transforms#468.


kemp-google:
Why do you want to revert this?

allieychen:
@kemp-google While the latest docker image cannot actually run with this command, I think it is safer to revert until it is fixed.

Issue #475, opened 2019-04-16
Uncrecognized flag "input_pattern" error in 0.7.0 release.


tneymanov:
0.7.0 seems to have introduced a bug for vcf_to_bq and vcf_to_bq_preprocessor pipelines with regards to input_pattern flag usage.

Operation Info: 

- description: |-
      Stopped running "/opt/gcp_variant_transforms/bin/vcf_to_bq_preprocess --project my_project --input_pattern gs://my_bucket/
test.g.vcf --report_path gs://my_bucket/test_preprocessor_report.tsv --resolved_headers_path gs://my_bucket/test_resolved_headers.vcf --report_all_conflicts true --temp_location gs://my_bucket/temp --job_name vcf-to
-bigquery-preprocess-2019-04-15-13-39-38-850 --runner DataflowRunner --region us-central1": exit status 1: getopt: unrecognized option '--input_pattern'
      getopt: unrecognized option '--report_path'
      getopt: unrecognized option '--resolved_headers_path'
      getopt: unrecognized option '--report_all_conflicts'
      getopt: unrecognized option '--job_name'
      getopt: unrecognized option '--runner'
      getopt: unrecognized option '--region'
    details:
      '@type': type.googleapis.com/google.genomics.v2alpha1.ContainerStoppedEvent
      actionId: 4
      exitStatus: 1
      stderr: |
        getopt: unrecognized option '--input_pattern'
        getopt: unrecognized option '--report_path'
        getopt: unrecognized option '--resolved_headers_path'
        getopt: unrecognized option '--report_all_conflicts'
        getopt: unrecognized option '--job_name'
        getopt: unrecognized option '--runner'
        getopt: unrecognized option '--region'
    timestamp: '2019-04-15T18:41:54.155391Z'

- commands:
      - /opt/gcp_variant_transforms/bin/vcf_to_bq_preprocess
      - ...
      entrypoint: ''
      environment: {}
      flags: []
      imageUri: gcr.io/gcp-variant-transforms/gcp-variant-transforms
      labels: {}
      mounts: []
      name: ''
      pidNamespace: ''
      portMappings: {}

tneymanov:
Ran the most primitive runs for both vcf_to_bq and vcf_to_bq_preprocess:

python2 -m gcp_variant_transforms.vcf_to_bq \
  --input_pattern gs://gcp-variant-transforms-testfiles/small_tests/valid-4.0.vcf \
[...]

PR #477, opened 2019-04-18
Modify entry script to execute command as is, if supplied to pipelines API directly.

tneymanov:
Fixes the issue in #475 

Tested via docker, gcloud and pipelines run.
Got the same error as customer with following command on older image, but passed with new image:
pipelines run action_input.json --scopes https://www.googleapis.com/auth/genomics,https://www.googleapis.com/auth/cloud-platform --zones us-west1-b

with action_input.json:
[
  {
    "imageUri": "gcr.io/tural-test-runner/gcp-variant-transforms:a53a8fa653d5110ac3da0d5ed71f1147dde3fd33",
    "commands": ["/opt/gcp_variant_transforms/bin/vcf_to_bq", "--project", "tural-test-runner", "--input_pattern", "gs://gcp-variant-transforms-testfiles/small_tests/valid-4.0.vcf", "--output_table", "tural-test-runner:turalds.test1800", "--runner", "DataflowRunner", "--job_name", "vcf-to-bigquery", "--temp_location", "gs://tural-test-runner/temp", "--staging_location", "gs://tural-test-runner/staging"],
  }
]

### DIFF
--- a/docker/pipelines_runner.sh
+++ b/docker/pipelines_runner.sh
@@ -51,6 +51,9 @@ function parse_args {
 }
 
 function main {
+  if [[ $1 == /opt/gcp_variant_transforms/bin/* ]]; then
+    exec $@
+  fi
   parse_args "$@"
 
   google_cloud_project="${google_cloud_project:-$(gcloud config get-value project)}"


docker/pipelines_runner.sh
```diff
@@ -51,6 +51,10 @@ function parse_args {
 }
 
 function main {
+  if [[ $1 == /opt/gcp_variant_transforms/bin/* ]]; then
+    $@
```
kemp-google: I think you want to `exec` it instead so the return status is correct

tneymanov: Done.

PR #484, opened 2019-05-24
Add support for bgzf.

allieychen:
If the index file is not present, use fusion break to speed up the pipeline.
Otherwise, parse the index file and read the bgzf in parallel.
The BGZF can end with .gz or .bgz, and must be in GCS.
The index file is required to be in the same path as bgzf ending with .tbi.
Only PyVCF Parser is supported.


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -447,6 +461,14 @@ def run(argv=None):
   metrics_util.log_all_counters(result)
 
 
+class FusionBreak(beam.PTransform):
```
tneymanov: Could you add a comment/doc here? This hack wouldn't make sense to a person without context.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -250,6 +254,7 @@ def __init__(self,
                compression_type,  # type: str
```
tneymanov: Silly github won't allow me to comment above this section... - I think we should augment the doc to explain compressed files logic below.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -259,17 +264,40 @@ def __init__(self,
...
+          file_name,
+          range_tracker,
+          representative_header_lines,
+          compression_type,
+          header_processor_fns=(
+              lambda x: not x.strip() or x.startswith('#'),
+              self._process_header_lines),
+          **kwargs)
+    elif compression_type == filesystems.CompressionTypes.GZIP:
+      text_source = BGZFSource(
```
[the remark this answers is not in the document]
allieychen: The build in Text one does not support BGZF, i.e., the reading would fail. Given one file that ends with .gz without index file present, we cannot easily tell whether it is non-blocked gzip or blocked gzip. The BGZFSource extended the build-in class, and it is able to read both BGZF and GZIP file. There is no performance degradation for reading GZIP using BGZFSource.

gcp_variant_transforms/beam_io/bgzf_io.py
```diff
@@ -0,0 +1,182 @@
...
+    last row of Block_i if needed. It assumes one VCF record can at most spans
+    in two Blocks.
+
+  Raises:
+    ValueError: If a block is larger than `MAX_BLOCK_SIZE`.
+  """
+  blocks = _get_block_offsets(_get_tbi_file(file_path))
+  blocks = _remove_invalid_blocks(blocks)
+  blocks = _fill_in_gap(blocks)
+  blocks = _merge_blocks(blocks)
```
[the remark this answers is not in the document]
allieychen: Yes, there are overlapped blocks, which is handled in `_merge_blocks` step.

gcp_variant_transforms/beam_io/bgzf_io.py
```diff
@@ -0,0 +1,176 @@
...
+from apache_beam.io import filesystems
+
+# `Block` provides real file offset of the start and end of a gzip block. Given
+# it, one can directly seek to the start of the gzip block and decompress it.
+Block = namedtuple('Block', ['start', 'end'])
+
+_INDEX_FILE_APPENDIX = '.tbi'
+
+# `MAX_BLOCK_SIZE` defines the maximum amount of data included in one `Block`.
+MAX_BLOCK_SIZE = 1 * 1024 * 1024
```
kemp-google: The comment doesn't tell me anything the name of the constant doesn't already.

Where does this value come from?  Is it a guess on our side or related to a specification value?

gcp_variant_transforms/beam_io/bgzf_io.py
```diff
@@ -0,0 +1,176 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Splits the BGZF file."""
+
+import struct
+from collections import namedtuple
+from typing import Iterable, List, Tuple  # pylint: disable=unused-import
```
kemp-google: Why disable the lint warning here? [excuse the python newbery if there is a normal reason]

gcp_variant_transforms/beam_io/bgzf_io.py
```diff
@@ -0,0 +1,176 @@
...
+
+# `Block` provides real file offset of the start and end of a gzip block. Given
+# it, one can directly seek to the start of the gzip block and decompress it.
+Block = namedtuple('Block', ['start', 'end'])
+
+_INDEX_FILE_APPENDIX = '.tbi'
+
+# `MAX_BLOCK_SIZE` defines the maximum amount of data included in one `Block`.
+MAX_BLOCK_SIZE = 1 * 1024 * 1024
+# It beyonds the normal bin range (1-37449), and contains other meta data rather
```
kemp-google: `# This bin index and any beyond it contain index metadata and are not used to address blocks.`

At least that's my understanding (eg, 37451 is also not expected)

gcp_variant_transforms/beam_io/bgzf_io.py
```diff
@@ -0,0 +1,176 @@
...
+  # type: (str) -> Iterable[Tuple[str, Block]]
+  """Splits BGZF to multiple blocks.
+
+  It reads the `Block` (GZIP block offsets) from the corresponding index file.
+  Given (`file_path`, `Block`), one is able to decompress the data within
+  `Block`. It further:
+  - Removes invalid Blocks.
+  - Fills in the gap (by inserting a new Block) if there are gaps between two
+    adjacent Blocks to make sure it has full coverage of the variant data.
+  - Merges Blocks. It eliminates overlapping Blocks, and merges adjacent Blocks
```
kemp-google: What circumstances lead to any of these conditions?  Invalid blocks and gaps seem unlikely?  Merging makes sense.

What am I missing for the invalid and gap cases?

gcp_variant_transforms/beam_io/bgzf_io.py
```diff
@@ -0,0 +1,176 @@
...
+  for block in blocks:
+    yield (file_path, block)
+
+
+def exists_tbi_file(file_path):
+  return filesystems.FileSystems.exists(_get_tbi_file(file_path))
+
+
+def _get_tbi_file(file_path):
+  return file_path + _INDEX_FILE_APPENDIX
```
kemp-google: For BAMs, we often have .bam.bai _or_ .bai.  Is this not true in this case as well?

gcp_variant_transforms/beam_io/bgzf_io.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def _get_block_offsets(index_file):
+  # type: (str) -> List[Block]
+  """Returns block offsets by parsing the `index_file`.
+
+  The file format can be found in http://samtools.github.io/hts-specs/tabix.pdf.
```
kemp-google: One benifit of pysam is that it can already read tabix indexes.  I wonder how much of this code we can eliminate if we move to pysam?

gcp_variant_transforms/beam_io/bgzf_io.py
```diff
@@ -0,0 +1,176 @@
...
+  """
+  read_buffer = _read(index_file)
+  offset = 4
+  n_ref, offset = _get_next_int(read_buffer, offset)
+  offset = 32
+  l_mn, offset = _get_next_int(read_buffer, offset)
```
kemp-google: What is l_mn?

gcp_variant_transforms/beam_io/bgzf_io.py
```diff
@@ -0,0 +1,176 @@
...
+      if not line:
+        break
+      lines.append(line)
+  return ''.join(lines)
+
+
+def _remove_invalid_blocks(blocks):
+  valid_blocks = []
+  for block in blocks:
+    if block.start != block.end:
```
kemp-google: Why this particular case only?  What about end < start, etc?  I don't think, in general, this is likely, but if you are testing for one it seems reasonable to test for all?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -660,3 +688,181 @@ def _get_variant_calls(self, variant_proto):
         call.info[k] = data
       calls.append(call)
     return calls
+
+
+def open_bgzf(file_name):
```
kemp-google: I would think this function and class definition belong in a separate file but I am not too familiar with pythons rules on this sort of thing...

gcp_variant_transforms/pipeline_common.py
```diff
@@ -101,6 +104,40 @@ def _get_all_patterns(input_pattern, input_file):
...
+
+  Raises:
+    ValueError: if the input files are not in the same format.
+  """
+  matches = filesystems.FileSystems.match(input_patterns)
+  extensions = [os.path.splitext(metadata.path)[1] for match in matches
+                for metadata in match.metadata_list]
+  if len(set(extensions)) != 1:
+    raise ValueError('All input files must be in the same format.')
+  if extensions[0].endswith('.bgz') or extensions[0].endswith('.gz'):
```
kemp-google: Presumably AUTO works for .gz?  Only .bgz we care about overriding?

gcp_variant_transforms/beam_io/bgzf.py
```diff
@@ -0,0 +1,199 @@
...
+
+from apache_beam.coders import coders
+from apache_beam.io import filesystem
+from apache_beam.io import filesystems
+from apache_beam.io import textio
+from apache_beam.io.gcp import gcsio
+
+
+class BGZFSource(textio._TextSource):
+
```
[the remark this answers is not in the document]
allieychen: Done.

gcp_variant_transforms/beam_io/bgzf.py
```diff
@@ -0,0 +1,214 @@
...
+def open_bgzf(file_name):
+  compression_type = filesystems.CompressionTypes.GZIP
+  mime_type = filesystems.CompressionTypes.mime_type(compression_type)
+  raw_file = gcsio.GcsIO().open(file_name, 'rb', mime_type=mime_type)
+  return BGZF(raw_file)
+
+
+class BGZF(filesystem.CompressedFile):
+  """File wrapper for easier handling of BGZF compressed files.
```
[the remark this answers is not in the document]
allieychen: Done.

gcp_variant_transforms/beam_io/bgzf.py
```diff
@@ -0,0 +1,214 @@
...
+      buf = self._file.read(self._read_size)
+      if buf:
+        decompressed = self._decompressor.decompress(buf)
+        del buf
+        self._read_buffer.write(decompressed)
+        self._decompress_decompressor_unused_data(num_bytes)
+      else:
+        self._read_eof = True
+
+  def _decompress_decompressor_unused_data(self, num_bytes):
```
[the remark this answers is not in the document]
allieychen: I have added detailed explanation on how the decompressor works, can you please check again whether the naming makes sense now? Thanks!

gcp_variant_transforms/beam_io/bgzf.py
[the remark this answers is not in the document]
allieychen: Thank you so much! Done.

gcp_variant_transforms/beam_io/bgzf.py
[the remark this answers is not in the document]
allieychen: Done. Nice one!

gcp_variant_transforms/beam_io/bgzf_io.py
kemp-google: So names_len then?

gcp_variant_transforms/pipeline_common.py
```diff
@@ -101,6 +104,40 @@ def _get_all_patterns(input_pattern, input_file):
...
+  # type: (List[str]) -> List[str]
+  """Returns the splittable bgzf matching `all_patterns`."""
+  matches = filesystems.FileSystems.match(all_patterns)
+  files = []
+  for match in matches:
+    for metadata in match.metadata_list:
+      if not (metadata.path.startswith('gs://') and
+              bgzf_io.exists_tbi_file(metadata.path)):
+        logging.info("Cannot find the index file for %s.", metadata.path)
+        return []
```
kemp-google: If the intent is to discard all the previous matches, shouldn't this be an error and raise an exception?

PR #490, opened 2019-06-11
Add support of BGZF for preprocessor

allieychen:
Tested: integration tests.


gcp_variant_transforms/pipeline_common.py
```diff
@@ -199,6 +200,45 @@ def read_headers(pipeline, pipeline_mode, all_patterns):
   return headers
 
 
+def read_variants(pipeline,  # type: beam.Pipeline
+                  all_patterns,  # type: List[str]
+                  pipeline_mode,  # type: int
+                  allow_malformed_records,  # type: bool
+                  representative_header_lines=None,  # type: List[str]
+                  vcf_parser=vcfio.VcfParserType.PYVCF  # type: int
```
tneymanov: Umm, I know the underlying value of the enum is int, but shouldn't the type of this argument still be VcfParserType? not sure about convention here.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -199,6 +200,45 @@ def read_headers(pipeline, pipeline_mode, all_patterns):
...
+                 ):
+  # type: (...) -> pvalue.PCollection
+  """Returns a PCollection of Variants by reading VCFs."""
+  compression_type = get_compression_type(all_patterns)
+  if compression_type == filesystem.CompressionTypes.GZIP:
+    splittable_bgzf = get_splittable_bgzf(all_patterns)
```
tneymanov: Is get_splittable_bgzf being used outside pipeline common anymore? if not, add an underscore.

allieychen: Done. Thanks!

gcp_variant_transforms/pipeline_common.py
```diff
@@ -274,3 +314,15 @@ def generate_unique_name(job_name):
   return '-'.join([job_name,
                    datetime.now().strftime('%Y%m%d-%H%M%S'),
                    str(uuid.uuid4())])
+
+
+class FusionBreak(beam.PTransform):
```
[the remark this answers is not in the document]
allieychen: Do you mean `ReadFromVcf` and `ReadAllFromVcf`? I didn't do it because I don't want to add the same code into both of them. Does this make sense to you?

gcp_variant_transforms/pipeline_common.py
allieychen: Done. Replaced with PipelineModes.

allieychen:
Thanks Aaron for the review. I didn't do it because otherwise we need to add the same code into both ReadFromVcf and ReadAllFromVcf. Does this make sense to you?

gcp_variant_transforms/pipeline_common.py
allieychen: Also changed the vcf_parser type.

gcp_variant_transforms/pipeline_common.py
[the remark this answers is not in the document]
allieychen: As discussed offline, moved `FusionBreak` into a separate script.

PR #491, opened 2019-06-14
Introduce PySam as an alternative parsing library.

tneymanov:
Uses multi threading and two pipes - to send data from parent thread to child, and from child into PySam class VariantFile (which is used by parent to read records). This is required to comply and to not overwrite base methods in VcfParser.

In order to send a line at a time in each and parse them one by one, readline() method is used - alternative methods deadlock, as they expect EOF before they can proceed. Therefore, every line read should be accompanied with a break-line character.

Due to quota issues, haven't checked the correctness of the integration tests yet, but manual run works and produces a table. Will update once quota is available and integration tests are run.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,6 +32,18 @@ ADD / /opt/gcp_variant_transforms/src/
 # Needed for installing mmh3 (one of the required packages in setup.py).
 RUN apt install -y g++
 
+# Install Pysam dependencies.
+RUN apt-get update && apt-get install -y autoconf \
+    automake \
+    make \
+    gcc \
+    perl \
+    zlib1g-dev \
+    libbz2-dev \
+    liblzma-dev \
+    libcurl4-openssl-dev \
+    libssl-dev
+
 # Install dependencies.
 RUN python -m pip install --upgrade pip && \
     python -m pip install --upgrade virtualenv && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,17 +32,19 @@ ADD / /opt/gcp_variant_transforms/src/
 # Needed for installing mmh3 (one of the required packages in setup.py).
 RUN apt install -y g++
 
-# Install Pysam dependencies.
-RUN apt-get update && apt-get install -y autoconf \
+# Install Pysam dependencies. These dependencies are only required becuase we
+# have a monolithic binary - they primarily have to be installed on the workers.
+RUN apt-get update && apt-get install -y \
+    autoconf \
     automake \
-    make \
     gcc \
-    perl \
-    zlib1g-dev \
     libbz2-dev \
-    liblzma-dev \
     libcurl4-openssl-dev \
-    libssl-dev
+    liblzma-dev \
+    libssl-dev \
+    make \
+    perl \
+    zlib1g-dev
 
 # Install dependencies.
 RUN python -m pip install --upgrade pip && \
--- a/gcp_variant_transforms/beam_io/vcf_parser.py
+++ b/gcp_variant_transforms/beam_io/vcf_parser.py
@@ -538,62 +538,41 @@ class PySamParser(VcfParser):
                                       allow_malformed_records,
                                       representative_header_lines,
                                       **kwargs)
-    # These members will be properly initiated in _init_with_header().
+    # These members will be properly initiated in _init_parent_process().
     self._vcf_reader = None
     self._to_child = None
     self._original_info_list = None
 
-  def _init_parent_process(self, p_read, c_write, c_read, p_write):
-    # Parent Process
-    os.close(c_read)
-    os.close(c_write)
-    into_pysam = os.fdopen(p_read)
-    self._to_child = os.fdopen(p_write, 'w')
+  def _init_parent_process(self, pysam_read, variants_write):
+    into_pysam = os.fdopen(pysam_read)
+    self._to_child = os.fdopen(variants_write, 'w')
     self._vcf_reader = libcbcf.VariantFile(into_pysam, 'r')
     self._original_info_list = self._vcf_reader.header.info.keys()
 
-  def _write_header_lines(self, to_pysam, header_lines):
-    # Feed the data with a break line, since we read data line by line - if
-    # we would use the conventional ``read()`` method, both pipes would lock,
-    # while waiting for an EOF signal.
+  def _write_header_lines(self, to_pysam_pipe, header_lines):
     for header in header_lines:
-      to_pysam.write(header + '\n')
-    to_pysam.flush()
+      to_pysam_pipe.write(header + '\n')
+    to_pysam_pipe.flush()
 
-  def _write_to_second_pipe(self, text_line):
-    self._to_child.write(text_line.encode('utf-8') + '\n')
-    self._to_child.flush()
-
-  def _read_from_second_pipe(self, from_parent):
-    return from_parent.readline()
-
-  def _write_to_first_pipe(self, to_pysam, text_line):
-    to_pysam.write(text_line)
-    to_pysam.flush()
-
-  def _read_from_first_pipe(self):
-    return next(self._vcf_reader)
-
-  def _init_child_process(self, p_read, c_write, c_read, p_write, header_lines):
+  def _init_child_process(self, variants_read, pysam_write, header_lines):
     # Child process' task is to populate data into the pipe that feeds
     # VariantFile class - first by populating all of the header lines, and then
     # by redirecting the data received from the second pipe.
 
-    os.close(p_read)
-    os.close(p_write)
-    from_parent = os.fdopen(c_read)
-    to_pysam = os.fdopen(c_write, 'w')
+    from_variants_pipe = os.fdopen(variants_read)
+    to_pysam_pipe = os.fdopen(pysam_write, 'w')
 
-    self._write_header_lines(to_pysam, header_lines)
+    self._write_header_lines(to_pysam_pipe, header_lines)
 
     while True:
-      text_line = self._read_from_second_pipe(from_parent)
+      text_line = from_variants_pipe.readline()
       if not text_line:
         break
-      self._write_to_first_pipe(to_pysam, text_line)
+      to_pysam_pipe.write(text_line)
+      to_pysam_pipe.flush()
 
-    from_parent.close()
-    to_pysam.close()
+    from_variants_pipe.close()
+    to_pysam_pipe.close()
     sys.exit(0)
 
   def _init_with_header(self, header_lines):
@@ -602,24 +581,24 @@ class PySamParser(VcfParser):
         FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='')):
       header_lines.insert(0, FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='4.0'))
 
-    # First pipe is responsible for supplying data from child process into the
-    # VariantFile, since it only takes an actual file descriptor as its input.
-    p_read, c_write = os.pipe()
-    # Since vcf_parser processes 1 line at a time, a second pipe is used to send
-    # 1 record at a time to the child process, to pass it downstream back to
-    # parent process via the first pipe/VariantFile class.
-    c_read, p_write = os.pipe()
+    # Pysam pipe is responsible for supplying lines from child process into the
+    # VariantFile, since it can only take an actual file descriptor as an input.
+    pysam_read, pysam_write = os.pipe()
+    # Since child process doesn't have access to the lines that need to be
+    # parsed, variants pipe is needed to supply them from _get_variant() method
+    # into the child process, to be propagated afterwards into the pysam pipe.
+    variants_read, variants_write = os.pipe()
 
-    processid = os.fork()
-    if processid:
-      self._init_parent_process(p_read, c_write, c_read, p_write)
+    if os.fork():
+      self._init_parent_process(pysam_read, variants_write)
     else:
-      self._init_child_process(p_read, c_write, c_read, p_write, header_lines)
+      self._init_child_process(variants_read, pysam_write, header_lines)
 
   def _get_variant(self, data_line):
     try:
-      self._write_to_second_pipe(data_line)
-      return self._convert_to_variant(self._read_from_first_pipe())
+      self._to_child.write(data_line.encode('utf-8') + '\n')
+      self._to_child.flush()
+      return self._convert_to_variant(next(self._vcf_reader))
     except (MemoryError, IOError) as e:
       logging.warning('VCF record read failed in %s for line %s: %s',
                       self._file_name, data_line, str(e))
--- a/gcp_variant_transforms/beam_io/vcf_parser.py
+++ b/gcp_variant_transforms/beam_io/vcf_parser.py
@@ -647,7 +647,7 @@ class PySamParser(VcfParser):
     calls = []
 
     for (name, sample) in samples.iteritems():
-      phaseset = DEFAULT_PHASESET_VALUE if sample.phased else None
+      phaseset = None
       genotype = None
       info = {}
       for (key, value) in sample.iteritems():
@@ -655,15 +655,19 @@ class PySamParser(VcfParser):
           if isinstance(value, tuple):
             genotype = []
             for elem in value:
-              genotype.append(elem if elem else MISSING_GENOTYPE_VALUE)
+              genotype.append(MISSING_GENOTYPE_VALUE if elem is None else elem)
           else:
-            genotype = value if value else MISSING_GENOTYPE_VALUE
+            genotype = MISSING_GENOTYPE_VALUE if value is None else value
 
         elif key == PHASESET_FORMAT_KEY:
           phaseset = list(value) if isinstance(value, tuple) else value
         else:
           info[key] = list(value) if isinstance(value, tuple) else value
 
+      # PySam samples are "phased" for haploids, so check for for the type
+      # before settings default phaseset value.
+      if phaseset is None and sample.phased and len(genotype) > 1:
+        phaseset = DEFAULT_PHASESET_VALUE
       calls.append(VariantCall(name, genotype, phaseset, info))
 
     return calls
--- a/gcp_variant_transforms/beam_io/vcf_parser.py
+++ b/gcp_variant_transforms/beam_io/vcf_parser.py
@@ -21,10 +21,10 @@ from __future__ import absolute_import
 
 from collections import namedtuple
 from copy import copy
+from typing import Iterable # pylint: disable=unused-import
 import logging
 import os
 import tempfile
-import typing  # pylint: disable=unused-import
 
 from apache_beam.coders import coders
 from apache_beam.io import filesystems
@@ -234,6 +234,7 @@ class VariantCall(object):
     return ', '.join(
         [str(s) for s in [self.name, self.genotype, self.phaseset, self.info]])
 
+
 class VcfParser(object):
   """Base abstract class for defining a VCF file parser.
 
@@ -320,8 +321,8 @@ class VcfParser(object):
     return
 
   def _next_non_empty_line(self, iterator):
+    # type: (Iterable[str]) -> str
     # Get next non-empty stripped record from iterator.
-    # type: (typing.Iterable[str]) -> str
     text_line = next(iterator).strip()
     while not text_line:  # skip empty lines.
       text_line = next(iterator).strip()
@@ -742,8 +743,8 @@ class PySamParser(VcfParser):
     return info
 
   def _parse_to_numeric(self, value):
-    # Attempt to parse a value to int, then float or return as string.
     # type: (Any) -> Any
+    # Attempt to parse a value to int, then float or return as string.
     if isinstance(value, str) and value.isdigit():
       return int(value)
     else:
--- a/gcp_variant_transforms/beam_io/vcf_parser.py
+++ b/gcp_variant_transforms/beam_io/vcf_parser.py
@@ -20,14 +20,15 @@ The 4.2 spec is available at https://samtools.github.io/hts-specs/VCFv4.2.pdf.
 from __future__ import absolute_import
 
 from collections import namedtuple
+from copy import copy
 import logging
 import os
-import sys
 import tempfile
 
 from apache_beam.coders import coders
 from apache_beam.io import filesystems
 from apache_beam.io import textio
+from apache_beam.io import range_trackers
 try:
   from nucleus.io.python import vcf_reader as nucleus_vcf_reader
   from nucleus.protos import variants_pb2
@@ -54,6 +55,9 @@ DEFAULT_PHASESET_VALUE = '*'  # Default phaseset value if call is phased, but
                               # no 'PS' is present.
 MISSING_GENOTYPE_VALUE = -1  # Genotype to use when '.' is used in GT field.
 FILE_FORMAT_HEADER_TEMPLATE = '##fileformat=VCFv{VERSION}'
+LAST_HEADER_LINE_PREFIX = '#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO'
+HEADER_INFO_TYPES = ['Integer', 'Float', 'Flag', 'Character', 'String', '.']
+HEADER_INFO_NUMBERS = ['A', 'R', 'G', '.']
 
 class Variant(object):
   """A class to store info about a genomic variant.
@@ -296,8 +300,55 @@ class VcfParser(object):
               self._process_header_lines),
           **kwargs)
 
+    if isinstance(self, PySamParser):
+      # In an absence of "#CHROM POS..." header line, PySam process would lock.
+      # However, we need to support cases when no records in files were provided
+      # even if the above header line is missing. To circumvent the error, read
+      # the file until the first line is found, and verify validity.
+      self._header_lines = None
+      if isinstance(range_tracker, range_trackers.UnsplittableRangeTracker):
+        secondary_tracker = range_trackers.UnsplittableRangeTracker(
+            copy(range_tracker._range_tracker))
+      else:
+        secondary_tracker = copy(range_tracker)
+      self._verify_header(copy(text_source), secondary_tracker)
+
     self._text_lines = text_source.read_records(self._file_name,
                                                 range_tracker)
+
+  def after_read(self):
+    return
+
+  def _verify_header(self, text_source, range_tracker):
+    try:
+      text_lines = text_source.read_records(self._file_name,
+                                            range_tracker)
+
+      text_line = next(text_lines).strip()
+      while not text_line.strip():  # skip empty lines.
+        # This natively raises StopIteration if end of file is reached.
+        text_line = next(text_lines).strip()
+      if not self._header_lines[-1].startswith(LAST_HEADER_LINE_PREFIX):
+        raise ValueError('Header missing CHROM line.')
+      header = libcbcf.VariantHeader()
+      for header_line in self._header_lines[:-1]:
+        header.add_line(header_line)
+
+      for k, v in header.info.iteritems():
+        if not k:
+          raise ValueError('Corrupt ID at header line {}.'.format(v.id))
+        if not v.description:
+          raise ValueError(
+              'Corrupt Description at header line {}.'.format(v.id))
+        if not v.type or v.record['Type'] not in HEADER_INFO_TYPES:
+          raise ValueError('Corrupt Type at header line {}'.format(v.id))
+        if not v.record['Number'] or (
+            v.record['Number'] not in HEADER_INFO_NUMBERS and
+            not v.record['Number'].isdigit()):
+          raise ValueError('Unknown Number at header line {}.'.format(v.id))
+    except StopIteration:
+      if not self._header_lines[-1].startswith(LAST_HEADER_LINE_PREFIX):
+        self._header_lines.append(LAST_HEADER_LINE_PREFIX)
 
   def _process_header_lines(self, header_lines):
     """Processes header lines from text source and initializes the parser.
@@ -309,13 +360,26 @@ class VcfParser(object):
       # We need to keep the last line of the header from the file because it
       # contains the sample IDs, which is unique per file.
       header_lines = self._representative_header_lines + header_lines[-1:]
-    self._init_with_header(header_lines)
+    if isinstance(self, PySamParser):
+      if header_lines and not header_lines[0].startswith(
+          FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='')):
+        header_lines.insert(
+            0, FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='4.0'))
+      header_lines = [line.strip() for line in header_lines if line.strip()]
+      self._header_lines = filter(None, [line.strip().replace(
+          'Number=G', 'Number=.').encode('utf-8') for line in header_lines])
+    else:
+      self._init_with_header(header_lines)
 
   def next(self):
-    text_line = next(self._text_lines).strip()
-    while not text_line:  # skip empty lines.
-      # This natively raises StopIteration if end of file is reached.
+    try:
       text_line = next(self._text_lines).strip()
+      while not text_line:  # skip empty lines.
+        # This natively raises StopIteration if end of file is reached.
+        text_line = next(self._text_lines).strip()
+    except StopIteration as e:
+      self.after_read()
+      raise e
     record = self._get_variant(text_line)
     if isinstance(record, Variant):
       return record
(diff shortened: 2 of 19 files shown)


allieychen:
Have you tried running tests on a large dataset yet? How is the speed?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +508,162 @@ def _get_variant_calls(self, record, formats):
...
+    try:
+      self._to_child.write(data_line + '\n')
+      self._to_child.flush()
+      return self._convert_to_variant(next(self._vcf_reader))
+    except (MemoryError, IOError) as e:
+      logging.warning('VCF record read failed in %s for line %s: %s',
+                      self._file_name, data_line, str(e))
+      return MalformedVcfRecord(self._file_name, data_line, str(e))
+
+  def _convert_to_variant(
```
allieychen: I think one line is easier to read (put `self` and `record` in the same line).

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +508,162 @@ def _get_variant_calls(self, record, formats):
...
+    p_read, c_write = os.pipe()
+    c_read, p_write = os.pipe()
+
+    processid = os.fork()
+    if processid:
+      os.close(c_read)
+      os.close(c_write)
+      into_pysam = os.fdopen(p_read)
+      self._to_child = os.fdopen(p_write, 'w')
+      self._vcf_reader = libcbcf.VariantFile(into_pysam, 'r')
```
allieychen: I don't know why we cannot use the file name directly to create the `VariantFile`.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +508,162 @@ def _get_variant_calls(self, record, formats):
...
+      if k != END_INFO_KEY:
+        if isinstance(v, tuple):
+          info[k] = list(v)
+        else:
+          info[k] = v
+
+    return info
+
+  def _get_variant_calls(self,
+                         samples  # type: libcvcf.VariantRecordSamples
```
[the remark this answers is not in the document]
tneymanov: Done

gcp_variant_transforms/beam_io/vcf_parser.py
tneymanov: Done

gcp_variant_transforms/beam_io/vcf_parser.py
tneymanov: Which file name?

So as discussed offline, `VariantFile` only accepts actual file descriptor, and mocking it is not an option. Therefore, we create pipe to push data in. To use pipe, we need to create child subthread (in the else clause). Ideally, I would supply the entire record list through child thread and write it into the pipe, to get it parsed from VariantFile. However, due to the nature of how `vcf_parser` is structured, the `next()` reads the `text_lines` and parses them line by line, and unless I override it, I can only add logic in the `get_variant()`, where I parse the records on 1-by-1 basis. To achieve this, I create a second pipe:

Now, after supplying all of the header lines into the pipe, child thread waits for new records to come. When `get_variant()` is invoked with a record in the parent thread, it sends the record to child thread via second pipe. Once received, child pushes the record into the first pipe to the VariantFile class in parent thread, from where get_variant reads it back and returns it.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -19,10 +19,14 @@
...
+from collections import namedtuple
+
+import sys
+import tempfile
+
 import logging
 import os
-import tempfile
-from collections import namedtuple
+from pysam import libcbcf
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -19,10 +19,14 @@
 
 from __future__ import absolute_import
 
+from collections import namedtuple
+
+import sys
+import tempfile
+
```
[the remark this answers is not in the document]
tneymanov: Refactored imports.

gcp_variant_transforms/testing/integration/vcf_to_bq_tests/presubmit_tests/small_tests/valid_4_0_bz2_pysam.json
```diff
@@ -0,0 +1,24 @@
+[
+  {
+    "test_name": "valid-4-0-bz2-pysam",
+    "table_name": "valid_4_0_bz2_pysam",
+    "input_pattern": "gs://gcp-variant-transforms-testfiles/small_tests/valid-4.0.vcf.bz2",
+    "runner": "DirectRunner",
+    "zones": ["us-west1-b"],
+    "vcf_parser": "PYVCF",
```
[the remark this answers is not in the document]
tneymanov: Wow, that was silly :D 

Done.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +504,155 @@ def _get_variant_calls(self, record, formats):
...
+  def _get_variant_calls(self, samples):
+    # type: (libcvcf.VariantRecordSamples) -> List[VariantCall]
+    calls = []
+
+    for (name, sample) in samples.iteritems():
+      if sample.phased:
+        phaseset = DEFAULT_PHASESET_VALUE
+      else:
+        phaseset = None
```
allieychen: simplify to `phaseset = DEFAULT_PHASESET_VALUE if sample.phased else None` if it can fit into one line

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +504,155 @@ def _get_variant_calls(self, record, formats):
...
+      genotype = None
+      info = {}
+      for (key, value) in sample.iteritems():
+        if key == GENOTYPE_FORMAT_KEY:
+          genotype = list(value) if isinstance(value, tuple) else value
+        elif key == PHASESET_FORMAT_KEY:
+          phaseset = list(value) if isinstance(value, tuple) else value
```
allieychen: The above phaseset value will be override.

gcp_variant_transforms/beam_io/vcf_parser.py
tneymanov: Hmmm, doing the above refactoring is not difficult, but seems a bit pointless - those methods would have like 1, maybe 2 line each, and would not be repeated anywhere.

Instead, parsed out the child thread into its own method, and added lots of comments. WDYT?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,173 @@ def _get_variant_calls(self, record, formats):
...
+  def _init_with_header(self, header_lines):
+    # The first header line must be similar to '##fileformat=VCFv.*'.
+    if header_lines and not header_lines[0].startswith(
+        FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='')):
+      header_lines.insert(0, FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='4.0'))
+
+    # First pipe is repsonsible for supplying data from child thread into the
+    # VariantFile, since it only takes an actual file descriptor as its input.
+    p_read, c_write = os.pipe()
+    # Since vcf_parser is setup to parse 1 line at a time, second pipe is used
```
kemp-google: `# Since vcf_parser processes 1 line at a time, a second pipe...`

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,173 @@ def _get_variant_calls(self, record, formats):
...
+  def _init_child_thread(self, p_read, c_write, c_read, p_write, header_lines):
+    # Child thread's task is to populate data into the pipe that feeds
+    # VariantFile class - first by populating all of the header lines, and then
+    # by redirecting the data received from the second pipe.
+
+    os.close(p_read)
+    os.close(p_write)
+    from_parent = os.fdopen(c_read)
+    to_pysam = os.fdopen(c_write, 'w')
+    # Feed the data with a break line, since we read data line by line - if
```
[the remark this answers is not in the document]
tneymanov: If we just use `read()`, threads would lock, as it would wait to read until EOF is reached - EOF is not reached, because processes get invoked when next line is 'requested' and wait on each other to get a new record to pass downstream. If the parser wasn't an iterator, we maybe could have just supplied all of the lines to parse, but alas, we cannot do that here.

`..next() invoked`
`......|`
`......V`
`..parent process  ---------------{raw record}---------------->  child process`
`..parent process <--{parsed record}-- VariantFile <--{raw record}--/`
`......|`
`......V`
`..return {parsed record}.`

Given that we cannot rely on EOF to do the reads, we can only rely on EOL, which we utilize via `readline()` instead. Therefore, we make sure that every line to parse has **single** newline char at the end (by trimming and then adding 1 to the end).

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,173 @@ def _get_variant_calls(self, record, formats):
...
+    # we would use the conventional ``read()`` method, both pipes would lock,
+    # while waiting for an EOF signal.
+    for header in header_lines:
+      to_pysam.write(header.strip() + '\n')
+    to_pysam.flush()
+
+    while 1:
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +504,155 @@ def _get_variant_calls(self, record, formats):
...
+    self._current_line = None
+    # This member will be properly initiated in _init_with_header().
+    self._vcf_reader = None
+    self._to_child = None
+
+  def _init_with_header(self, header_lines):
+    # The first header line must be similar to '##fileformat=VCFv.*'.
+    if header_lines and not header_lines[0].startswith(
+        FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='')):
+      header_lines.insert(0, FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='4.0'))
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,169 @@ def _get_variant_calls(self, record, formats):
...
+    to_pysam.close()
+    sys.exit(0)
+
+  def _init_with_header(self, header_lines):
+    # The first header line must be similar to '##fileformat=VCFv.*'.
+    if header_lines and not header_lines[0].startswith(
+        FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='')):
+      header_lines.insert(0, FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='4.0'))
+
+    # First pipe is repsonsible for supplying data from child thread into the
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/beam_io/vcf_parser.py
tneymanov: Done.

tneymanov:
Ok, now, based on UID I figure out if caller is a super user and weather sudo is required. Works both in local virtual environment with sudo (although will require user to enter their password obviously), and on cloud without it. Fully ready for review.

docker/Dockerfile
```diff
@@ -32,6 +32,18 @@ ADD / /opt/gcp_variant_transforms/src/
 # Needed for installing mmh3 (one of the required packages in setup.py).
 RUN apt install -y g++
 
+# Install Pysam dependencies.
```
kemp-google: I think it's worth calling out that this is only needed because we have a monolithic binary.  Otherwise we only need these dependencies on the Dataflow worker VM.

docker/Dockerfile
```diff
@@ -32,6 +32,18 @@ ADD / /opt/gcp_variant_transforms/src/
...
+RUN apt-get update && apt-get install -y autoconf \
```
kemp-google: put autoconf on the next line and sort please

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,189 @@ def _get_variant_calls(self, record, formats):
...
+  that will feed into PySam object - 1 will feed the data from main process
+  throughout into the child one, while second will get that data in child thread
+  and feed it to PySam library.
+
+  The requirement of using two pipelines comes from the design of VcfParser base
+  class - we could only use a single pipe, but it will divert the parsers.
+  """
+
+  def __init__(self,
+               file_name,  # type: str
```
kemp-google: python style question: we don't line these up we leave them jagged? (the type annotations)

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,189 @@ def _get_variant_calls(self, record, formats):
...
+              ):
+    # type: (...) -> None
+    super(PySamParser, self).__init__(file_name,
+                                      range_tracker,
+                                      file_pattern,
+                                      compression_type,
+                                      allow_malformed_records,
+                                      representative_header_lines,
+                                      **kwargs)
+    # This member will be properly initiated in _init_with_header().
```
kemp-google: Just _vcf_reader?  If so, put the comment after it?

`self._vcf_reader = None  # Initialized in _init_with_header().`

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,189 @@ def _get_variant_calls(self, record, formats):
...
+    self._vcf_reader = None
+    self._to_child = None
+    self._original_info_list = None
+
+  def _init_child_thread(self, p_read, c_write, c_read, p_write, header_lines):
```
kemp-google: These aren't threads they are processes (I thought I'd left that comment but I guess not?)

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,189 @@ def _get_variant_calls(self, record, formats):
...
+    # Child thread's task is to populate data into the pipe that feeds
+    # VariantFile class - first by populating all of the header lines, and then
+    # by redirecting the data received from the second pipe.
+
+    os.close(p_read)
```
kemp-google: Why close these streams at the outset?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,189 @@ def _get_variant_calls(self, record, formats):
...
+    p_read, c_write = os.pipe()
+    # Since vcf_parser processes 1 line at a time, a second pipe is used to send
+    # 1 record at a time to the child thread, to pass it downstream back to
+    # parent thread via the first pipe/VariantFile class.
+    c_read, p_write = os.pipe()
+
+    processid = os.fork()
+    if processid:
+      # Parent Thread
+      os.close(c_read)
```
kemp-google: Why bother closing these?

gcp_variant_transforms/beam_io/vcf_parser.py
kemp-google: I generally agree with Allie here that it would be nicer if we had

if processid:
  writer()
else:
  reader()

sort of thing.

setup.py
```diff
@@ -40,6 +53,51 @@
...
+
+  def finalize_options(self):
+    pass
+
+  def RunCustomCommand(self, command_list):
+    print 'Running command: %s' % command_list
+    p = subprocess.Popen(
+        command_list,
+        stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    # Can use communicate(input='y\n'.encode()) if the command run requires
```
kemp-google: Drop these comments

setup.py
```diff
@@ -40,6 +53,51 @@
...
+
+  def initialize_options(self):
+    pass
+
+  def finalize_options(self):
+    pass
+
+  def RunCustomCommand(self, command_list):
+    print 'Running command: %s' % command_list
+    p = subprocess.Popen(
```
kemp-google: Does python not have a simpler exec form when we don't care about using the standard input?

tneymanov: Done, but error handling is worse this way.

gcp_variant_transforms/beam_io/vcf_parser.py
kemp-google: It makes it less clear in my mind: you are touching things that you don't need to touch

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,208 @@ def _get_variant_calls(self, record, formats):
...
+    to_pysam.close()
+    sys.exit(0)
+
+  def _init_with_header(self, header_lines):
+    # PySam requires a version header to be present, so add one if absent.
+    if header_lines and not header_lines[0].startswith(
+        FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='')):
+      header_lines.insert(0, FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='4.0'))
+
+    # First pipe is responsible for supplying data from child process into the
```
kemp-google: pysam_read, pysam_write ?

variants_read, variants_write ?

I'd explain here that the reason for the variants pipe is that we don't have access to the pysam handle to call into the library from the get_variant function because it's a different process.  It's not really about lines or anything else, it's just that we can't call into pysam from that process.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,187 @@ def _get_variant_calls(self, record, formats):
...
+  def _init_parent_process(self, pysam_read, variants_write):
+    into_pysam = os.fdopen(pysam_read)
+    self._to_child = os.fdopen(variants_write, 'w')
+    self._vcf_reader = libcbcf.VariantFile(into_pysam, 'r')
+    self._original_info_list = self._vcf_reader.header.info.keys()
+
+  def _write_header_lines(self, to_pysam_pipe, header_lines):
+    for header in header_lines:
+      to_pysam_pipe.write(header + '\n')
+    to_pysam_pipe.flush()
```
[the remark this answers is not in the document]
tneymanov: Writelines doesn't add new line characters. Refactored a bit, to make it cleaner.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +506,187 @@ def _get_variant_calls(self, record, formats):
...
+
+  def _init_child_process(self, variants_read, pysam_write, header_lines):
+    # Child process' task is to populate data into the pipe that feeds
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -296,9 +300,56 @@ def __init__(self,
...
+      text_line = next(text_lines).strip()
+      while not text_line.strip():  # skip empty lines.
+        # This natively raises StopIteration if end of file is reached.
+        text_line = next(text_lines).strip()
+      if not self._header_lines[-1].startswith(LAST_HEADER_LINE_PREFIX):
+        raise ValueError('Header missing CHROM line.')
+      header = libcbcf.VariantHeader()
+      for header_line in self._header_lines[:-1]:
+        header.add_line(header_line)
+
```
kemp-google: It would be nice to describe the checks here and why they exist (presumably these are things you found by trial and error than htslib doesn't like?)

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -309,13 +360,26 @@ def _process_header_lines(self, header_lines):
       # We need to keep the last line of the header from the file because it
       # contains the sample IDs, which is unique per file.
       header_lines = self._representative_header_lines + header_lines[-1:]
-    self._init_with_header(header_lines)
+    if isinstance(self, PySamParser):
```
kemp-google: Same here: a description of why this header munging is needed for the pysam case is helpful

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -309,13 +360,26 @@ def _process_header_lines(self, header_lines):
...
+    else:
+      self._init_with_header(header_lines)
 
   def next(self):
-    text_line = next(self._text_lines).strip()
-    while not text_line:  # skip empty lines.
-      # This natively raises StopIteration if end of file is reached.
+    try:
       text_line = next(self._text_lines).strip()
+      while not text_line:  # skip empty lines.
```
kemp-google: Can this not be a little function called next_non_empty_line()?  And then we can save a few lines of duplication and drop the comment?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -425,8 +489,10 @@ def _convert_to_variant(
       formats: The PyVCF dict storing FORMAT extracted from the VCF header.
         The key is the FORMAT key and the value is
         :class:`~vcf.parser._Format`.
+
```
kemp-google: avoid making unrelated whitespace changes - they just complicate the diff

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +570,224 @@ def _get_variant_calls(self, record, formats):
...
+  def _parse_to_numeric(self, value):
+    # type: (Any) -> Any
+    if isinstance(value, str) and value.isdigit():
+      return int(value)
+    else:
+      try:
+        return self._parse_float(float(value))
+      except ValueError:
+        # non float
+        if isinstance(value, unicode):
```
kemp-google: What is the unicode case trying to do exactly?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +570,224 @@ def _get_variant_calls(self, record, formats):
...
+        else:
+          # If a field was not provided in the header, make it a list to
+          # follow the PyVCF standards.
+          info[k] = (
+              self._convert_field(v) if k in self._original_info_list else
+              [self._convert_field(v)])
+
+    return info
+
+  def _parse_to_numeric(self, value):
```
kemp-google: What is this function meant to do?  A comment on the function would be very helpful

It looks like it is trying a variety of conversions to coerce a string (or number?) value into another type of numeric value

Is there no python built in functions for this?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +570,224 @@ def _get_variant_calls(self, record, formats):
...
+    if precise_value.is_integer():
+      return int(precise_value)
+    return precise_value
+
+  def _convert_field(self, value, numeric=True):
+    # type: (Any) -> Any
+    # PySam currently doesn't recognize '.' value for String fields as missing.
+    if value == '.' or value == b'.' or not value:
+      return None
+    elif isinstance(value, unicode):
```
kemp-google: Again I'm not sure what this is doing - what is a 'unicode' object in python and why does it need special handling?

tneymanov: Well unicode an 'unencoded' sring. PySam sometimes returns unicode values, sometimes str-s in utf-8 encoding, so I encode all to utf-8, to stay consistent.

pgrosu: @tneymanov But do you know why that is?  There is a very specific reason.

tneymanov: Did some analysis over the valid-4.1-large.vcf file - seems like whenever PySam adds a string of its own (like '.' or 'PASS'), or when an INFO field is set to be string, PySam return unicode. In fact the only time it returns str type is for REF alleles...

Is there any other reason? Or any insight why it works like that above?

pgrosu: Pretty good first analysis!  To make a short story short, it has to have some consistency among Python versions.  Strings in Python 3 are unicode, while in Python 2 you have to make them unicode.  See: https://portingguide.readthedocs.io/en/latest/strings.html

That's why they mentioned it specifically starting with release 0.8.1 for Python 3 in https://pysam.readthedocs.io/en/latest/release.html#release-0-8-1

That's why you see things like this in [libcutils.pyx](https://github.com/pysam-developers/pysam/blob/master/pysam/libcutils.pyx#L157-L168):

```Python
cdef force_str(object s, encoding=TEXT_ENCODING):
    """Return s converted to str type of current Python
    (bytes in Py2, unicode in Py3)"""
    if s is None:
        return None
    if PY_MAJOR_VERSION < 3:
        return s
    elif PyBytes_Check(s):
        return s.decode(encoding)
    else:
        # assume unicode
        return s
```

Or in [libcbcf.pyx](https://github.com/pysam-developers/pysam/blob/master/pysam/libcbcf.pyx#L151-L154):

```Python
    if PY_MAJOR_VERSION < 3:
        val = s
    else:
        val = PyUnicode_DecodeUTF8(s, strlen(s), NULL)
```

setup.py
```diff
@@ -33,13 +46,52 @@
...
+      self.RunCustomCommand(PYSAM_INSTALLATION_COMMAND)
+
+    except RuntimeError:
+      raise RuntimeError(
+          'PySam installation has failed. Make sure you have the ' + \
+          'following packages installed: autoconf automake gcc libbz2-dev ' + \
+          'liblzma-dev libcurl4-openssl-dev libssl-dev make perl zlib1g-dev')
+
+class build(_build):  # pylint: disable=invalid-name
+  """A build command class that will be invoked during package install.
```
allieychen: Please leave one blank line between the summary and the additional description.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -296,9 +300,68 @@ def __init__(self,
...
+      if not self._header_lines[-1].startswith(LAST_HEADER_LINE_PREFIX):
+        raise ValueError('Header missing CHROM line.')
+      # Construct VariantHeader class from the metadata lines, to verify the
+      # validity of the file. PySam would not raise an error, will omit the
+      # metadata and hide the issue, if the header is incorrect.
+      header = libcbcf.VariantHeader()
+      for header_line in self._header_lines[:-1]:
+        header.add_line(header_line)
+
+      for k, v in header.info.iteritems():
```
allieychen: Does PyVCF raise errors for these cases? I assume PySam does not so we have to check it by ourselves?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -296,9 +300,68 @@ def __init__(self,
...
+      text_line = next(iterator).strip()
+    return text_line
+
+  def _verify_header(self, text_source, range_tracker):
+    try:
+      text_lines = text_source.read_records(self._file_name,
+                                            range_tracker)
+      # Read until first non empty record
+      self._next_non_empty_line(text_lines)
+      if not self._header_lines[-1].startswith(LAST_HEADER_LINE_PREFIX):
```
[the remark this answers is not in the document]
tneymanov: That's exactly what this section is doing:
It reads the text_line. If line starts with # (ie it's a header), it's skipped so read doesn't happen (due to process_header_lines section). If there are no records, text_lines = text_source... will raise StopIteration right away, which we catch and in that case, if absent, add default CHROM... line (just to "read" an empty file and skip it without complications). However, if text_lines = text_source... read happens, that means that there is at least 1 actual record, and thus we check if CHROM... line exists, and if it does not, raise an error.

gcp_variant_transforms/beam_io/vcf_parser.py
tneymanov: Yeah, PyVCF does, there are tests for that.

PySam is too smart for its own good, and tries to swallow this issues to allow clients to go through even with incorrect headers. However, for our case, we would want to report header issues upstream to the caller.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +587,225 @@ def _get_variant_calls(self, record, formats):
...
+      self._to_child.write(data_line.encode('utf-8') + '\n')
+      self._to_child.flush()
+      return self._convert_to_variant(next(self._vcf_reader))
+    except (MemoryError, IOError, ValueError) as e:
+      logging.warning('VCF record read failed in %s for line %s: %s',
+                      self._file_name, data_line, str(e))
+      return MalformedVcfRecord(self._file_name, data_line, str(e))
+
+  def _verify_record(self, record):
+    if record.start < 0:
```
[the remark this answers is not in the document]
tneymanov: Because if pos is for example String for some reason, PySam puts the value as MIN_INT (-2147483648), instead of throwing an error.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -504,6 +587,225 @@ def _get_variant_calls(self, record, formats):
...
+          # follow the PyVCF standards.
+          info[k] = (
+              self._convert_field(v) if k in self._original_info_list else
+              [self._convert_field(v)])
+
+    return info
+
+  def _parse_to_numeric(self, value):
+    # Attempt to parse a value to int, then float or return as string.
+    # type: (Any) -> Any
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -296,9 +300,68 @@ def __init__(self,
...
+
     self._text_lines = text_source.read_records(self._file_name,
                                                 range_tracker)
 
+  def after_read(self):
+    return
+
+  def _next_non_empty_line(self, iterator):
+    # Get next non-empty stripped record from iterator.
+    # type: (typing.Iterable[str]) -> str
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/beam_io/vcf_parser.py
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -261,22 +261,20 @@ def __init__(self,
...
+      self._vcf_parser_class = vcf_parser.PyVcfParser
+    elif vcf_parser_type == VcfParserType.PYSAM:
+      self._vcf_parser_class = vcf_parser.PySamParser
+    else:
+      raise ValueError(
+          'Unrecognized vcf_parser_type: %s.' % str(vcf_parser_type))
 
   def _read_records(self, (file_path, block)):
     # type: (Tuple[str, Block]) -> Iterable(Variant)
     """Reads records from `file_path` in `block`."""
```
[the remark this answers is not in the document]
tneymanov: Hmmm, sure, but my preferences would had been to have one. Done.

gcp_variant_transforms/beam_io/vcfio.py
```diff
@@ -373,7 +371,7 @@ def __init__(
       desired_bundle_size=DEFAULT_DESIRED_BUNDLE_SIZE,  # type: int
       compression_type=CompressionTypes.AUTO,  # type: str
       allow_malformed_records=False,  # type: bool
-      vcf_parser_type=VcfParserType.PYVCF,
+      vcf_parser_type=VcfParserType.PYVCF, # int
```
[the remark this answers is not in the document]
tneymanov: Right, my bad. Done.

tneymanov:
Finally ready for review, apologies for the delay.

Changes in the latest commit:
-   Pull latest changes to include region changes which are a necessary now (sorry about that :( ).
-   Modify all instances where PyVCF reader and internal fields are hard-coded to either be ambiguous on which parser is being used, or to rely on specific selection of the parser (either by PySam or PyVCF, since, if Nucleus is selected, PyVCF would had been defaulted).
-   Add integration test for infer headers with PySam ([still running](https://pantheon.corp.google.com/cloud-build/builds/846c7de2-5615-4702-b62f-e19b3f6a393a?project=tural-test-runner)), modify old PySam integration tests to use region instead of zone.
-   Fix a bug where PyVCF and PySam would hang if "#CHROM..." line wasn't provided in with the "merge headers" selection.
-   Modify deployment of integration tests to not install pysam manually - not required, the issue was with other sections of the code.

Note that documentation wasn't properly modified as the PR is enormous as is - docs and comments will be changed in the PR that removes PyVCF and Nucleus entirely.

Currently in the process of modifying the documentation on PySam changes to include additional, header related findings.
Additionally tested DirectRunner infer headers on valid-4.1-large.vcf, now initiating a run on 1k genomes.

samanvp:
Thanks Tural.
It seems this PR presents a transient state and your next PR (which will remove Nucleus and make PySAM as our _only_ VCF parser) will be our long term state. As we discuss offline we better not to spend too much time on this PR and rather focus on the next PR.

Issue #398, opened 2018-10-23
Add an option to specify filename per call


arostamianfar:
There are certain use cases where users need to distinguish between call names (perhaps even having the same ID), so having the filename as one of the columns would be super useful (e.g. call.filename).

See https://groups.google.com/forum/#!topic/google-genomics-discuss/ODrX6H6B48c for context.

We also need to ensure to exclude this new column when running BQ -> VCF.

smrgit:
@allieychen (since @arostamianfar has left) -- do you know if/when this might be implemented/released?  it would be very very useful!  thx!

kemp-google:
@smrgit we will try to prioritize for early in Q1.  I'll post another update in early January.

smrgit:
any news on this enhancement request? @kemp-google

allieychen:
one use case: https://groups.google.com/forum/#!topic/google-genomics-discuss/hWUBDsiBeTE

smrgit:
another use case is where there are multiple VCFs even for the exact same sample (or pair of samples) -- based on different bam files or different aligners or different versions of the code, etc

as a generalization of including the input VCF uri in a new column, on each row of the table written to BigQuery, I suggest that the --input_file option (#441 and also my suggestion) be extended to be able to include a 2nd column in that file where the 1st column is an input file name (or pattern) and the 2nd column is a "flag" or "tag" to be added to the resulting BigQuery table for each call from each corresponding file

the reasoning is that although an input VCF uri is probably a good and unique identifier, the user may prefer to specify the associated file GUID -- etc

kemp-google:
Hi Sheila,

Sorry for the delay on this.

The reason we have not implemented it yet is because we're worried about the increase in query cost that would result if we just add more data into each call.

In particular, filenames would represent a much larger source of data than the existing data stored in the calls.  Since BigQuery charges based on the amount of data scanned, this would result in queries being more expensive.

We'd like, instead, to get this auxiliary data out into another table and allow you to do JOINs.  This is complicated by the need for a stable join key.  We're looking into this now.

smrgit:
[...]

PR #493, opened 2019-06-17
Create a sample info table

allieychen:
- Use flag --create_sample_info_table to create this table.
- The table contains sample id and the file path.
- The sample id is also added into the variant record.
- Ignore sample id column for BQ to VCF pipeline.

issue: https://github.com/googlegenomics/gcp-variant-transforms/issues/398


allieychen:
Close this PR since I am going to break it into some smaller ones.

PR #495, opened 2019-06-18
Create sample info table.

allieychen:
Use flag --create_sample_info_table to create this table.
The table contains sample id and the file path.

issue: #398
To Do: add the sample id to the variant table and integration tests.


gcp_variant_transforms/libs/sample_info_table_schema_generator.py
```diff
@@ -0,0 +1,42 @@
...
+
+SAMPLE_ID = 'sample_id'
+FILE_PATH = 'file_path'
+
+
+def generate_schema():
+  # type: () -> bigquery.TableSchema
+  schema = bigquery.TableSchema()
+  schema.fields.append(create_sample_id_field())
+  schema.fields.append(bigquery.TableFieldSchema(
```
tneymanov: Seems weird that first field is added via create_sample_id_field() method, while the second one is added directly. I think both should be added in the same way (whichever you find more appropriate - your discretion).

gcp_variant_transforms/libs/sample_info_table_schema_generator.py
```diff
@@ -0,0 +1,42 @@
...
+      name=FILE_PATH,
+      type=bigquery_util.TableFieldConstants.TYPE_STRING,
+      mode=bigquery_util.TableFieldConstants.MODE_NULLABLE,
+      description=('File path of the sample.')))
+  return schema
+
+
+def create_sample_id_field():
```
tneymanov: start with '_' for private method.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -29,9 +29,12 @@
 from apache_beam import pvalue  # pylint: disable=unused-import
 from apache_beam.io import filesystem
 from apache_beam.io import filesystems
+from apache_beam.io.gcp import gcsio
 from apache_beam.options import pipeline_options
 from apache_beam.runners.direct import direct_runner
```
tneymanov: remove line

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -429,6 +448,14 @@ def run(argv=None):
 
   beam_pipeline_options = pipeline_options.PipelineOptions(pipeline_args)
   pipeline = beam.Pipeline(options=beam_pipeline_options)
+  google_cloud_options = beam_pipeline_options.view_as(
+      pipeline_options.GoogleCloudOptions)
+  file_path_to_file_hash = pipeline_common.create_file_path_to_file_hash_map(
```
tneymanov: This could be moved in either the if statement below, or even into the _create_sample_info_table method directly.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -378,6 +379,24 @@ def _run_annotation_pipeline(known_args, pipeline_args):
   return annotated_vcf_pattern
 
 
+def _create_sample_info_table(pipeline,  # type: beam.Pipeline
+                              pipeline_mode,  # type: PipelineModes
+                              known_args,  # type: argparse.Namespace
+                              file_path_to_file_hash  # type: Dict[str, str]
+                             ):
+  # type: (...) -> None
+  headers = pipeline_common.read_headers(pipeline,
```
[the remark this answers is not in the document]
tneymanov: Oh yeah, that's ugly, let's not do that.

Btw, what if user has supplied representative_header_file? Should we not read that file instead?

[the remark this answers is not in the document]
allieychen: Even when the `representative_header_file` is supplied, we still need to read the headers - we need the samples and the file path for each file, and that will be saved in the `VcfHeader`. We then can write the corresponding info into BigQuery.

allieychen:
I am going to send another PR since the design has changed.

Issue #494, opened 2019-06-17
move docker images to gcr.io/cloud-lifesciences 


allieychen:
Move all docker images to gcr.io/cloud-lifesciences. This is needed to operate Variant Transforms within VPC projects. For future releases, the docker image would be published in gcr.io/cloud-lifesciences as well.

PR #498, opened 2019-06-25
Move the vep cache and vep image to cloud-lifesciences project

allieychen:
- The vep cache and vep image are copied to cloud-lifesciences project.
- Change the default cache and vep image to point to cloud-lifesciences project.

Related issues: https://github.com/googlegenomics/gcp-variant-transforms/issues/494


allieychen:
Yes, the cache file and the image are already copied to the new bucket.

PR #499, opened 2019-06-25
Use new BigQuery sink and remove num_bigquery_write_shards flag usage.

tneymanov:
Modify BigQuery writing mechanism to use the new WriteToBigQuery PTransform, and deprecate num_bigquery_write_shards flag.

Previously, issue #199 forced us to use a hack to shard the variants before they are written to BigQuery, which negatively affects the processing speed. With the implementation of the new sink, the flag is no longer need.

The flag will remain as a dummy in order to not break any of the current callers.


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -173,7 +173,8 @@ def add_arguments(self, parser):
     parser.add_argument(
         '--num_bigquery_write_shards',
         type=int, default=1,
-        help=('Before writing the final result to output BigQuery, the data is '
+        help=('Note: This flag is now deprecated and should not be used! '
```
kemp-google: Can we nuke all the help text and just say "This flag is deprecated and may be removed in future releases" ?

docs/large_inputs.md
```diff
@@ -110,32 +108,14 @@ more expensive. However, when choosing a large machine (e.g. `n1-standard-16`),
 they can reduce cost as they can avoid idle CPU cycles due to disk IOPS
 limitations.
 
-As a result, we recommend using SSDs if [merging](variant_merge.md) or
-[--num_bigquery_write_shards](#--num_bigquery_write_shards) is enabled: these
-operations require "shuffling" the data (i.e. redistributing the data among
-workers), which require significant disk I/O.
+As a result, we recommend using SSDs if [merging](variant_merge.md) is enabled:
+these operations require "shuffling" the data (i.e. redistributing the data
```
allieychen: nit: It is not 'these' any more.

gcp_variant_transforms/options/variant_transform_options.py
allieychen: > Can we nuke all the help text and just say "This flag is deprecated and may be removed in future releases" ?

+1

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -136,35 +130,14 @@ def expand(self, pcoll):
...
-                  create_disposition=(
-                      beam.io.BigQueryDisposition.CREATE_IF_NEEDED),
-                  write_disposition=(
-                      beam.io.BigQueryDisposition.WRITE_APPEND
-                      if self._append
-                      else beam.io.BigQueryDisposition.WRITE_TRUNCATE))))
+                    beam.io.BigQueryDisposition.WRITE_APPEND
+                    if self._append
+                    else beam.io.BigQueryDisposition.WRITE_TRUNCATE),
+                method=beam.io.WriteToBigQuery.Method.STREAMING_INSERTS))
```
allieychen: Why do you choose to use streaming here? It costs extra money.

docs/large_inputs.md
tneymanov: Done.

gcp_variant_transforms/options/variant_transform_options.py
tneymanov: Done.

gcp_variant_transforms/transforms/variant_to_bigquery.py
tneymanov: Done.

tneymanov:
> LGTM. Is the performance of the bq sink similar to the beam sink? Also, use_beam_bq_sink is an experiment flag, I don't know whether there is any known downside or the plan for this flag from Beam's team? It might worth to talk to them.

For the new 1k genomes dataset it finished in 47 min 19 sec with 3,712 cores. Without it, it takes more than hour.

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -29,7 +28,6 @@
 from gcp_variant_transforms.libs import processed_variant
 from gcp_variant_transforms.libs import vcf_field_conflict_resolver
 from gcp_variant_transforms.libs.variant_merge import variant_merge_strategy  # pylint: disable=unused-import
-from gcp_variant_transforms.transforms import limit_write
 
 
 # TODO(samanvp): remove this hack when BQ custom sink is added to Python SDK,
```
samanvp: Please also remove the TODO and the const under it.

tneymanov:
Addressed Saman's comments from July.

gcp_variant_transforms/transforms/sample_info_to_bigquery.py
```diff
@@ -62,6 +62,7 @@ def __init__(self, output_table_prefix, append=False,
     self._append = append
     self.samples_span_multiple_files = samples_span_multiple_files
     self._schema = sample_info_table_schema_generator.generate_schema()
+    self._temp_location = temp_location
```
samanvp: Where `temp_location` is initialized?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -383,7 +383,8 @@ def _run_annotation_pipeline(known_args, pipeline_args):
 
 def _create_sample_info_table(pipeline,  # type: beam.Pipeline
                               pipeline_mode,  # type: PipelineModes
-                              known_args,  # type: argparse.Namespace
+                              known_args,  # type: argparse.Namespace,
+                              pipeline_args,  # type: List[str]
```
samanvp: `pipeline_args` is not used anywhere in this method.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -480,6 +480,8 @@ def run(argv=None):
     num_partitions = 1
 
   if known_args.output_table:
+    options = pipeline_options.PipelineOptions(pipeline_args)
```
samanvp: This variable is used only once in the next line. Is it possible to combine these two lines?

gcp_variant_transforms/transforms/variant_to_bigquery.py
```diff
@@ -136,35 +127,15 @@ def expand(self, pcoll):
...
-                      beam.io.BigQueryDisposition.CREATE_IF_NEEDED),
-                  write_disposition=(
-                      beam.io.BigQueryDisposition.WRITE_APPEND
-                      if self._append
-                      else beam.io.BigQueryDisposition.WRITE_TRUNCATE))))
+                    beam.io.BigQueryDisposition.WRITE_APPEND
+                    if self._append
+                    else beam.io.BigQueryDisposition.WRITE_TRUNCATE),
+                method=beam.io.WriteToBigQuery.Method.FILE_LOADS,
+                custom_gcs_temp_location=self._temp_location))
```
samanvp: It seems we don't need to set this argument as long as w have it in pipeline_args:
https://github.com/apache/beam/blob/7bea94c59d1ead24659e09b0e467beeb82f4cadd/sdks/python/apache_beam/io/gcp/bigquery.py#L1261

> custom_gcs_temp_location (str): A GCS location to store files to be used for file loads into BigQuery. By default, this will use the pipeline's temp_location, but for pipelines whose temp_location is not appropriate for BQ File Loads, users should pass a specific one.

Could you please test this and make sure it will work without setting this arg. If that's the case there is no need to pass temp_location to this module as well as `sample_info_to_bigquery.py`.

gcp_variant_transforms/testing/integration/vcf_to_bq_tests/presubmit_tests/small_tests/valid_4_1_pysam.json
```diff
@@ -1,23 +0,0 @@
-[
```
samanvp: Why this file is deleted?

samanvp:
Thanks Tural.
Please make sure we add the removed test in next PR.
Also this PR shouldn't merge unless we get a successful 1000Genome run without any missing rows.

PR #505, opened 2019-07-16
Fix BGZF parser failure:

allieychen:
- Fix the bug when discarding the first line and reading the last line for one block if the record is larger than one GZIP block.
- Use a variable to check whether it is the first time.

### DIFF
--- a/gcp_variant_transforms/beam_io/bgzf.py
+++ b/gcp_variant_transforms/beam_io/bgzf.py
@@ -56,7 +56,7 @@ class BGZF(filesystem.CompressedFile):
                fileobj,
                compression_type=filesystem.CompressionTypes.GZIP):
     super(BGZF, self).__init__(fileobj, compression_type)
-    self._first_time = True
+    self._first_fetch = True
 
   def _fetch_to_internal_buffer(self, num_bytes):
     """Fetches up to `num_bytes` into the internal buffer."""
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/beam_io/bgzf.py
```diff
@@ -52,13 +52,19 @@ class BGZF(filesystem.CompressedFile):
   This class repeatedly fetches the `unused_data` from the decompressor until
   all GZIP blocks are decompressed.
   """
+  def __init__(self,
+               fileobj,
+               compression_type=filesystem.CompressionTypes.GZIP):
+    super(BGZF, self).__init__(fileobj, compression_type)
+    self._first_time = True
```
kemp-google: _first_fetch ?

gcp_variant_transforms/beam_io/bgzf.py
```diff
@@ -99,7 +105,7 @@ def _read_data_from_source(self):
     return self._file.read(self._read_size)
 
   def _is_first_time(self):
```
kemp-google: is it really worth wrapping this field in a method?

gcp_variant_transforms/beam_io/bgzf.py
```diff
@@ -153,13 +159,9 @@ class BGZFBlock(BGZF):
 
   It reads contents from `self._block`.
     - The string before first `\n` is discarded.
-    - Decompress the next 64KB after self._block and reads the first line. It
-      assumes one row can at most spans in two Blocks.
+    - Retrieves the next 16MB after self._block and reads the first line. It
```
kemp-google: Why the change here from 64K to 16MB?

gcp_variant_transforms/beam_io/bgzf.py
allieychen: Done.

gcp_variant_transforms/beam_io/bgzf.py
allieychen: Done.

gcp_variant_transforms/beam_io/bgzf.py
allieychen: I have updated the doc to explain where the 16 MB comes from.The bug occurs when the record spans more than one GZIP block (max size 64 KB). 16 MB is the default read size for compressed files. I think it should be safe enough to cover one record (in compressed form).

PR #511, opened 2019-07-22
Create call info table

allieychen:
- Use flag --generate_call_info_table to form a call_info table.
- call_info table contains call_id, call_name, file_path.


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -145,6 +145,14 @@ def add_arguments(self, parser):
     parser.add_argument('--output_table',
                         default='',
                         help='BigQuery table to store the results.')
+    parser.add_argument(
+        '--schema_version',
+        type=int, default=0,
```
tneymanov: nit: how about making default 1 and the new one 2?

Even better would be to use a proper enum for this, rather than just digits - like for example 'Schema.V1' and 'Schema.V2' or something of that sort.

gcp_variant_transforms/options/variant_transform_options_test.py
```diff
@@ -93,7 +93,8 @@ def _make_args(self, args):
 
   def test_valid_table_path(self):
     args = self._make_args(['--append',
-                            '--output_table', 'project:dataset.table'])
+                            '--output_table', 'project:dataset.table',
```
tneymanov: I think we should improve this test and verify somewhere that the code actually works - for example, we can add a test to catch an error when "<table_id>_call_info" actually exists.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -154,6 +156,21 @@ def _get_file_names(input_file):
     return contents
 
 
+def create_file_path_to_file_hash_map(input_patterns, project):
```
tneymanov: As discussed offline, this doesn't support local files. I agree with your assessment that at least a graceful failure is required in case that a local input pattern had been supplied.

@kemp-google we should decide whether we want to provide support for the local files.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -378,6 +379,53 @@ def _run_annotation_pipeline(known_args, pipeline_args):
...
+                                                file_path_to_file_hash,
+                                                known_args.append))
+
+
+def _write_to_bigquery_v1(pipeline,  # type: beam.Pipeline
+                          pipeline_mode,  # type: PipelineModes
+                          known_args,  # type: argparse.Namespace
+                          file_path_to_file_hash  # type: Dict[str, str]
+                         ):
+  _create_call_info_table(pipeline, pipeline_mode, known_args,
```
tneymanov: Why do we need a 'sub-function' here? Will it be removed in the future?

gcp_variant_transforms/pipeline_common.py
```diff
@@ -154,6 +156,21 @@ def _get_file_names(input_file):
...
+  # type: (List[str], str) -> Dict[str, str]
+  """Returns a mapping from file path to its MD5 hash."""
+  matches = filesystems.FileSystems.match(input_patterns)
+  all_files = [metadata.path for match in matches
+               for metadata in match.metadata_list]
+  file_path_to_file_hash = {}
+  for input_file in all_files:
+    bucket_name, blob_name = gcsio.parse_gcs_path(input_file)
+    bucket = storage.Client(project).get_bucket(bucket_name)
```
tneymanov: Also, as discussed online, I attempted to run this logic with my service account on 'gcp-variant-transforms-test' on the following files:
- gs://gcp-variant-transforms-testfiles/small_tests/valid-4.2.vcf
- gs://gcp-variant-transforms-testfiles/small_tests/valid-4.0.vcf 

The bucket's data is public so I can access the files, but storage.Client(project) command fails on me, presumably because my service account is not associated with the project.

This is probably going to be an occurring scenario for our customers, as I would presume they would want to use public datasets, without being directly involved with the project hosting them (e.g. 1k genomes).

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -77,6 +78,7 @@ def __init__(self,
       alts: A dictionary mapping alt keys to vcf alt metadata values.
       formats: A dictionary mapping format keys to vcf format metadata values.
       contigs: A dictionary mapping contig keys to vcf contig metadata values.
+      samples: A list of sample names.
       file_name: The file name of the vcf file.
```
tneymanov: As discussed offline, given that this actually encapsulates 'gs:/<bucket>/<pseudo-folder>file_name', lets rename it to file_path.

gcp_variant_transforms/pipeline_common.py
allieychen: Thanks so much for helping testing it! Yes, the `get_bucket` would require bucket level permissions, while the user might only have the object-level permission. Fixed.

gcp_variant_transforms/vcf_to_bq.py
allieychen: No, it won't be removed in the future. Instead, more functions (`_create_call_table`, `create_variant_table`, `create_variant_meta_table`, etc.) will be added into `_write_to_bigquery_v1`.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -464,20 +522,14 @@ def run(argv=None):
...
-               append=known_args.append,
-               update_schema_on_append=known_args.update_schema_on_append,
-               allow_incompatible_records=known_args.allow_incompatible_records,
-               omit_empty_sample_calls=known_args.omit_empty_sample_calls,
-               num_bigquery_write_shards=known_args.num_bigquery_write_shards,
-               null_numeric_value_replacement=(
-                   known_args.null_numeric_value_replacement)))
+      if (known_args.schema_version ==
+          variant_transform_options.BigQueryWriteOptions.V1):
+        _write_to_bigquery_v1(variants[i],
```
tneymanov: seems rather weird that v1 and v2 writes are so far apart in code... how come?

allieychen: Yes, you are absolutely right. When the other tables are added, they will be surely added here. Done, moved the v1 schema here.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -154,6 +156,24 @@ def _get_file_names(input_file):
...
+  # type: (List[str], str) -> Dict[str, str]
+  """Returns a mapping from file path to its MD5 hash."""
+  matches = filesystems.FileSystems.match(input_patterns)
+  all_files = [metadata.path for match in matches
+               for metadata in match.metadata_list]
+  file_path_to_file_hash = {}
+  for input_file in all_files:
+    if not input_file.startswith('gs://'):
+      raise ValueError('Only input files in GCS are supported for BigQuery'
+                       'schema_version v1.')
```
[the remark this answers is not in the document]
allieychen: Done. Thank you !

gcp_variant_transforms/libs/hashing_util.py
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+"""Generates hashing code for string."""
+
+import sys
+
+import farmhash
+
+
+def generate_int64_hash_code(string):
```
samanvp: I am thinking to make this function a bit more specialized:
`def generate_int64_hash_code(string, min_hash_value = 0, max_hash_value = sys.max_int):`

so that we can call it like `def generate_int64_hash_code('test', 0, 256)` which means I am expecting the hash values to be in the [0, 256) range.

Let's discuss this offline, meanwhile I found this useful:

http://www.cim.mcgill.ca/~langer/250/27-hashing.pdf

gcp_variant_transforms/pipeline_common.py
samanvp: If we hash the whole file name using our `hashing_util.generate_int64_hash_code` and do not rely on `file_blob.md5_hash` then don't need to limit this to GCS files only, right?
I think I prefer that because:
 * we are able to control the range of the hashed values (in case we need to perform partitioning)
 * we will be using farm_fingerprint hashing consistently every where.

gcp_variant_transforms/transforms/call_info_to_bigquery.py
```diff
@@ -0,0 +1,80 @@
...
+  """Extracts call info from `VcfHeader` and converts it to a BigQuery row."""
+
+  def __init__(self, file_path_to_file_hash):
+    # type: (Dict[str, str]) -> None
+    self._file_path_to_file_hash = file_path_to_file_hash
+
+  def process(self, vcf_header):
+    # type: (vcf_header_io.VcfHeader) -> Dict[str, Union[int, str]]
+    for sample in vcf_header.samples:
+      call_id = hashing_util.generate_int64_hash_code(
```
samanvp: We should make sure we don't repeat this with each row of VCF file.

setup.py
```diff
@@ -29,11 +29,8 @@
     # Nucleus needs uptodate protocol buffer compiler (protoc).
     'protobuf>=3.6.1',
     'mmh3<2.6',
-    # Need to explicitly install v<=1.14.0. apache-beam requires
-    # google-cloud-pubsub 0.39.1, which relies on google-cloud-core<0.30dev,
-    # >=0.29.1. google-cloud-storage also has requirements on google-cloud-core,
-    # and version 1.14.0 resolves the dependency conflicts.
-    'google-cloud-storage<=1.14.0'
+    'google-cloud-storage',
```
samanvp: Is this still needed?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -478,6 +492,8 @@ def run(argv=None):
                num_bigquery_write_shards=known_args.num_bigquery_write_shards,
                null_numeric_value_replacement=(
                    known_args.null_numeric_value_replacement)))
+      if known_args.generate_call_info_table:
+        _create_call_info_table(pipeline, pipeline_mode, known_args)
```
samanvp: Is this the best place to do this? How about somewhere like `_merge_headers` which we are reading headers anyways to avoid reading headers multiple times?

gcp_variant_transforms/pipeline_common.py
```diff
@@ -18,7 +18,7 @@
 PTransforms and writing the output.
 """
 
-from typing import List  # pylint: disable=unused-import
+from typing import Dict, List  # pylint: disable=unused-import
```
samanvp: Is this still needed?

gcp_variant_transforms/libs/hashing_util_test.py
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+import unittest
+
+from gcp_variant_transforms.libs import hashing_util
+
+
+class HashingUtilTest(unittest.TestCase):
+
+  def test_generate_int64_hash_code(self):
```
samanvp: Please add a few more tests with other values set for max_hash_value such as 1, 8, 1000, ...

gcp_variant_transforms/libs/call_info_table_schema_generator.py
```diff
@@ -0,0 +1,46 @@
...
+"""Generates call_info table schema."""
+
+from apache_beam.io.gcp.internal.clients import bigquery
+
+from gcp_variant_transforms.libs import bigquery_util
+
+CALL_ID = 'call_id'
+CALL_NAME = 'call_name'
+FILE_PATH = 'file_path'
+TABLE_SUFFIX = '_call_info'
```
samanvp: Do you think we should include "_" in the suffix? I am asking this because for other suffixes (such as "chr1" or "variants" and "non_variants") we need to join it with the main name with a "_" added between. That's why I kind a feel we better remove it from here and instead user "_" to join them.

gcp_variant_transforms/libs/hashing_util.py
```diff
@@ -0,0 +1,30 @@
...
+  # type: (str) -> int
+  """Generates a forever-fixed int64 hash code for `string`.
+
+  The hash code generated by farmhash.fingerprint64 is unsigned. The int64 field
```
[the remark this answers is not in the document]
samanvp: Hmmm, this is a bit concerning. I always assumed that given a same string as key we will get the same hash using either Python library or BQ implementation, but it seems not?!!
I think we need to dig a little bit deeper to find out what is the root cause of dependency between these two.

gcp_variant_transforms/vcf_to_bq.py
allieychen: I thought about reusing the code in `_merge_headers`. The thing is, if `--representative_header_file` is supplied, we won't merge headers. I feel it won't be clean if this step is also added there (a bunch of if-else cases have to be added). Also, merge_headers is a separate pipeline (different dataflow job), it generates the `representative_header_file` for the main pipeline (vcf to bq). If we also write the meta info to BQ in merge_headers, the logic is not clean. Even worse, imaging some failures happen in the main pipeline, it would be difficult to roll back the changes for this table to keep appending for all tables as an atomic operation. Does this make sense to you?

setup.py
allieychen: Yes, we need this package for GCS usage.

gcp_variant_transforms/libs/call_info_table_schema_generator.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2019 Google LLC.
```
samanvp: In VCF specs, it seems instead of "call_info" they used "samples" or "sample_ids" to refer to them. Do you think it would be more clear if we rename all new modules and classes with one of those two?

gcp_variant_transforms/libs/hashing_util.py
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+
+"""Generates hashing code for string."""
+
+import sys
+
+import farmhash
+
+
+def generate_unsigned_hash_code(string, max_hash_value=sys.maxint):
```
samanvp: Could we modify this function to receive a list of strings and then serialize them as explained [here](https://jameshfisher.com/2018/01/09/how-to-hash-multiple-values/)?

gcp_variant_transforms/transforms/call_info_to_bigquery.py
```diff
@@ -0,0 +1,70 @@
...
+  def __init__(self, output_table_prefix, append=False):
+    # type: (str, Dict[str, str], bool) -> None
+    """Initializes the transform.
+
+    Args:
+      output_table_prefix: The prefix of the output BigQuery table.
+      append: If true, existing records in output_table will not be
+        overwritten. New records will be appended to those that already exist.
+    """
+    self._output_table = ''.join([
```
samanvp: I am thinking about moving this logic to a separate module that it's job is to generate table name based on the table_prefix and suffixes for different output tables. We need this logic for the following tables:
* Per chromosome table
* Variants vs non-variants tables
* partitioned and clustered by calls
* and this samples table

I am thinking we can move all the suffix consts to that module (including TABLE_SUFFIX = '_call_info' in call_info_table_schema.py) to that module and also the logic of adding underscore characters between name segments. That way user can modify only one spot to change the naming logic if they wanted to.

Please let me know what you think.

allieychen:
@kemp-google PTAL, thanks :)

gcp_variant_transforms/libs/call_info_table_schema_generator.py
kemp-google: I definitely would like to see us move to sample over time :)

So naming this call_info now might be an error regardless of what we call it internally (eg renaming the classes can come later)

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -144,6 +145,16 @@ def add_arguments(self, parser):
     parser.add_argument('--output_table',
                         default='',
                         help='BigQuery table to store the results.')
+    parser.add_argument(
+        '--generate_call_info_table',
```
kemp-google: Why are we defaulting this off?  I would rather see this on by default.

allieychen: Yes, ultimately this should be on. We have a follow-up PR that will modify the main table accordingly (e.g., replace the call_name with the sample_id), otherwise the sample_info table alone is not useful. I think we probably should turn on this flag after that.

gcp_variant_transforms/transforms/sample_info_to_bigquery.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class ConvertSampleInfoToRow(beam.DoFn):
+  """Extracts sample info from `VcfHeader` and converts it to a BigQuery row."""
+
+  def process(self, vcf_header):
+    # type: (vcf_header_io.VcfHeader) -> Dict[str, Union[int, str]]
+    for sample in vcf_header.samples:
+      sample_id = hashing_util.generate_unsigned_hash_code(
+          [vcf_header.file_path, sample], max_hash_value=pow(2, 63))
```
samanvp: Let's make the max hash value `pow(2, 63) - 1` to make sure we do not reach the max INT64 value. Also, probably we need to define this value as a const in `sample_info_table_schema_generator`.

allieychen: Since we used % in generate_unsigned_hash_code, I don't think we will reach the max value, right?

gcp_variant_transforms/transforms/sample_info_to_bigquery.py
```diff
@@ -0,0 +1,70 @@
...
+    # type: (str, Dict[str, str], bool) -> None
+    """Initializes the transform.
+
+    Args:
+      output_table_prefix: The prefix of the output BigQuery table.
+      append: If true, existing records in output_table will not be
+        overwritten. New records will be appended to those that already exist.
+    """
+    self._output_table = '_'.join([
+        output_table_prefix, sample_info_table_schema_generator.TABLE_SUFFIX])
```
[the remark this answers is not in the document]
allieychen: Done. I moved this into `sample_info_table_schema_generator`. I think `pipeline_common` is more about common pipelines. In the future, if we have more tables, we may put all the schema in something like `table_schema_generator` (renamed from `sample_info_table_schema_generator`).

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -144,6 +145,16 @@ def add_arguments(self, parser):
     parser.add_argument('--output_table',
                         default='',
                         help='BigQuery table to store the results.')
+    parser.add_argument(
+        '--generate_sample_info_table',
```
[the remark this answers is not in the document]
allieychen: Ah, right. I will leave this to future PR to simplify this one. Thanks!

PR #512, opened 2019-07-22
Refactor the code to prepare for the new schema

allieychen:


allieychen:
@tneymanov Thank you Tural. The script `bigquery_vcf_data_converter.py` contains two classes, one converts Variant to BQ row, the other converts BQ row to Variant (used in BQ to VCF pipeline). If we are changing the name, we need to extract the other class to another script. I can imagine in the future, if we are supporting BQ to VCF for the new schema, similar converting from BQ row to some internal objects would apply as well.

tneymanov:
Ahh, got it. No, I don't think there is a need to do that right now.

allieychen:
@tneymanov I have been thinking about the `bigquery_row_generator` script, I think actually it would be more neat if we move all classes that generating bq rows (variant row, call row, variant meta data row, etc.) into this script. So I further refactored the code and moved `VariantCallRowGenerator` in this script. PTAL. :)

allieychen:
Thank you Tural!

Issue #504, opened 2019-07-16
FR: Support running the annotations pipeline standalone


mbookman:
It would be great to be able to `----run_annotation_pipeline` as a standalone step. Some users might not want to upload to BigQuery or you might want to operationally separate the annotation step from the upload to BigQuery.

Today you can make this happen by causing the load to BigQuery to fail. For example, you can pass a non-existent VCF header, like `--representative_header_file DOES_NOT_EXIST`.

PR #516, opened 2019-08-08
run annotation pipeline as a standalone tool

allieychen:
This PR allows the user to run annotation as a standalone tool: if --run_annotation_pipeline is set and --output_table is not provided.

FR: https://github.com/googlegenomics/gcp-variant-transforms/issues/504

### DIFF
--- a/gcp_variant_transforms/options/variant_transform_options.py
+++ b/gcp_variant_transforms/options/variant_transform_options.py
@@ -129,12 +129,6 @@ class AvroWriteOptions(VariantTransformsOptions):
                         default='',
                         help='The output path to write Avro files under.')
 
-  def validate(self, parsed_args):
-    # type: (argparse.Namespace) -> None
-    if not parsed_args.output_table and not parsed_args.output_avro_path:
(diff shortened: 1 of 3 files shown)


kemp-google:
What is the reason for this change?  Given our strategy of moving to storing annotation tables in BigQuery and joining post-hoc I'm not sure it makes sense?

allieychen:
Matt has requested this feature #504. Somethings he only wants to do annotation separately rather than loading into BigQuery.

mbookman:
Any chance this PR could be resurrected? The desire to do annotations as a separate step from the load to BigQuery still exists. The reason for this is for discrete, debuggable, restartable units of work. We have had issues annotating thousands of files where some shards fail and require investigation, and then we only want to re-run the annotations work on those failed shards. It is best operationally if this is a separate step. The annotated VCFs are also of value on their own, even though we do also load them into BigQuery.

tneymanov:
Hmmm, Interesting. This feature doesn't seem to be especially challenging. I'll bring it up in our next team meeting, but even if it's green lit it won't be implemented for a while, due to other more urgent commitments (e.g. Python3 migration).

PR #524, opened 2019-10-31
Fix our build

samanvp:
It seems our build is broken due to the following reasons:
 * Newer cloud-sdk:slim image requires us to replace all `pip install` commands with `python -m pip install`.
 * Updating from BEAM 2.14 to 2.16 requires us to send `--region` to all pipeline executions.

### DIFF
--- a/run_presubmit.sh
+++ b/run_presubmit.sh
@@ -20,13 +20,13 @@ set -euo pipefail
 echo ========== Running unit tests.
 if [[ -z `which coverage` ]];then
   echo "coverage is not installed. Installing ..."
-  pip install coverage
+  python -m pip install coverage
 fi
 coverage run --source=gcp_variant_transforms setup.py test
 
 echo ========== Running pylint.
 if [[ -z `which pylint` ]];then
   echo "pylint is not installed. Installing ..."
-  pip install pylint
+  python -m pip install pylint
 fi
 pylint gcp_variant_transforms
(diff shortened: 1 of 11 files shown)


run_presubmit.sh
```diff
@@ -20,13 +20,13 @@ set -euo pipefail
...
 fi
 coverage run --source=gcp_variant_transforms setup.py test
 
 echo ========== Running pylint.
 if [[ -z `which pylint` ]];then
   echo "pylint is not installed. Installing ..."
-  pip install pylint
+  python -m pip install pylint
 fi
 pylint gcp_variant_transforms
```
tneymanov: This may not be needed, but for consistency, maybe modify this to python -m pylint?

samanvp: done.

setup.py
```diff
@@ -33,7 +33,7 @@
     # google-cloud-pubsub 0.39.1, which relies on google-cloud-core<0.30dev,
```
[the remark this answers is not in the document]
samanvp: done.

run_presubmit.sh
```diff
@@ -20,13 +20,13 @@ set -euo pipefail
 echo ========== Running unit tests.
 if [[ -z `which coverage` ]];then
```
kemp-google: Just a general question: is it relatively fast to just install the tools required?  eg, if this were apt, I'd just install everything rather than check/install.

Not something to change in this PR, just a general question about pip.

PR #526, opened 2019-11-12
Add a flag to include file in the sample_id hash

samanvp:
For some inputs such as 1000Genome where data is split by chromosome, if we use `sample_id farmhash([file, sample_name])` we will end up with 23 `sample_id`s for a single sample.
By introducing this new flag we let users to decide whether or not this is the intended use.


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -155,6 +155,12 @@ def add_arguments(self, parser):
               '[EXPERIMENTAL]'
              ).format(sample_info_table_schema_generator.TABLE_SUFFIX))
 
+    parser.add_argument(
+        '--include_file_in_sample_id',
```
kemp-google: In general I'd rather we use flags that convey a meaning rather than a specific implementation detail, for two specific reasons:
1. It might be more meaningful to the user
2. We can reuse this information in other contexts

In this case, the real question is whether a single sample spans multiple files.  So something like --samples_span_multiple_files might be more appropriate.

What do you think?

samanvp: Thanks Aaron, I modified the flag name.

PR #532, opened 2019-12-13
Remove PyVCF and Nucleous.

tneymanov:
- Fully remove non-pysam functionality from GVT codebase.
- Rework vcf_header_io to use Pysam header reader in tests.
- Make sure to not use headers supplied when infer_header functionality is used.


gcp_variant_transforms/libs/vcf_header_parser.py
```diff
@@ -104,11 +77,13 @@ def get_metadata_header_lines(input_file):
 def _line_generator(file_name):
```
samanvp: `_header_line_generator()` might be a better name.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -301,18 +296,17 @@ def __init__(self,
...
-        secondary_tracker = range_trackers.UnsplittableRangeTracker(
-            copy(range_tracker._range_tracker))
-      else:
-        secondary_tracker = copy(range_tracker)
-      self._verify_header(copy(text_source), secondary_tracker)
+    # In an absence of "#CHROM POS..." header line, PySam process would lock.
+    # However, we need to support cases when no records in files were provided
+    # even if the above header line is missing. To circumvent the error, read
+    # the file until the first line is found, and verify validity.
+    self._header_lines = None
```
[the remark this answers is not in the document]
tneymanov: Because once you run _verify_header() with secondary tracker, you would populate it for Pysam. Now it's obsolete, since I've removed _verify_header()

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -301,18 +296,17 @@ def __init__(self,
...
+    if isinstance(range_tracker, range_trackers.UnsplittableRangeTracker):
+      secondary_tracker = range_trackers.UnsplittableRangeTracker(
+          copy(range_tracker._range_tracker))
+    else:
+      secondary_tracker = copy(range_tracker)
+    self._verify_header(copy(text_source), secondary_tracker)
```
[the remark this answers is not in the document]
tneymanov: Not really, since it has to happen before actual read has happened.

However, this method is not really needed, since now vcf_header_io, which always happens before reading actual variants does the same. I was going to do it later, but since it's creating points of contention and especially since it has negative affect on the performance, I've removed it.

Also moved a test to vcf_header_io_test, but it needs expanding to encapsulate more cases. Added a TODO there.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -301,18 +296,17 @@ def __init__(self,
...
 
     self._text_lines = text_source.read_records(self._file_name,
                                                 range_tracker)
```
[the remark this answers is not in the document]
tneymanov: This whole class needs to be reworked to only have a single class (since we are removing other parses completely), but I was going to do it later. Would you like me to do it in this PR?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -353,7 +347,7 @@ def _verify_header(self, text_source, range_tracker):
         # ID, Description, Type and Number are mandatory fields.
         if not k:
           raise ValueError('Corrupt ID at header line {}.'.format(v.id))
-        if not v.description:
+        if v.description is None:
```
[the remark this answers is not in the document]
tneymanov: Not really - I wanted to redo it later, but, as mentioned in another comment, since it hurts performance, removed it.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -383,19 +377,24 @@ def _process_header_lines(self, header_lines):
...
+          0, FILE_FORMAT_HEADER_TEMPLATE.format(VERSION='4.0'))
+    header_lines = [line.strip() for line in header_lines if line.strip()]
+    self._header_lines = []
+    for line in header_lines:
       # Number='G' is to be deprecated and is unsupported by PySam - replace
       # with 'unknown' number identifier if found.
-      self._header_lines = filter(None, [line.strip().replace(
-          'Number=G', 'Number=.').encode('utf-8') for line in header_lines])
-    else:
-      self._init_with_header(header_lines)
```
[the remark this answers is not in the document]
tneymanov: This functionality is reinstated, but would not be really needed if we remove base/sub-class. If you want me to remove it in this PR, I'll rework this part as well.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -404,7 +403,10 @@ def next(self):
       # clean up, once iterator is depleted.
       self.after_read()
       raise e
-    record = self._get_variant(text_line)
+    try:
+      record = self._get_variant(text_line)
+    except Exception as e:
+      raise ValueError('asd')
```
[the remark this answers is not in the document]
tneymanov: Oops, artifact of debugging. Removed.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -815,187 +610,22 @@ def _get_variant_calls(self, samples):
...
-      data = self._convert_list_value(variant_proto.info[k],
-                                      self._is_info_repeated(k))
-      # Avoid including missing flags as `false` or other fields valued as `[]`.
-      if not data:
+  def _convert_list(self, data):
+    cast_type = self._get_cast_type(data)
+    parsed_data = []
+    for elem in data:
+      parsed_data.append(None if elem is None else cast_type(elem))
+    return parsed_data
```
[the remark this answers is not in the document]
tneymanov: Done with certain adjustments - added None's and made sure not to force all values to be float and allow ints as well.

FYI, although type of None is NoneType, it doesn't exist as a var, so I have to extract it.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -646,14 +434,19 @@ def _init_parent_process(self, pysam_read, variants_write):
...
     # Child process' task is to populate data into the pipe that feeds
     # VariantFile class - first by populating all of the header lines, and then
     # by redirecting the data received from the second pipe.
 
     # Write Header Lines into PySam.
     to_pysam_pipe = os.fdopen(pysam_write, 'w')
-    to_pysam_pipe.write('\n'.join(header_lines) + '\n')
+    if pre_infer_headers:
+      to_pysam_pipe.write(header_lines[0] + '\n') # fileformat line
+      to_pysam_pipe.write(header_lines[-1] + '\n') # `#CHROM...` line
```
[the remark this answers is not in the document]
tneymanov: If I understood your comment correctly, I have to disagre, since I don't think we can use them as consts - first line defines the format and last line is CHROM line, that specifies samples.

The existence of first line is forced in this file, since if we cannot find it, we add it ourselves. The existence of last line is also required as otherwise it would break pysam anyways - it is checked in vcf_header_io in the verify_header() method (that was removed from here in this PR).

Please correct if I misunderstood you.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -131,30 +132,24 @@ def __repr__(self):
...
   def _get_infos_pysam(self, infos):
+    self.verify_infos(infos)
     results = collections.OrderedDict()
     for item in infos.items():
       result = collections.OrderedDict()
       result['id'] = item[0]
-      result['num'] = item[1].number if item[1].number != '.' else None
-      result['type'] = item[1].type
+      result['num'] = item[1].number
+      result['type'] = item[1].record['Type']
```
samanvp: Let's add enough checks making sure this field exist, similar to what you do below for 'source' and 'version'. Also please make the same change to `_get_formats_pysam()` function.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -131,30 +132,24 @@ def __repr__(self):
...
       result['desc'] = item[1].description
       # Pysam doesn't return these fields in info
-      result['source'] = None
-      result['version'] = None
+      result['source'] = (item[1].record['Source'].strip("\"")
```
samanvp: Is it possible that `item[1].record` is missing altogether and accessing it causes an exception?

Also it seems every time we access a field in `record' we need to perform these checks. So perhaps we can defined a method to do this operation:

`_read_field_from_record()`

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -131,30 +132,24 @@ def __repr__(self):
...
-      # https://docs.python.org/2/library/collections.html#collections.namedtuple
-      ordered_dict[key] = header[key]._asdict()  # pylint: disable=W0212
-    return ordered_dict
-
   def _get_infos_pysam(self, infos):
+    self.verify_infos(infos)
     results = collections.OrderedDict()
     for item in infos.items():
       result = collections.OrderedDict()
       result['id'] = item[0]
```
samanvp: Let's assign `item[0]` to a variable with a more clear name, how about this:
```
id = item[0]
result['id'] = id
...
results[id] = result
```

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -131,30 +132,24 @@ def __repr__(self):
...
+                          item[1].record['Source'] is not None else None)
+      result['version'] = (item[1].record['Version'].strip("\"")
+                           if 'Version' in item[1].record and
+                           item[1].record['Version'] is not None else None)
       results[item[0]] = result
-    return dict(results.items())
+    return results
 
   def _get_filters_pysam(self, filters):
     results = collections.OrderedDict()
```
samanvp: Similar to `_get_info_pysam()` let's assign `item[0]` to a meaningful variable. 
Also we could check the 'PASS' in this look and skip those items instead of deleting them later"
```
id = item[0]
if id == 'PASS'
  continue
result['id'] = id
...
results[id] = result
```

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -174,18 +170,18 @@ def _get_alts_pysam(self, alts):
       result['id'] = item[0]
```
samanvp: Again, please add `id = item[0]`.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -174,18 +170,18 @@ def _get_alts_pysam(self, alts):
...
       result['desc'] = item[1]['Description'].strip("\"")
```
samanvp: We sometimes we need to perform `strip("\"")`? I saw it a few other times in this module?
Is it just because of passing our integration tests? If that's the case we better modify our tests and remove all these extra operations.
Not needed for this PR, I added an item to [the issue](#534) that tracks these future clean ups.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -194,7 +190,7 @@ def _get_contigs_pysam(self, contigs):
       result['id'] = item[0]
```
samanvp: Again, please add `id = item[0]`.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -194,7 +190,7 @@ def _get_contigs_pysam(self, contigs):
...
       result['length'] = item[1].length
       results[item[0]] = result
-    return dict(results.items())
+    return results
 
   def _get_samples_pysam(self, sample_line):
     sample_tags = sample_line.split('\t')
```
samanvp: We should explain where this const 9 is coming from, either add a comment or use this variable:
```
default_items_num = len(LAST_HEADER_LINE_PREFIX.split('\t')
if len(sample_tags) > default_items_num:
  return sample_tags[(default_items_num):]
```
Also shouldn't this be `> 8`? `LAST_HEADER_LINE_PREFIX` has 8 fields.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -203,58 +199,74 @@ def _get_samples_pysam(self, sample_line):
     else:
       return []
 
+  def verify_infos(self, fields):
```
samanvp: Please add description and type hint of input args.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -203,58 +199,74 @@ def _get_samples_pysam(self, sample_line):
...
+      # ID, Description, Type and Number are mandatory fields.
+      if not k:
+        raise ValueError('Corrupt ID at header line {}.'.format(v.id))
+      if v.description is None:
+        raise ValueError(
+            'Corrupt Description at header line {}.'.format(v.id))
+      # Type can only be Integer, Float, Flag, Character or String
+      if not v.type or v.record['Type'] not in HEADER_TYPES:
+        raise ValueError('Corrupt Type at header line {}'.format(v.id))
+      # Number can only be a number or one of 'A', 'R', 'G' and '.'.
```
samanvp: Shouldn't we verify `v.number` instead of `v.record['Number']`

Also as a more general question, here do we need to verify `record` at all?  My intuition is that we are verifying the mandatory fields which we expect to find in `pysam.libcbcf.VariantHeader` and there is no reason to verify `pysam.libcbcf.VariantHeaderRecord`. Please correct me if I am wrong.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -263,34 +275,7 @@ def _read_headers(self, file_path):
...
-        break
-    yield VcfHeader(infos=header.info,
-                    filters=header.filters,
-                    alts=header.alts,
-                    formats=header.formats,
-                    contigs=header.contigs,
-                    samples=sample_line,
-                    file_path=file_path,
-                    vcf_parser=vcfio.VcfParserType.PYSAM)
+      yield record.strip()
```
samanvp: Why did you add this line? Before adding this line this method does what its name claims, returning header lines. but now it also returns the first none header line!

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -303,15 +288,14 @@ def open_file(self, file_path):
 class ReadVcfHeaders(PTransform):
   """A PTransform for reading the header lines of VCF files.
 
-  Parses VCF files (version 4) using PyVCF library.
+  Parses VCF files.
```
samanvp: Remove this line.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -331,20 +315,17 @@ def __init__(
...
-        validate=validate,
-        vcf_parser=vcf_parser)
+        validate=validate)
 
   def expand(self, pvalue):
     return pvalue.pipeline | Read(self._source)
 
 
-def _create_vcf_header_source(
+def CreateVcfHeaderSource(
```
samanvp: Why did you rename this method? This new name does not follow our naming convention (ie its name looks like a class name).

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -526,29 +505,21 @@ def _format_header_key(self, key):
...
     Raises:
       ValueError: if the number is not valid.
     """
     if number is None:
       return None
-    elif number >= 0:
+    if number >= 0:
       return str(number)
-    if number in VCF_HEADER_INFO_NUM_FIELD_CONVERSION:
-      return VCF_HEADER_INFO_NUM_FIELD_CONVERSION[number]
```
samanvp: Deleting this extra condition changes the logic, ie without this condition 'A', 'G', and 'R' will raise an exception while it shouldn't!
Here is what I propose:
```
VALID_VCF_HEADER_INFO_NUM_FIELD = [
    'A', # value when number of alternate alleles dictates the count
    'G', # value when number of genotypes dictates the count
    'R', # value when number of alleles and ref dictates the count
]
...
if number in VALID_VCF_HEADER_INFO_NUM_FIELD:
  return number
``` 

If I am right, why our tests did not catch this then? Should we add a test to check this? I already added an item to the [list of issues](#534).

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -203,58 +199,74 @@ def _get_samples_pysam(self, sample_line):
...
+      if line.startswith('##'):
+        header.add_line(line.strip())
+      elif line.startswith('#'):
+        sample_line = line
+      elif line:
+        if not sample_line:
+          raise ValueError('Header line is missing')
+      else:
+        if not sample_line:
+          sample_line = LAST_HEADER_LINE_PREFIX
```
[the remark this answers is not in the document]
tneymanov: So the idea is to handle the case when header exists without records (in which case, CHROM line is not required).

_read_headers() attempts to return 1 record line at the end, if present (empty string if no record lines are present). 

If there is at least one record line, #CHROM line is mandatory and its absence is an error. If there are no record lines, uses a dummy #CHROM line when extracting samples (which there will be 0 of).

Added comments to make it more clear.

gcp_variant_transforms/beam_io/vcf_header_io.py
tneymanov: Yeah, it's for verification of the #CHROM line - if records exist, #CHROM line is mandatory and its absence would break pysam downstream (in vcf_parser). If no record lines are present, we use dummy #CHROM line to "extract samples" (which there will be None).

Naming is not perfect, but I cannot think of a better one. Added a comment for clarification.

gcp_variant_transforms/beam_io/vcf_header_io.py
tneymanov: Done.

gcp_variant_transforms/beam_io/vcf_header_io.py
tneymanov: Ummm, pylint forced me to do it :( . Are our naming conventions different from pylint ones? That doesn't seem right...

> Function name "create_vcf_header_source" doesn't conform to u'^(?:(?P<camel_case>_?[A-Z][a-zA-Z0-9]*)|(?P<snake_case>_?[a-z][a-z0-9_]*))$' pattern

gcp_variant_transforms/beam_io/vcf_header_io.py
tneymanov: Umm, so 'A' >= 0 is True in python. PyVCF had these values as negatives (ie 'A' was -1, 'G' was -2 I think...) so this check was needed to convert integers to strings, and -1 ... -3 to A/G/R. So from that point, this code works fine (it turns strings AND ints to string, so no issue there). However, it should raise an error for non-A/G/R strings, which it wouldn't at them moment, as 'any string' >=0 is True.

Used HEADER_NUMBER instead (since number can be '.' as well. Made to check that number is int. Also removed unnecessary None check - that PySAM returns '.' instead of None.

Also, I'm leaving numer >=0 since we have a test for that, although I'm not sure if it's ever possible to have our vcf header with number < 0 - in theory, I don't think that should ever happen, but just in case, left that condition.

gcp_variant_transforms/beam_io/vcf_header_io.py
tneymanov: No, it's because Header fields have for strings come with quotes (ie. Description="Some Description"), this operation is very much needed.

Record gets everything after '=', so we get record['Description']='"SomeDescription"'.

Also, I verified, VCF file format says that they specifically should be double quotes, never single quotes.

gcp_variant_transforms/beam_io/vcf_header_io.py
tneymanov: Done through for loop.

gcp_variant_transforms/beam_io/vcf_header_io.py
tneymanov: When there are no samples, there are 8 fields. When there are samples, there is also FORMAT field that describes call formats.

Used your suggestion, added +1, added a comment for explanation.

gcp_variant_transforms/beam_io/vcf_header_io.py
```diff
@@ -35,6 +34,8 @@
 from gcp_variant_transforms.beam_io import vcfio
 
 LAST_HEADER_LINE_PREFIX = '#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO'
+HEADER_TYPES = ['Integer', 'Float', 'Flag', 'Character', 'String', '.']
```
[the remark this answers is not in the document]
samanvp: Sounds good. I added an action item to #534 please feel free to add some of the details in your reply to that item so in future we will remember the context. After that mark this as resolved.

gcp_variant_transforms/libs/vcf_reserved_fields.py
```diff
@@ -22,38 +22,36 @@
...
 def _get_field_count(value):
   # type: (str) -> Optional[int]
-  return parser.field_counts[value]
+  return value
 
 
 INFO_FIELDS = {
     'AA': _ReservedDefinition('AA', 1, 'String', 'Ancestral allele'),
-    'AC': _ReservedDefinition('AC', _get_field_count('A'), 'Integer',
+    'AC': _ReservedDefinition('AC', 'A', 'Integer',
```
samanvp: Let's define the following 3 consts, again these are duplicated definitions but later we can unify them:
```
_FIELD_COUNT_ALTERNATE_ALLELE = 'A'
_FIELD_COUNT_ALL_ALLELE = 'R'
_FIELD_COUNT_GENOTYPE = 'G'
```

gcp_variant_transforms/testing/vcf_header_util.py
```diff
@@ -16,28 +16,17 @@
...
-
   Args:
     header_num_dict: a dictionary mapping info keys to string num values.
   """
   infos = {}
   for k, v in header_num_dict.iteritems():
-    if v in parser.field_counts:
-      pyvcf_num_field_value = parser.field_counts[v]
-    else:
-      pyvcf_num_field_value = int(v)
```
samanvp: What is the type of `v`?
Don't we need to follow the old logic and convert it to int? Here is what I think it should be:
```
num_field_value = v if v in HEADER_NUMBERS else int(v)
infos[k] = vcf_header_io.CreateInfoField(k, num_field_value, '.', '')
```

gcp_variant_transforms/beam_io/vcf_header_io.py
tneymanov: _Cannot comment on the other convo unfortunately, will reply here._

So the problem isn't the existence or absence of samples, but of the CHROM line in general - if VariantFile is instantiated without it, it would hang and wait for the CHROM line.

Unfortunately, there does exist a case where people dont' supply it - particularly, if they have records in multiple files, and the headers separately. This is an actual case, that (I think) Matt had encountered and described to Allie, which forced me to add these precautions - although I don't remember it perfectly, I think he had files with records that started with CHROM line and no headers, but there was also at least 1 file with only header metadata.

Now that I think about, I think me removing the check from vcf_parser and putting it into header would correctly catch the case when CHROM line is missing but records are existent, but would now fail to prevent the case I described above... Investigating it now, will need to run some tests. In the meanwhile sending the comments so you could continue reviewing.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -802,10 +601,12 @@ def _get_variant_calls(self, samples):
...
         else:
-          info[key] = (list(map(self._convert_field, value)) if
-                       isinstance(value, tuple) else self._convert_field(value))
+          info[key] = (self._homogenize_list_dtype(
+              map(self._convert_field, value, [True] * (len(value)))) if
+                       isinstance(value, tuple) else
+                       self._convert_field(value, True))
 
       # PySam samples are "phased" for haploids, so check for for the type
       # before settings default phaseset value.
```
samanvp: In the following if statement, do we need the last condition `len(genotype) > 1`? It seems `sample.phased` includes that condition:
https://github.com/pysam-developers/pysam/blob/430c883ecd5f04080d00aa950ebd259aa38f40bc/pysam/libcbcf.pyx#L3420

gcp_variant_transforms/libs/vcf_reserved_fields.py
tneymanov: Ack.

gcp_variant_transforms/libs/vcf_reserved_fields.py
```diff
@@ -22,38 +22,36 @@
...
                               'Allele count in genotypes, for each ALT allele, '
                               'in the same order as listed'),
-    'AD': _ReservedDefinition('AD', _get_field_count('R'), 'Integer',
+    'AD': _ReservedDefinition('AD', 'R', 'Integer',
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -726,7 +520,7 @@ def _convert_to_variant(self, record):
         reference_name=record.chrom.encode('utf-8'),
         start=record.start,
         end=record.stop,
-        reference_bases=self._convert_field(record.ref),
+        reference_bases=self._convert_field(record.ref, True),
```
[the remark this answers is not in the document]
samanvp: Please not I am specifically asking about converting `record.ref` to `int`. Isn't reference always string of 4 base pairs (`'ACGT'`)? Unless there is some corner cases that we might have numbers in this field.

Could you please make sure for PyVcf we do the same conversion for this field?

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -738,48 +538,62 @@ def _convert_to_variant(self, record):
...
         else:
-          # If a field was not provided in the header, make it a list to
-          # follow the PyVCF standards.
-          info[k] = (
-              self._convert_field(v) if k in self._original_info_list else
-              [self._convert_field(v)])
+          # If a field was not provided in the header, make it a list.
+          info[info_id] = (self._convert_field(field, True) if
+                           info_id in self._original_info_list else
+                           [self._convert_field(field, True)])
```
samanvp: Thanks for updating this method, code is more clear now.

Now my question is why we need to attempt to convert singular values (non tuples) to integer?

tneymanov:
Thanks Saman.

As discussed offline, it seems like our changes and reliability on naively selected header types made clear several issues with infer_headers logic that we were swallowing before, even in PyVCF version of the codebase. Had to fix them by making infer_headers_util more intelligent.

Seems like it has not affected integration tests, which is both good and bad news - this shows that our integration tests did not cover these errors as they were being looked over. They pass now, but we probably should have more robust tests in future PRs.


Also, I'm adding changes for fix in YAML and integration tests timeout here, so I don't have to rebase and change all of the commits - review comments would had be jumbled.

PR #536, opened 2020-01-23
Add Ingestion_datetime to our sample_info_table.

samanvp:
We got feedback from couple of our users about adding this column to our sample_info_table. We can use this column to partition the table as well.


gcp_variant_transforms/transforms/sample_info_to_bigquery.py
```diff
@@ -14,28 +14,42 @@
...
 
+from datetime import datetime
+
 import apache_beam as beam
 
 from gcp_variant_transforms.beam_io import vcf_header_io  # pylint: disable=unused-import
 from gcp_variant_transforms.libs import sample_info_table_schema_generator
 from gcp_variant_transforms.libs import hashing_util
 
+_DATETIME_FORMAT = '%Y-%m-%d %H:%M'
```
tneymanov: non-blocker: why was minute granularity chosen? Why not second or millsec? Is it BQ requirement?

samanvp: We discussed this offline. To summarize, we don't see any use case why granularity more than minute is needed. If that rises in future, we can easily address it because the column's type is `datetime`

PR #539, opened 2020-01-27
Store schema JSON in temp file on creation of BQ tables.

tneymanov:


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -496,7 +499,8 @@ def run(argv=None):
                omit_empty_sample_calls=known_args.omit_empty_sample_calls,
                num_bigquery_write_shards=known_args.num_bigquery_write_shards,
                null_numeric_value_replacement=(
-                   known_args.null_numeric_value_replacement)))
+                   known_args.null_numeric_value_replacement),
+               temp_schema_file_path=schema_file))
```
samanvp: I don't think we need to send the json file to `VariantToBigQery` class. Here is what I think we should do for new schema:
* in `vcf_to_bq` (which is simply running on the single monitoring worker) we have all components to create all output BQ tables using `bq make table ...`:
  * Similar to what we do in `VariantToBigQuery` we can find the schema of output table: `self._schema = schema_converter.generate_schema_from_header_fields(self._header_fields, self._proc_var_factory, self._variant_merger)`
  * Read input partition config file so we know exactly how many output tables we have. Also we get the suffix for each output table from that config file.
  * For each output table we make the full name `output_table = sample_info_table_schema_generator.compose_table_name(
        output_table_prefix, table_suffix)`
  * At this point depending on whether we are appending to existing tables or creating new ones, we will call `bq make table ... ` to create integer ranged partitioned tables.
  * We call `VariantToBigQery`to write rows to the already created tables. This module will always work in "append" state: either it appends to an empty table (that was just created) or it will append to a none empty table (that was created in previous runs of VT).

Note we don't need to do any of these actions in this PR, I just wanted to document what I have in mind.
Please let me know if all of this make sense; if so please revert changes of `VariantToBigQery` accordingly.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -479,6 +481,16 @@ def run(argv=None):
...
   if known_args.output_table:
+    schema_file = tempfile.mkstemp(prefix=known_args.output_table,
+                                   suffix=_BQ_SCHEMA_FILE_SUFFIX)[1]
+    schema = (
+        schema_converter.generate_schema_from_header_fields(
+            header_fields, processed_variant_factory, variant_merger))
+    schema_json = (
+        schema_converter.convert_table_schema_to_json_bq_schema(schema))
+    with filesystems.FileSystems.create(schema_file) as file_to_write:
+      file_to_write.write(schema_json)
```
samanvp: Doesn't a simple `json.dump()` do the same thing? Note this will be run locally on the main monitor machine.

tneymanov: Discussed offline. `json.dump()` is used to convert dict to json string, not to write to a file.

docker/Dockerfile
```diff
@@ -47,6 +47,7 @@ RUN apt-get update && apt-get install -y \
     zlib1g-dev
 
 # Install dependencies.
+
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/vcf_to_bq.py
tneymanov: Discussed offline.

In order to not recreate schema a second time, and to simplify `variant_to_bigquery.py` file, we will create schema in `vcf_to_bq.py` and push it to the above file as well as use it to generate a json temp file.

PR #540, opened 2020-01-28
Rename 'partitioning' to 'sharding'

samanvp:
Partition in BigQuery's terminology has a very clear meaning. We use the same word for the Variant Transforms' functionality that shards output tables, for example one output BQ table per chromosome.

To avoid confusion, we rename this feature of Variant Transforms to "sharding".


gcp_variant_transforms/libs/variant_sharding_test.py
```diff
@@ -0,0 +1,412 @@
...
+    self.assertEqual(sharder.get_shard('2', 0), 4)
+
+    self.assertEqual(sharder.get_shard('c4', 0), 4)
+    self.assertEqual(sharder.get_shard('cr5', 0), 4)
+    self.assertEqual(sharder.get_shard('chr6', 0), 4)
+
+  def test_config_failed_missing_region(self):
+    tempdir = temp_dir.TempDir()
+    missing_region = [
+        '-  partition:',
```
tneymanov: We still have partitions here, throughout the file.

Is it because we are not modifying the config file that gets generated?

samanvp: These are field names in the input YAML file:
https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/data/partition_configs/homo_sapiens_default.yaml

I decided not to modify those because we will replace that config file in the next PR.

PR #545, opened 2020-02-04
New sharding config

samanvp:
In this PR we have two main changes:
* We use a new sharding config file that includes information for BigQuery integer range partitioning.
* We always perform output table sharding as opposed to the optional sharding we used to offer using a flag.


gcp_variant_transforms/libs/variant_sharding.py
```diff
@@ -78,13 +74,13 @@ def add_region(self, start, end, shard_index):
...
-      raise ValueError(
-          'Cannot add overlapping region {}-{}'.format(start, end))
+      raise ValueError('Wrong sharding config file, regions must be unique in '
+                       'config file: {}-{}'.format(start, end))
     # If everything goes well we add the new region to the interval tree.
     self._interval_tree.addi(start, end, shard_index)
 
   def get_shard_index(self, pos=0):
-    """Finds a region that includes pos, if none _UNDEFINED_PARTITION_INDEX."""
+    """Finds a region that includes pos, if none _UNDEFINED_SHARD_INDEX."""
```
tneymanov: nit: I know it was wrong in the original version, but replace with "..., if none _return_ _UNDEFINED_SHARD_INDEX."

cloudbuild_CI.yaml
```diff
@@ -41,9 +41,6 @@ steps:
       - '--skip_build'
       - '--project ${PROJECT_ID}'
       - '--image_tag ${COMMIT_SHA}'
-      - '--run_unit_tests'
```
tneymanov: Return the tests :D

gcp_variant_transforms/libs/variant_sharding.py
```diff
@@ -28,32 +28,28 @@
...
-# [0, 22) are reserved for standard reference_names. Every other identifiers
-# will be matched to next available partitions [22, 27).
-_DEFAULT_NUM_SHARDS = _RESERVED_AUTO_SHARDS + 5
-
-# At most 1000 partitions can be set as output of VariantTransform.
-_MAX_NUM_SHARDS = 1000
-# Each partition can contain at most 64 regions.
-_MAX_NUM_REGIONS = 64
+# At most 100 shards (distinct tables) can be set as output of VariantTransform.
+_MAX_NUM_SHARDS = 100
```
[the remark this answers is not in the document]
samanvp: No reason, I just thought it's a bit unusual to have so many regions for one output table. If a user complain about this we can easily increase it but I really don't foresee any usecase for it.

gcp_variant_transforms/libs/variant_sharding.py
```diff
@@ -94,198 +90,234 @@ def get_shard_index(self, pos=0):
...
+    self._total_base_pairs = []
+
+    self._parse_config(config_file_path)
+
+  def _is_residual_shard(self, regions):
+    # type: (List[str]) -> bool
+    return (len(regions) == 1 and
+            regions[0].strip() == _RESIDUAL_REGION_LITERAL)
 
   def _validate_config(self, config_file_path):
```
tneymanov: nit: _validate_config_and_check_intervals maybe?

gcp_variant_transforms/libs/variant_sharding.py
```diff
@@ -94,198 +90,234 @@ def get_shard_index(self, pos=0):
...
+  def get_output_table_suffix(self, shard_index):
+    # type: (int) -> Optional[str]
+    if not self._is_index_in_the_range(shard_index):
+      raise ValueError(
+          'Given shard index {} is outside of expected range: '
+          '[0, {})'.format(shard_index, self._num_shards))
+    return self._table_name_suffixes[shard_index]
+
+  def get_output_table_total_base_pairs(self, shard_index):
+    # type: (int) -> Optional[int]
```
tneymanov: nit: shouldn't Optional be imported as well?

gcp_variant_transforms/libs/variant_sharding.py
samanvp: Sorry, missed this from last round.
Done.

PR #551, opened 2020-02-13
Bq mk table

samanvp:
This PR adds the ability to create integer range partitioned tables using `bq make table...` command.
It's rebased to #550 and will land after that PR.

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_util.py
+++ b/gcp_variant_transforms/libs/bigquery_util.py
@@ -29,7 +29,7 @@ from gcp_variant_transforms.beam_io import vcfio
 
 _VcfHeaderTypeConstants = vcf_header_io.VcfHeaderFieldTypeConstants
 
-_NUM_BQ_RANGE_PARTITIONS = 4000
+_MAX_BQ_NUM_PARTITIONS = 4000
 _TOTAL_BASE_PAIRS_SIG_DIGITS = 4
 _PARTITION_SIZE_SIG_DIGITS = 1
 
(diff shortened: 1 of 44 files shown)


gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -312,3 +313,19 @@ def _get_merged_field_schemas(
         existing_field_schema.fields = _get_merged_field_schemas(
             existing_field_schema.fields, field_schema.fields)
   return merged_field_schemas
+
+
+def calculate_optimal_partition_size(total_base_pairs):
+  # These two operations adds [10^4, 2 * 10^4) buffer to total_base_pairs.
```
tneymanov: s/adds/add

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -28,6 +29,10 @@
 
 _VcfHeaderTypeConstants = vcf_header_io.VcfHeaderFieldTypeConstants
 
+_NUM_BQ_RANGE_PARTITIONS = 4000
```
tneymanov: Since this is not the actual amount we use (we sub 1), I say lets rename it to _MAX_BQ_RANGE_PARTITIONS instead.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -312,3 +313,19 @@ def _get_merged_field_schemas(
         existing_field_schema.fields = _get_merged_field_schemas(
             existing_field_schema.fields, field_schema.fields)
   return merged_field_schemas
+
+
+def calculate_optimal_partition_size(total_base_pairs):
```
tneymanov: Since this is a public facing method, add type def. Plus a doc string with explanation why we are doing these calculations would be preferred.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -329,3 +330,26 @@ def generate_unique_name(job_name):
   return '-'.join([job_name,
                    datetime.now().strftime('%Y%m%d-%H%M%S'),
                    str(uuid.uuid4())])
+
+
```
tneymanov: remove 1 nl.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -329,3 +330,26 @@ def generate_unique_name(job_name):
...
+
+BQ_CREATE_PARTITIONED_TABLE = (
```
tneymanov: Lets rename it to BQ_CREATE_PARTITIONED_TABLE_COMMAND.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -329,3 +330,26 @@ def generate_unique_name(job_name):
...
+    'bq mk --table '
+    '--range_partitioning=start_position,0,{TOTAL_BASE_PAIRS},{PARTITION_SIZE} '
+    '--clustering_fields=start_position,end_position '
+    '{FULL_TABLE_ID} {SCHEMA_JSON}')
+
+def create_output_table_if_needed(known_args, full_table_id, total_base_pairs):
```
tneymanov: Add typedef and docstring.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -481,11 +480,15 @@ def run(argv=None):
         schema_converter.convert_table_schema_to_json_bq_schema(schema))
     with filesystems.FileSystems.create(schema_file) as file_to_write:
       file_to_write.write(schema_json)
+    known_args.schema_json = schema_file
```
tneymanov: I'd rather have this var as `schema_json_file` or `schema_file` - it's not immediately clear that the value is not holding actual schema, especially since before a similar var was doing it. Also maybe we shouldn't overpopulate known_args - schema is used only once AFAICS, so lets keep it as single var.

gcp_variant_transforms/pipeline_common.py
tneymanov: I think it's clearer if we pass append and schema_json directly. WDYT?

samanvp: Agree, modified.

samanvp: done.

gcp_variant_transforms/vcf_to_bq.py
samanvp: As you advised I am sending schema_file directly to the method, so no need to add to known_args anymore.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -329,3 +330,36 @@ def generate_unique_name(job_name):
...
+
+BQ_CREATE_PARTITIONED_TABLE_COMMAND = (
+    'bq mk --table '
+    '--range_partitioning=start_position,0,{TOTAL_BASE_PAIRS},{PARTITION_SIZE} '
+    '--clustering_fields=start_position,end_position '
+    '{FULL_TABLE_ID} {SCHEMA_FILE_PATH}')
+
+def create_output_table(full_table_id, total_base_pairs, schema_file_path):
+  # type: (str, int, str) -> None
+  """Creates an integer range partitioned table using `bq mk table...` command.
```
tneymanov: no-op: Umm I only wanted a top line you've added here, I think it would had been sufficient :D .

Still doesn't hurt to have a nice write up though! Explains well why we have to jump through so many hoops.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -329,3 +330,36 @@ def generate_unique_name(job_name):
...
+
+  Since beam.io.BigQuerySink is unable to create an integer range partition
+  we use `bq mk table...` to achieve this goal. Note this command runs on the
```
tneymanov: nit: Note _that_ ...

PR #553, opened 2020-02-14
Store sample ID as int hash, instead of string hex.

tneymanov:
Also fixed an error where multiple sample tables are being created, one for each chromosome.

DirectRunner test passed, running integration tests now.

### DIFF
--- a/gcp_variant_transforms/beam_io/vcf_parser.py
+++ b/gcp_variant_transforms/beam_io/vcf_parser.py
@@ -184,7 +184,7 @@ class VariantCall(object):
   """
 
   def __init__(self, sample_id=None, genotype=None, phaseset=None, info=None):
-    # type: (str, List[int], str, Dict[str, Any]) -> None
+    # type: (int, List[int], str, Dict[str, Any]) -> None
     """Initialize the :class:`VariantCall` object.
 
     Args:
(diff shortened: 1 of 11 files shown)


samanvp:
Please search all code base for type hints of `sample_id` and modify them to `int`. I remember these two modules specifically from your PR #548:
* `combine_sample_ids.py`
* `densify_variants.py`

tneymanov:
Done.

Also it seems like your last pr that was pushed into master has deleted vcf_to_bq code that creates sample_ id table. Probably sync error. Re-added that code, but now need to verify that it actually works. Will run int tests again.

gcp_variant_transforms/beam_io/vcf_parser.py
```diff
@@ -614,10 +614,10 @@ def _lookup_encoded_sample_name(self, sample_name):
     sample_id = self._encoded_sample_names.get(sample_name)
     if not sample_id:
       if self._sample_name_encoding == SampleNameEncoding.WITH_FILE_PATH:
-        sample_id = hex(hashing_util.generate_sample_id(
-            sample_name, self._file_name))
+        sample_id = hashing_util.generate_sample_id(
+            sample_name, self._file_name)
```
samanvp: nit: break line between the two input arguments, this way it will be easier to compare to the `else` section that only `sample_name` is being sent.

tneymanov: Done.

PR #554, opened 2020-02-15
Remove --partition_config_path and --generate_sample_info_table flags.

tneymanov:
Should be submitted after #553

PR #555, opened 2020-02-20
Modify TABLE_SUFFIX_SEPARATOR logic to distinguish between shards and sample table

tneymanov:
Running integration tests atm.


gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -32,6 +32,15 @@
 _MAX_BQ_NUM_PARTITIONS = 4000
 _TOTAL_BASE_PAIRS_SIG_DIGITS = 4
 _PARTITION_SIZE_SIG_DIGITS = 1
+TABLE_SUFFIX_SEPARATOR = '__'
+SAMPLE_TABLE_SUFFIX_SEPARATOR = '_'
```
samanvp: We only use these two consts in this module (and one test module). Consider adding `_` to them.

tneymanov: Applied this change in #554 . Reverting this PR.

PR #558, opened 2020-02-28
Replace BigQuerySink with WriteToAvro

samanvp:
Due to several reasons, we decided to modify our Dataflow pipeline to generate AVRO files by default. After Dataflow is done we will create multiple output tables using `bq mk table ...` comand. Output tables are sharded per chromosome and each table utilizes [BigQuery integer range partitioning](https://cloud.google.com/bigquery/docs/creating-integer-range-partitions). After that we load AVRO files into BigQuery tables using `bq load ...` command.

This PR also introduces a new flag `--keep_intermediate_avro_files` which allows users to keep AVRO files. We believe AVRO files are a great way to store a back up of BigQuery tables. They are more compressed and also can be stored on GCS for a fraction of the BigQuery storage costs. We will update our docs to reflect these benefits in a follow up PR.


cloudbuild_CI.yaml
```diff
@@ -56,4 +56,4 @@ steps:
       # - '--gs_dir bashir-variant_integration_test_runs'
 images:
   - 'gcr.io/${PROJECT_ID}/gcp-variant-transforms:${COMMIT_SHA}'
-timeout: 200m
+timeout: 250m
```
tneymanov: Do we still need this or is this legacy of the non-avro stuff?

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -136,6 +121,12 @@ def add_arguments(self, parser):
     parser.add_argument('--output_table',
                         default='',
                         help='BigQuery table to store the results.')
+    parser.add_argument(
+        '--keep_intermediate_avro_files',
+        type='bool', default=False, nargs='?', const=True,
+        help= ('If set to True, the intermediate AVRO files will be kept. They '
+               'are stored in your temp directory (set by --temp_location) '
```
tneymanov: Do we know force temp_location to be set or is DirectRunner runs are still ok?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -483,42 +521,34 @@ def run(argv=None):
...
-              omit_empty_sample_calls=known_args.omit_empty_sample_calls,
-              null_numeric_value_replacement=(
-                  known_args.null_numeric_value_replacement))
-      )
-
-  result = pipeline.run()
-  result.wait_until_finish()
-
-  metrics_util.log_all_counters(result)
+      pipeline_common.load_avro_to_output_table(table_name,
```
tneymanov: Hmm, does it make sense to multi-thread this to parallelize BQ load? If BQ doesn't complain I think we should.

Edit: Hmmm, now that I read a bit more seems like handling of deleting tables if Exceptions is raised could be problematic (but yet, not impossible). If time delay is not too too bad, maybe it's ok to keep this... I'm open to discussion.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -363,3 +367,23 @@ def create_output_table(full_table_id, total_base_pairs, schema_file_path):
...
             bq_command))
+
+
+def load_avro_to_output_table(full_table_id, avro_file_base_name):
+  bq_command = _BQ_LOAD_AVRO_COMMAND.format(
+      FULL_TABLE_ID=full_table_id, AVRO_FILE_BASE_NAME=avro_file_base_name)
+  result = os.system(bq_command)
+  if result != 0:
+    raise ValueError(
+        'Failed to load AVRO file to bigquery table using "{}" command.'.format(
```
tneymanov: nit: s/bigquery/BigQuery

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -483,42 +521,34 @@ def run(argv=None):
...
+    for table_name in updated_tables:
+      if known_args.append:
+        logging.warning('Since tables were appended, cannot revert the change'
+                        'in this table: {}'.format(table_name))
+      else:
+        if pipeline_common.delete_table(table_name) != 0:
+          logging.error('Failed to delete table: {}'.format(table_name))
+        else:
+          logging.info('Table was successfully deleted: {}'.format(table_name))
+    logging.info('Since write to BQ stage failed, we do not delete AVRO files. '
```
tneymanov: nit: s/write to BQ stage/Write-to-BQ stage?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -395,6 +395,22 @@ def _run_annotation_pipeline(known_args, pipeline_args):
...
+  schema_file = tempfile.mkstemp(suffix=_BQ_SCHEMA_FILE_SUFFIX)[1]
+  with filesystems.FileSystems.create(schema_file) as file_to_write:
+    file_to_write.write(schema_json)
+  return schema_file
+
+
+def _get_avro_root_path(beam_pipeline_options):
+  google_cloud_options = beam_pipeline_options.view_as(
+      pipeline_options.GoogleCloudOptions)
+  return filesystems.FileSystems.join(google_cloud_options.temp_location,
```
tneymanov: Ummm just to make sure, this adds slashes, right?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -438,42 +454,64 @@ def run(argv=None):
...
+    num_shards = sharding.get_num_shards()
+  else:
+    num_shards = sharding.get_num_shards() - 1
+
+  if known_args.update_schema_on_append:
+    for i in num_shards:
+      table_suffix = sharding.get_output_table_suffix(i)
+      table_name = sample_info_table_schema_generator.compose_table_name(
+          known_args.output_table, table_suffix)
+      bigquery_util.update_bigquery_schema_on_append(schema.fields, table_name)
```
tneymanov: I thought we couldn't update schema for existing tables? Is this the way we circumvent it?

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -136,6 +121,12 @@ def add_arguments(self, parser):
     parser.add_argument('--output_table',
                         default='',
                         help='BigQuery table to store the results.')
+    parser.add_argument(
```
kemp-google: I'd really prefer we not include this flag yet.  If we specifically want to include functionality for lower cost backups (and the savings over compressed VCF are real) I'd rather introduce that as a feature than as an accidental feature like this.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -54,6 +58,11 @@
     '--range_partitioning=start_position,0,{TOTAL_BASE_PAIRS},{PARTITION_SIZE} '
     '--clustering_fields=start_position,end_position '
     '{FULL_TABLE_ID} {SCHEMA_FILE_PATH}')
+_BQ_LOAD_AVRO_COMMAND = (
+    'bq load --source_format=AVRO {FULL_TABLE_ID} {AVRO_FILE_BASE_NAME}-*')
```
kemp-google: We should discuss the relative merits of this versus making API requests and skipping the intermediate files altogether.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -473,61 +481,79 @@ def run(argv=None):
...
+      variants[i] |= ('MergeVariants' + suffix >>
                       merge_variants.MergeVariants(variant_merger))
     variants[i] |= (
-        'ProcessVariants' + str(i) >>
-        beam.Map(processed_variant_factory.create_processed_variant).\
-            with_output_types(processed_variant.ProcessedVariant))
+        'ProcessVariants' + suffix >>
+        beam.Map(processed_variant_factory.create_processed_variant). \
+        with_output_types(processed_variant.ProcessedVariant))
+    # TODO(bashir2): Add an integration test that outputs to Avro files and
```
kemp-google: This TODO doesn't seem useful?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -473,61 +481,79 @@ def run(argv=None):
...
+          known_args.output_table, suffix)
+      if known_args.append:
+        logging.warning('Since tables were appended, cannot revert the change'
+                        'in this table: %s', table_name)
+      else:
+        if pipeline_common.delete_table(table_name) != 0:
+          logging.error('Failed to delete table: %s', table_name)
+        else:
+          logging.info('Table was successfully deleted: %s', table_name)
+    logging.info('Since write to BQ stage failed, we do not delete AVRO files. '
```
kemp-google: This seems like it has the potential to cost users a lot of money if they don't realize they need to garbage collect them...

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -473,61 +481,77 @@ def run(argv=None):
...
+    load_avro = pipeline_common.LoadAvro(
+        avro_root_path, known_args.output_table, suffixes, total_base_pairs)
+    load_avro.start_loading()
+  except Exception as e:
+    logging.error('Something unexpected happened during the loading of AVRO '
+                  'files to BigQuery: %s', str(e))
+    logging.warning('Trying to revert as much as possible...')
+    if known_args.append:
+      logging.warning(
+          'Since tables were appended, we cannot revert added rows. You can '
```
tneymanov: nit: s/'Since tables were appended, we cannot revert added rows.'/'Since tables were appended, added rows cannot be reverted.'

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -473,61 +481,77 @@ def run(argv=None):
...
-
-  metrics_util.log_all_counters(result)
+            table_name, total_base_pairs[i], schema_file)
+        logging.info('Integer range partitioned table %s was created.',
+                     table_name)
+    load_avro = pipeline_common.LoadAvro(
+        avro_root_path, known_args.output_table, suffixes, total_base_pairs)
+    load_avro.start_loading()
+  except Exception as e:
+    logging.error('Something unexpected happened during the loading of AVRO '
```
tneymanov: nit: s/'Something unexpected...'/'An error occurred during the AVRO to BigQuery load stage'.

tneymanov: Since these errors are paramount for users to see, it would be real nice to be able to color the messages or have them in bold font. Looking into how to do it, [this link](https://stackoverflow.com/questions/384076/how-can-i-color-python-logging-output) provides a few options, but it's not super trivial. I'm ok with punting this for now and addressing it in the future.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -473,61 +481,77 @@ def run(argv=None):
...
+        else:
+          logging.error('Failed to delete table: %s', table_name)
+    logging.info('Because write to BigQuery stage failed, we did not delete '
+                 'AVRO files in your GCS bucket. You can manually import them '
+                 'BigQuery. To avoid extra storage charges, delete them if you '
+                 'do not need them, they are located at: %s', avro_root_path)
+    raise e
+  else:
+    logging.warning('All AVRO files were successfully loaded to BigQuery.')
+    if pipeline_common.delete_gcs_files(avro_root_path) != 0:
```
tneymanov: `if pipeline_common.delete_gcs_files(avro_root_path):`

would do the trick, but could be a bit less obvious that an error was raised. your call.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -473,61 +481,77 @@ def run(argv=None):
...
+          'https://cloud.google.com/bigquery/table-decorators')
+    else:
+      for suffix in suffixes:
+        table_name = sample_info_table_schema_generator.compose_table_name(
+            known_args.output_table, suffix)
+        if pipeline_common.delete_table(table_name) == 0:
+          logging.info('Table was successfully deleted: %s', table_name)
+        else:
+          logging.error('Failed to delete table: %s', table_name)
+    logging.info('Because write to BigQuery stage failed, we did not delete '
```
tneymanov: nit: s/Because/Since the

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -473,61 +481,77 @@ def run(argv=None):
...
+                 'AVRO files in your GCS bucket. You can manually import them '
+                 'BigQuery. To avoid extra storage charges, delete them if you '
```
tneymanov: nit: s/BigQuery/to BigQuery

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -473,61 +481,77 @@ def run(argv=None):
...
+                 'do not need them, they are located at: %s', avro_root_path)
```
tneymanov: nit: s/, they are/. AVRO files are

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -473,61 +481,77 @@ def run(argv=None):
...
+    raise e
+  else:
+    logging.warning('All AVRO files were successfully loaded to BigQuery.')
+    if pipeline_common.delete_gcs_files(avro_root_path) != 0:
+      logging.error('Was not able to delete intermediate AVRO files located '
```
tneymanov: nit: s/Was not.../Deletion of intermediate AVRO files located at "{}" has failed.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -361,5 +370,90 @@ def create_output_table(full_table_id, total_base_pairs, schema_file_path):
...
+              ):
+    assert len(suffixes) == len(total_base_pairs)
+
+    self._avro_root_path = avro_root_path
+    self._table_base_name = table_base_name.replace(':', '.')
+    self._suffixes = suffixes
+    self._total_base_pairs = total_base_pairs
+
+    self._num_load_jobs_retries = 0
+    self._suffixes_to_load_jobs = {}  # type: Dict[str, bigquery.job.LoadJob]
```
tneymanov: It would be great in future to initiate load to BQ as soon as corresponding AVRO file has been created, instead of waiting for all AVRO files to be created. However, that seems like too much hassle right now. We should look into it later.

gcp_variant_transforms/pipeline_common.py
```diff
@@ -361,5 +370,90 @@ def create_output_table(full_table_id, total_base_pairs, schema_file_path):
...
+  bq_command = _BQ_DELETE_TABLE_COMMAND.format(FULL_TABLE_ID=full_table_id)
+  return os.system(bq_command)
+
+
+def delete_gcs_files(root_path):
+  gcs_command = _GCS_DELETE_FILES_COMMAND.format(ROOT_PATH=root_path)
+  return os.system(gcs_command)
+
+
+class LoadAvro(object):
```
tneymanov: What do you say about moving this piece of code, and at least delete_gcs_files() method into a different libs file (not sure if delete_table should live there, up to your discretion)? pipeline_common is getting rather crowded and I think afaiu, usually either preprocess or bq-to-vcf have to also invoke it (or maybe these methods should be invoked in multiple places - not the case for us).

gcp_variant_transforms/pipeline_common.py
```diff
@@ -54,6 +57,12 @@
     '--range_partitioning=start_position,0,{TOTAL_BASE_PAIRS},{PARTITION_SIZE} '
     '--clustering_fields=start_position,end_position '
     '{FULL_TABLE_ID} {SCHEMA_FILE_PATH}')
+_BQ_DELETE_TABLE_COMMAND = 'bq rm -f -t {FULL_TABLE_ID}'
+_GCS_DELETE_FILES_COMMAND = 'gsutil -m rm -f -R {ROOT_PATH}'
+_BQ_LOAD_JOB_NUM_RETRIES = 5
+_MAX_NUM_CONCURRENT_BQ_LOAD_JOBS = 5
```
tneymanov: So this is BQ limitation right? Not affected by the count of cores or the count of workers invoked?

gcp_variant_transforms/pipeline_common.py
```diff
@@ -361,5 +370,90 @@ def create_output_table(full_table_id, total_base_pairs, schema_file_path):
...
+
+  def _cancel_all_running_load_jobs(self):
+    for _, load_job in self._suffixes_to_load_jobs.items():
+      load_job.cancel()
+
+  def _handle_failed_load_job(self, suffix, load_job):
+    if self._num_load_jobs_retries < _BQ_LOAD_JOB_NUM_RETRIES:
+      self._num_load_jobs_retries += 1
+      # Retry the failed job after 5 minutes wait.
+      time.sleep(300)
```
tneymanov: why such a drastic wait time? doesn't reducing the count of parallel bq jobs do the trick?

gcp_variant_transforms/pipeline_common.py
```diff
@@ -361,5 +370,90 @@ def create_output_table(full_table_id, total_base_pairs, schema_file_path):
...
+    # Waits until current jobs are done and then add remaining jobs one by one.
+    while self._suffixes_to_load_jobs:
+      time.sleep(60)
+      processed_suffixes = self._suffixes_to_load_jobs.keys()
+      for suffix in processed_suffixes:
+        load_job = self._suffixes_to_load_jobs.get(suffix)
+        if load_job.done():
+          del self._suffixes_to_load_jobs[suffix]
+          state = load_job.state
+          if state != 'DONE':
```
tneymanov: We could inline to `if load_job.state != 'DONE':`

gcp_variant_transforms/testing/integration/run_vcf_to_bq_tests.py
```diff
@@ -124,7 +124,7 @@ def run_assertion(self):
     num_retries = 0
     while True:
       try:
-        iterator = query_job.result(timeout=30)
+        iterator = query_job.result(timeout=300)
```
tneymanov: We still need this amount, right?

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -360,3 +375,115 @@ def calculate_optimal_partition_size(total_base_pairs):
...
+  bq_command = _BQ_CREATE_PARTITIONED_TABLE_COMMAND.format(
+      TOTAL_BASE_PAIRS=total_base_pairs_enlarged,
+      PARTITION_SIZE=partition_size,
+      FULL_TABLE_ID=full_table_id,
+      SCHEMA_FILE_PATH=schema_file_path)
+  result = os.system(bq_command)
+  if result != 0:
+    time.sleep(30)  # In our integration tests sometime we overwhelm BQ server.
+    result_second_attempt = os.system(bq_command)
+    if result_second_attempt != 0:
```
samanvp: I added an extra re-try to avoid cases where our integration tests fail due to this error:
```
BigQuery error in mk operation: Got 503 response from discovery url:
https://www.googleapis.com/discovery/v1/apis/bigquery/v2/rest
```
As soon as we switch to using BigQuery Python API for creating tables we will remove this temporary solution.

PR #567, opened 2020-03-19
Set machine-type of the PAPI worker

samanvp:
Since this worker only launches the Dataflow pipeline, monitors the result, and finally loads AVRO files to the destination BQ tables, it does not need a lot of resources. Using a smaller machine would reduce the overall cost of running VariantTransforms.


cloudbuild_CI.yaml
```diff
@@ -41,10 +41,9 @@ steps:
       - '--skip_build'
       - '--project ${PROJECT_ID}'
       - '--image_tag ${COMMIT_SHA}'
-      - '--run_unit_tests'
       - '--run_preprocessor_tests'
       - '--run_bq_to_vcf_tests'
-      - '--run_all_tests'
+      - '--run_presubmit_tests'
```
tneymanov: fix this after you're done debugging.

samanvp: Done.

PR #568, opened 2020-03-19
Add a flag to allow users to keep AVRO files

samanvp:
AVRO files could be an effective solution as a ready-to-load-into-BQ.

Storing AVRo file in GCS costs a fraction of BQ's storage cost. Also AVRO files compression makes them significantly smaller than their corresponding BQ tables.

### DIFF
--- a/gcp_variant_transforms/options/variant_transform_options.py
+++ b/gcp_variant_transforms/options/variant_transform_options.py
@@ -118,12 +118,16 @@ class BigQueryWriteOptions(VariantTransformsOptions):
     parser.add_argument('--output_table',
                         default='',
                         help='BigQuery table to store the results.')
-
+    parser.add_argument(
+        '--output_avro_path',
+        default='',
+        help=('This flag is deprecated and will be removed in the next '
+              'release, instead use --keep_intermediate_avro_files flag.'))
     parser.add_argument(
         '--keep_intermediate_avro_files',
         type='bool', default=False, nargs='?', const=True,
-        help=('If set to True, the intermediate AVRO files will be kept. They '	
-              'are stored in your temp directory (set by --temp_location) '	
+        help=('If set to True, the intermediate AVRO files will be kept. They '
+              'are stored in your temp directory (set by --temp_location) '
               'under gs://[YOUR-TEMP-DIRECTORY]/avro/JOB_NAME/'))
     parser.add_argument(
         '--sharding_config_path',
--- a/gcp_variant_transforms/options/variant_transform_options.py
+++ b/gcp_variant_transforms/options/variant_transform_options.py
@@ -120,6 +120,12 @@ class BigQueryWriteOptions(VariantTransformsOptions):
                         help='BigQuery table to store the results.')
 
     parser.add_argument(
+        '--keep_intermediate_avro_files',
+        type='bool', default=False, nargs='?', const=True,
+        help=('If set to True, the intermediate AVRO files will be kept. They '	
+              'are stored in your temp directory (set by --temp_location) '	
+              'under gs://[YOUR-TEMP-DIRECTORY]/avro/JOB_NAME/'))
+    parser.add_argument(
         '--sharding_config_path',
         default=('gcp_variant_transforms/data/sharding_configs/'
                  'homo_sapiens_default.yaml'),
--- a/gcp_variant_transforms/vcf_to_bq.py
+++ b/gcp_variant_transforms/vcf_to_bq.py
@@ -541,9 +541,13 @@ def run(argv=None):
     raise e
   else:
     logging.warning('All AVRO files were successfully loaded to BigQuery.')
-    if bigquery_util.delete_gcs_files(avro_root_path) != 0:
-      logging.error('Deletion of intermediate AVRO files located at "%s" has '
-                    'failed.', avro_root_path)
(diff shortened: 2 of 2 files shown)


bashir2:
Sorry, I have not followed closely the changes in VT for a long time but just from the title of this pull request: Isn't this option already available? The first time that I added some avro support in VT, i.e., PR #411 this was the main intended use case through `--output_avro_path`; is that code now dead? BTW, great to see the shift to more usage of avro format for the intermediate files.

samanvp:
Thanks @bashir2 for your comment and paying attention to our development.

Your code is *not* dead, in fact, the [variant_to_avro](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/transforms/variant_to_avro.py) is used by default in all VT runs. In #558 we completely abandoned the old BQ sink and instead we write the output of Dataflow to AVRO files. The resulting AVRO files will be loaded to the final BQ destination tables using BQ API. We had multiple reasons to make this decision:
* With each new release of Beam we found new bugs in the old sink. Beam team were also reluctant to spend time on it and instead they rather to finish new sink.
* We also could not switch to the new sink because: 
  * In their latest release it was failing for large inputs such as 1000 Genomes.
  * New sink has multiple "sub nodes", combined with our per chromosome sharding we get `"the dataflow graph is too large"` error which is because we hit this [Dataflow limit](https://cloud.google.com/dataflow/docs/guides/common-errors#job-graph-too-large).
  * New BQ sink still in not able to perform [integer range partitioning](https://cloud.google.com/bigquery/docs/creating-integer-range-partitions).

We were told by Beam team these changes will be released in about 6 months. So we decided to use the Dataflow -> AVRO -> BQ approach. 

It turns out AVRO files offer some additional benefits that might be valuable to some users (I will share a doc with you soon). In summary AVRO files are more compressed and less expensive to store on GCS. So we decided to add this flag to allow users to save intermediate AVRO files if they wanted to.

Hope this reply makes our intentions more clear.

bashir2:
> Thanks @bashir2 for your comment and paying attention to our development.
> 
> Your code is _not_ dead, in fact, the [variant_to_avro](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/transforms/variant_to_avro.py) is used by default in all VT runs. In #558 we completely abandoned the old BQ sink and instead we write the output of Dataflow to AVRO files. The resulting AVRO files will be loaded to the final BQ destination tables using BQ API. We had multiple reasons to make this decision:
> 
> * With each new release of Beam we found new bugs in the old sink. Beam team were also reluctant to spend time on it and instead they rather to finish new sink.
> * We also could not switch to the new sink because:
>   
>   * In their latest release it was failing for large inputs such as 1000 Genomes.
>   * New sink has multiple "sub nodes", combined with our per chromosome sharding we get `"the dataflow graph is too large"` error which is because we hit this [Dataflow limit](https://cloud.google.com/dataflow/docs/guides/common-errors#job-graph-too-large).
>   * New BQ sink still in not able to perform [integer range partitioning](https://cloud.google.com/bigquery/docs/creating-integer-range-partitions).
> 
> We were told by Beam team these changes will be released in about 6 months. So we decided to use the Dataflow -> AVRO -> BQ approach.
> 
> It turns out AVRO files offer some additional benefits that might be valuable to some users (I will share a doc with you soon). In summary AVRO files are more compressed and less expensive to store on GCS. So we decided to add this flag to allow users to save intermediate AVRO files if they wanted to.
> 
> Hope this reply makes our intentions more clear.

Thanks @samanvp for the details. Great to see Avro files playing a more central role in VT. In general, this will make VT a more useful tool in contexts other than BigQuery too. That was the main reason for implementing it in the first place.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -120,7 +120,17 @@ def add_arguments(self, parser):
     parser.add_argument('--output_table',
                         default='',
                         help='BigQuery table to store the results.')
-
+    parser.add_argument(
+        '--output_avro_path',
```
tneymanov: Are we reading this just to deprecate it? If so, sg.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -120,7 +120,17 @@ def add_arguments(self, parser):
...
+        default='',
+        help=('This flag is deprecated and will be removed in the next '
+              'release, instead use --keep_intermediate_avro_files flag.'))
+    parser.add_argument(
+        '--keep_intermediate_avro_files',
```
tneymanov: Did you and Aaron come to consensus on this? I thought the idea was to not have any flags like these?

gcp_variant_transforms/options/variant_transform_options.py
samanvp: Yes, I should have not deleted this in my earlier PR.

gcp_variant_transforms/options/variant_transform_options.py
samanvp: Based in [his comment](https://github.com/googlegenomics/gcp-variant-transforms/pull/558#discussion_r387878993) I assumed he did not want this feature to be added in #558. @kemp-google please let me know if this is the case.

PR #569, opened 2020-03-19
Remove extra flags after AVRO change

samanvp:
After #558 some of the input flags are not needed anymore, namely:
* `--optimize_for_large_inputs`
* `--num_bigquery_write_shards`


gcp_variant_transforms/pipeline_common_test.py
```diff
@@ -50,25 +49,22 @@ def test_validation_failure_for_invalid_input_pattern(self):
           input_pattern='nonexistent_file.vcf', input_file=None)
 
   def test_get_mode_optimize_set(self):
```
tneymanov: This test doesn't make sense anymore, please remove.

gcp_variant_transforms/pipeline_common_test.py
```diff
@@ -178,26 +173,19 @@ def _create_mock_args(self, **args):
   def _get_pipeline_mode(self, args):
     all_patterns = pipeline_common._get_all_patterns(args.input_pattern,
                                                      args.input_file)
-    return pipeline_common.get_pipeline_mode(all_patterns,
-                                             args.optimize_for_large_inputs)
+    return pipeline_common.get_pipeline_mode(all_patterns)
 
   def test_get_mode_optimize_set(self):
```
tneymanov: Same here, this test should be unnecessary.

samanvp: Done.

PR #571, opened 2020-03-27
Encorceing output_table does not contain "__"

samanvp:
We use this special string to mark sharded output table, for example if the give output table is:
`project_id:dataset_id.table_id`

and in the sharding config file we have 3 shards with these suffixes:
* `chr1`
* `chr2`
* `residual`

then we will have these 4 outputs:

* `project_id:dataset_id.table_id__chr1`
* `project_id:dataset_id.table_id__chr2`
* `project_id:dataset_id.table_id__residual`
* `project_id:dataset_id.table_id__sample_info_table`

That's why we don't allow "__" to be used in the base table name.

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_util.py
+++ b/gcp_variant_transforms/libs/bigquery_util.py
@@ -380,7 +380,7 @@ def compose_table_name(base_name, suffix, is_sample=False):
   return separator.join([base_name, suffix])
 
 def get_table_base_name(table_name):
-  return table_name.split(SAMPLE_TABLE_SUFFIX_SEPARATOR)[0]
+  return table_name.split(TABLE_SUFFIX_SEPARATOR)[0]
 
 class LoadAvro(object):
   def __init__(self,
--- a/gcp_variant_transforms/libs/bigquery_util.py
+++ b/gcp_variant_transforms/libs/bigquery_util.py
@@ -379,6 +379,8 @@ def compose_table_name(base_name, suffix, is_sample=False):
       SAMPLE_TABLE_SUFFIX_SEPARATOR if is_sample else TABLE_SUFFIX_SEPARATOR)
   return separator.join([base_name, suffix])
 
+def get_table_base_name(table_name):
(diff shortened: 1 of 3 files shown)


gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -379,6 +379,8 @@ def compose_table_name(base_name, suffix, is_sample=False):
       SAMPLE_TABLE_SUFFIX_SEPARATOR if is_sample else TABLE_SUFFIX_SEPARATOR)
   return separator.join([base_name, suffix])
 
+def get_table_base_name(table_name):
+  return table_name.split(SAMPLE_TABLE_SUFFIX_SEPARATOR)[0]
```
tneymanov: Hmmm, technically this should be TABLE_SUFFIX_SEPARATOR before my follow up PR, but I guess this is fine. No-op.

gcp_variant_transforms/libs/bigquery_util_test.py
```diff
@@ -460,6 +460,28 @@ def test_raise_error_if_dataset_not_exists(self):
...
 
+  def test_get_table_base_name(self):
+    without_suffix1 = 'project_id.dataset_id.table_id'
+    without_suffix2 = 'project_id:dataset_id.table_id'
+    self.assertEqual(without_suffix1,
+                     bigquery_util.get_table_base_name(without_suffix1))
+    self.assertEqual(without_suffix2,
+                     bigquery_util.get_table_base_name(without_suffix2))
+
+    with_suffix1 = without_suffix1 + '___chr1'
```
tneymanov: Please modify to `__chr1`, here and everywhere else below (ie. 2 underscores, if you are using SAMPLE_TABLE_SUFFIX_SEPARATOR).

gcp_variant_transforms/libs/bigquery_util.py
samanvp: You are right, changed.
My intention was/is to later remove `SAMPLE_TABLE_SUFFIX_SEPARATOR` and change the value of `TABLE_SUFFIX_SEPARATOR` to `__`.

PR #573, opened 2020-03-31
Enable users to store sample-lookup-optimized tables

samanvp:
This PR includes the following changes:
 + Add a new input flag and validate its value.
 + Add BQ queries to flatten call column.
 + Extract the schema of the flatten table.
 + Add unit tests to verify the correctness of extracted schema.
 + Create tables using extracted schema which are partitioned on `call_sample_id` column.
 + Copy date from variant-lookup-optimized tables to sample-lookup-optimized tables.

We will add integration tests in a follow up PR.


gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -482,6 +501,156 @@ def create_sample_info_table(output_table_id):
...
+          num_retries += 1
+          time.sleep(90)
+        else:
+          logging.error('Copy to table query failed: %s', output_table_id)
+          raise e
+      else:
+        break
+    logging.info('Copy to table query was successful: %s', output_table_id)
+
+  def _create_temp_flatten_table(self):
```
tneymanov, line 592: Hmm naming doesn't really represent what's happening here, but there is no really good name...

How about _create_temp_flatten_table_with_1_row at least?

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -482,6 +501,156 @@ def create_sample_info_table(output_table_id):
...
+  def get_flatten_table_schema(self, schema_file_path):
+    temp_table_id = self._create_temp_flatten_table()
+    full_table_id = '{}:{}.{}'.format(
+        self._project_id, self._dataset_id, temp_table_id)
+    bq_command = _BQ_EXTRACT_SCHEMA_COMMAND.format(
+        FULL_TABLE_ID=full_table_id,
+        SCHEMA_FILE_PATH=schema_file_path)
+    result = os.system(bq_command)
+    if result != 0:
+      logging.error('Failed to extract flatten table schema using "%s" command',
```
tneymanov, line 621: We log errors but don't throw exceptions? Below makes more sense, but here?

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -124,9 +124,27 @@ class BigQueryWriteOptions(VariantTransformsOptions):
...
+              ' * ... '
+              ' * `output_table`__residual '
+              'where "chr1", "chr2", ..., and "residual" suffixes correspond '
+              'to the value of `table_name_suffix` in the sharding config file '
+              '(see --sharding_config_path).'))
+
+    parser.add_argument(
+        '--sample_lookup_optimized_output_table',
+        default='',
+        help=('In addition to the default output tables (which are optimized '
```
tneymanov, line 143: Does it make sense to allow either output_table, sample_lookup_optimized_output_table or both instead of enforcing output_table?

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -206,54 +224,66 @@ def validate(self, parsed_args, client=None):
...
+      credentials = GoogleCredentials.get_application_default().create_scoped(
+          ['https://www.googleapis.com/auth/bigquery'])
+      client = bigquery.BigqueryV2(credentials=credentials)
+    if not parsed_args.output_table:
+      raise ValueError('--output_table must have a value.')
+    self._validate_output_tables(
+        client, parsed_args.output_table,
+        parsed_args.sharding_config_path, parsed_args.append)
+
+    if parsed_args.sample_lookup_optimized_output_table:
```
tneymanov, line 238: We should make sure sample_lookup_optimized_output_table does not equal output_table.

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -124,9 +124,27 @@ class BigQueryWriteOptions(VariantTransformsOptions):
...
+              'to the value of `table_name_suffix` in the sharding config file '
+              '(see --sharding_config_path).'))
+
+    parser.add_argument(
+        '--sample_lookup_optimized_output_table',
+        default='',
+        help=('In addition to the default output tables (which are optimized '
+              'for variant look up queries), you can store a second copy of '
+              'your data in BigQuery tables that are optimized for sample '
+              'look up queries. Note that setting this option will double your '
```
tneymanov, line 146: Since the tables are going to be unflattened, are their storage costs going to be more than the original output_tables? Or do they end up roughly equal?

gcp_variant_transforms/options/variant_transform_options.py
samanvp: Not really, we fill `sample_lookup_optimized_output_table` using BQ queries based on the schema and content of `output_table`.
On a more conceptual level, we assume VT is expected to always create variant lookup optimized tables while sample lookup optimized tables are needed by only some of the users.

gcp_variant_transforms/options/variant_transform_options.py
samanvp: If data is not joint genotyped then it's approximately equal.
I updated the help to make this issue clear, thanks for spotting it.

gcp_variant_transforms/options/variant_transform_options.py
samanvp: Thanks for spotting this corner case!

gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -206,54 +224,66 @@ def validate(self, parsed_args, client=None):
...
+      self._validate_output_tables(
+          client, parsed_args.sample_lookup_optimized_output_table,
+          parsed_args.sharding_config_path, parsed_args.append)
+
+  def _validate_output_tables(self, client,
+                              output_table_base_name,
+                              sharding_config_path, append):
+    if (output_table_base_name !=
+        bigquery_util.get_table_base_name(output_table_base_name)):
+      raise ValueError(('Output table cannot contain "{}" we reserve this  '
```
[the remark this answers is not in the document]
samanvp: Done.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -508,6 +527,199 @@ def create_sample_info_table(output_table_id):
...
+      assert self._sub_fields
+    return self._sub_fields
+
+  def _get_flatten_column_names(self):
+    column_names = self._get_column_names()
+    sub_fields = self._get_call_sub_fields()
+    select_list = []
+    for column in column_names:
+      if column != ColumnKeyConstants.CALLS:
+        select_list.append(_MAIN_TABLE_ALIAS + '.' + column + ' AS `'+
```
tneymanov, line 605: This should be done in template.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -508,6 +527,199 @@ def create_sample_info_table(output_table_id):
...
+
+    select_columns = self._get_flatten_column_names()
+    cp_query = _FLATTEN_CALL_QUERY.format(SELECT_COLUMNS=select_columns,
+                                          PROJECT_ID=self._project_id,
+                                          DATASET_ID=self._dataset_id,
+                                          TABLE_ID=self._schema_table_id,
+                                          MAIN_TABLE_ALIAS=_MAIN_TABLE_ALIAS,
+                                          CALL_COLUMN=ColumnKeyConstants.CALLS,
+                                          CALL_TABLE_ALIAS=_CALL_TABLE_ALIAS)
+    cp_query += ' LIMIT 1'  # We need this table only to extract its schema.
```
tneymanov, line 646: Maybe put this into const too.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -508,6 +527,199 @@ def create_sample_info_table(output_table_id):
       SCHEMA_FILE_PATH=SAMPLE_INFO_TABLE_SCHEMA_FILE_PATH)
   _run_table_creation_command(bq_command)
 
+class FlattenCallColumn(object):
+  """Flattens call column to convert varinat opt tables to sample opt tables."""
```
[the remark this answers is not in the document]
samanvp: Done.

gcp_variant_transforms/libs/bigquery_util.py
samanvp: Done.

gcp_variant_transforms/libs/bigquery_util.py
samanvp: I don't think this will be used in any other method in this module. So for now, if you don't mine, I will keep it here.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -508,6 +527,199 @@ def create_sample_info_table(output_table_id):
...
+    cp_query = _FLATTEN_CALL_QUERY.format(SELECT_COLUMNS=select_columns,
+                                          PROJECT_ID=self._project_id,
+                                          DATASET_ID=self._dataset_id,
+                                          TABLE_ID=self._schema_table_id,
+                                          MAIN_TABLE_ALIAS=_MAIN_TABLE_ALIAS,
+                                          CALL_COLUMN=ColumnKeyConstants.CALLS,
+                                          CALL_TABLE_ALIAS=_CALL_TABLE_ALIAS)
+    cp_query += ' LIMIT 1'  # We need this table only to extract its schema.
+    self._copy_to_flatten_table(full_output_table_id, cp_query)
+    logging.info('A new table with 1 row was crated: %s', full_output_table_id)
```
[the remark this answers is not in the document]
samanvp: Done.

PR #577, opened 2020-04-09
Generate Sample Info when representative header is being supplied.

tneymanov:
As per offline discussion, attempted to add proper checks but they seemed a bit too complicated, since input data may be compressed files - will need to generate actual Beam input source and verify that compression is handled correctly.

Rather opted for fast and dirty approach.


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -435,6 +435,8 @@ def run(argv=None):
   # See https://issues.apache.org/jira/browse/BEAM-2801.
   header_fields = vcf_header_parser.get_vcf_headers(
       known_args.representative_header_file)
+  beam_pipeline_options = pipeline_options.PipelineOptions(pipeline_args)
```
samanvp, line 438: Any specific reason why you moved this line? I prefer its previous spot because `beam_pipeline_options` object is used exactly in the next line.

gcp_variant_transforms/testing/integration/vcf_to_bq_tests/presubmit_tests/medium_tests/gnomad_genomes_GRCh37_chrX_head2500_run_vep.json
```diff
@@ -12,6 +12,10 @@
     "shard_variants": "False",
     "num_workers": 2,
     "assertion_configs": [
+      {
+        "query": ["NUM_OUTPUT_TABLES"],
+        "expected_result": {"num_tables": 26}
+      },
```
samanvp, line 18: Could you please make sure all integration tests now have this check. And also for all of them the expected number must include `sample_info` table.

gcp_variant_transforms/vcf_to_bq.py
tneymanov: Artifact of old implementation. Reverted.

gcp_variant_transforms/testing/integration/vcf_to_bq_tests/presubmit_tests/medium_tests/gnomad_genomes_GRCh37_chrX_head2500_run_vep.json
tneymanov: Done.

tneymanov:
Regarding the proposed removal of this logic for cases without VCF files, it's not as cut and dry - right now we assume that sample_info is generated everywhere, and to allow flexible sample_info creation will need to handle lots of cases. Aside from obvious bq_to_vcf flow, we will need to consider every place where append is being used to check if sample_info is generated - sometimes it's possible that initially the tables wasn't generated, but it needs to be generated now as new data has samples now. Rollback mechanisms have to address this too.

This change is going to be a headache of its own, so we need to approach it carefully, separately from this PR.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -325,16 +323,21 @@ def _merge_headers(known_args, pipeline_args,
...
 
   with beam.Pipeline(options=options) as p:
     headers = pipeline_common.read_headers(
         p, pipeline_mode,
         known_args.all_patterns)
-    _ = (headers | 'SampleInfoToBigQuery' >>
-         sample_info_to_bigquery.SampleInfoToBigQuery(
+    _ = (headers
+         | 'SampleInfoToBigQuery'
+         >> sample_info_to_bigquery.SampleInfoToBigQuery(
```
samanvp, line 335: nit: This change is not needed.

PR #580, opened 2020-04-20
Use partition column as a flag in _BQ_CREATE_PARTITIONED_TABLE_COMMAND

samanvp:
We will be using this command to create two types of output tables:
 + Usual output tables, partitioned by `start_position` optimized for variant lookup queries.
 + Duplicated output tables, partitioned by `sample_id` optimized for sample lookup queires.

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_util.py
+++ b/gcp_variant_transforms/libs/bigquery_util.py
@@ -37,13 +37,13 @@ SAMPLE_INFO_TABLE_SUFFIX = 'sample_info'
 TABLE_SUFFIX_SEPARATOR = '__'
 
 _MAX_BQ_NUM_PARTITIONS = 4000
-_TOTAL_BASE_PAIRS_SIG_DIGITS = 4
-_PARTITION_SIZE_SIG_DIGITS = 1
+_RANGE_END_SIG_DIGITS = 4
+_RANGE_INTERVAL_SIG_DIGITS = 1
 
 START_POSITION_COLUMN = 'start_position'
 _BQ_CREATE_PARTITIONED_TABLE_COMMAND = (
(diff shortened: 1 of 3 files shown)


gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -40,17 +40,16 @@
 _TOTAL_BASE_PAIRS_SIG_DIGITS = 4
 _PARTITION_SIZE_SIG_DIGITS = 1
 
+START_POSITION_COLUMN = 'start_position'
 _BQ_CREATE_PARTITIONED_TABLE_COMMAND = (
-    'bq mk --table '
-    '--range_partitioning=start_position,0,{TOTAL_BASE_PAIRS},{PARTITION_SIZE} '
+    'bq mk --table --range_partitioning='
+    '{PARTITION_COLUMN},0,{TOTAL_BASE_PAIRS},{PARTITION_SIZE} '
```
tneymanov, line 46: If the partition column can vary, does it still make sense to call end of range partition TOTAL_BASE_PAIRS?

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -40,17 +40,16 @@
 _TOTAL_BASE_PAIRS_SIG_DIGITS = 4
 _PARTITION_SIZE_SIG_DIGITS = 1
 
+START_POSITION_COLUMN = 'start_position'
```
tneymanov, line 43: s/START_POSITION_COLUMN/START_POSITION_COLUMN_NAME

samanvp: I realized we have `ColumnKeyConstants` so there is no need for this const value.

Issue #582, opened 2020-04-20
Use BigQuery API to create integer range partition tables


samanvp:
We currently use `bq mk --table ...` to create integer ranged partitioned tables:
https://github.com/googlegenomics/gcp-variant-transforms/blob/708c34acd04062f82048b658d246452bcc5678fd/gcp_variant_transforms/libs/bigquery_util.py#L43

We should use BQ API to accomplish this task, however, this needs newer version of google-cloud-bigquery library:
https://github.com/googleapis/google-cloud-python/releases/tag/bigquery-1.22.0

We are currently pinned to 1:17.0 due to Beam requirements. As soon as Beam's pinned version updates to >= 1.22.0 we should update our code to replace `bq` system call with API calls.

samanvp:
[Beam 2.20.0](https://github.com/apache/beam/releases/tag/v2.20.0) updates bigquery library version:
```
    'google-cloud-bigquery>=1.6.0,<=1.24.0',
```
As soon as #583 lands we can resolve this issue.

PR #584, opened 2020-04-20
Remove an extra unused flag

samanvp:
This was left due to issue #582

### DIFF
--- a/gcp_variant_transforms/vcf_to_bq.py
+++ b/gcp_variant_transforms/vcf_to_bq.py
@@ -516,20 +516,19 @@ def run(argv=None):
   # After pipeline is done, create output tables and load AVRO files into them.
   schema_file = _write_schema_to_temp_file(schema)
   suffixes = []
-  total_base_pairs = []
   try:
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -516,20 +516,19 @@ def run(argv=None):
   # After pipeline is done, create output tables and load AVRO files into them.
   schema_file = _write_schema_to_temp_file(schema)
   suffixes = []
-  total_base_pairs = []
   try:
     for i in range(num_shards):
       suffixes.append(sharding.get_output_table_suffix(i))
-      total_base_pairs.append(sharding.get_output_table_total_base_pairs(i))
+      total_base_pairs = sharding.get_output_table_total_base_pairs(i)
```
tneymanov, line 522: Lets move this into the if statement.

samanvp: If you don't mind I'd prefer to keep this line here, since we are reading both of these values from the config file.

PR #588, opened 2020-04-24
Don't hash sample names on shard variants.

tneymanov:
Also adds an option to not hash samples if we wish to, although that option is currently disabled.


tneymanov:
Int tests reran after sync, and passed.

Also ran annotation pipeline manually, and verified correct hashing.

PR #593, opened 2020-05-07
Remove two large tests temporarily

samanvp:
Our tests have been failing so we are looking for ways to reduce the load:
+ Added a 5 sec delay between running each verification query.
+ Removed two large tests that was originally added for verifying the logic of sharding. we will latter add them back as medium tests (instead of whole platinum as input we will create VCF files with only 3K randomly selected lines).


samanvp:
I am fairly sure added 5 seconds delay will solve our current issue. I think this issue appeared when instead of `.*` queries we started using multiple queries for each sharded output table.

A test run is running now, if green please LGTM.

PR #594, opened 2020-05-07
Modify two merge tests from large to medium

samanvp:
For merge logic we do not need to use the whole platinum files, instead here we use only chr1, 2, and 3 from platinum input files.


cloudbuild_CI.yaml
```diff
@@ -41,10 +41,7 @@ steps:
       - '--skip_build'
       - '--project ${PROJECT_ID}'
       - '--image_tag ${COMMIT_SHA}'
-      - '--run_unit_tests'
-      - '--run_preprocessor_tests'
-      - '--run_bq_to_vcf_tests'
-      - '--run_all_tests'
```
tneymanov, line 47: please revert.

PR #596, opened 2020-05-12
Delete empty tables after AVRO copy stage

samanvp:
In order to make the output of VT more clear, in this PR we examine all output tables and delete empty ones.
Empty tables are created when in the sharding config we have an entry for a chromosome (say ChrM) while the input VCF file does not have any line with that reference_name.


samanvp:
@tneymanov this PR is also ready for review now. This is a prerequisit of #573

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -196,6 +196,7 @@ def table_exist(client, project_id, dataset_id, table_id):
       raise
   return True
 
+
```
tneymanov, line 199: nit: unnecessary line.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -461,10 +467,30 @@ def _monitor_load_jobs(self):
           if load_job.state != 'DONE':
             self._handle_failed_load_job(suffix, load_job)
           else:
+            self._delete_empty_table(suffix, load_job)
```
tneymanov, line 470: So if a job fails, we keep potentially empty table? Is it because load_job will not have output_rows?

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -420,6 +425,7 @@ def start_loading(self):
       self._start_one_load_job(self._remaining_load_jobs.pop())
 
     self._monitor_load_jobs()
+    return self._not_empty_suffixes
```
tneymanov, line 428: So now we return value but don't use it in either bq_util_test or vcf_to_bq... Artifact of testing or is it going to be used somewhere in the future?

gcp_variant_transforms/libs/bigquery_util.py
samanvp: There are two empty lines between all functions in this module, I just made this one consistent with the rest of them.

gcp_variant_transforms/libs/bigquery_util.py
samanvp: We will use this returned value in #573 : only generate sample lookup optimized tables for not empty ones.

gcp_variant_transforms/libs/bigquery_util.py
samanvp: If a job fails more than `_BQ_NUM_RETRIES` times then this class raise an exception. At that point it's up to the caller what to do with the newly created not empty tables.

In our case, we call `rollback_newly_created_tables` in `vcf_to_bq` module.

PR #597, opened 2020-05-13
Capping range_end value

samanvp:
Making sure our range_end does not exceed BigQuery's limit of `pow(2, 63) - 1`

### DIFF
--- a/gcp_variant_transforms/libs/bigquery_util.py
+++ b/gcp_variant_transforms/libs/bigquery_util.py
@@ -38,6 +38,7 @@ TABLE_SUFFIX_SEPARATOR = '__'
 SAMPLE_INFO_TABLE_SCHEMA_FILE_PATH = (
     'gcp_variant_transforms/data/schema/sample_info.json')
 
+MAX_RANGE_END = pow(2, 63) - 1
 _MAX_BQ_NUM_PARTITIONS = 4000
 _RANGE_END_SIG_DIGITS = 4
 _RANGE_INTERVAL_SIG_DIGITS = 1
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -357,12 +358,18 @@ def calculate_optimal_range_interval(range_end):
...
+  by BigQuery integer range partitioning.
+
   Args:
     range_end: the maximum value of the column subject to partitioning
 
   Returns:
     A tuple (partition size, partition size * 3999).
   """
+  if range_end >= MAX_RANGE_END:
+    return(int(MAX_RANGE_END / float(_MAX_BQ_NUM_PARTITIONS)),
```
tneymanov, line 371: Hmm, I see what you are doing here, but can't we find a suitable name for the constant MAX_RANGE_END/_MAX_BQ_NUM_PARTITIONS? it's not really max_interval, but if we can find a suitable name, we should move it.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -374,8 +381,13 @@ def calculate_optimal_range_interval(range_end):
...
       math.ceil(range_interval / pow(10, _RANGE_INTERVAL_SIG_DIGITS)) *
       math.pow(10, _RANGE_INTERVAL_SIG_DIGITS))
-  return (range_interval_round_up,
-          range_interval_round_up * (_MAX_BQ_NUM_PARTITIONS - 1))
+  range_end_round_up = range_interval_round_up * (_MAX_BQ_NUM_PARTITIONS - 1)
+
+  if  range_end_round_up < MAX_RANGE_END:
+    return (range_interval_round_up, range_end_round_up)
+  else:
+    return(int(MAX_RANGE_END / float(_MAX_BQ_NUM_PARTITIONS)),
```
tneymanov, line 389: Same here.

PR #599, opened 2020-05-16
Refactor how we process `query_job.result` output

samanvp:
in `run_vcf_to_bq_tests.QueryAssertion.run_assertion()`

### DIFF
--- a/gcp_variant_transforms/testing/integration/run_vcf_to_bq_tests.py
+++ b/gcp_variant_transforms/testing/integration/run_vcf_to_bq_tests.py
@@ -125,7 +125,7 @@ class QueryAssertion(object):
     num_retries = 0
     while True:
       try:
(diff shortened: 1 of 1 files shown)


samanvp:
> this is why tests would stop if we get empty outputs

It might be, after observing the same issue (not empty tables return 0 rows) I tried to debug and realized the way we call API does not follow the API guidelines.

> Can we restore removed tests after this?

As far as I know, the only removed tests we have is [this one](https://github.com/googlegenomics/gcp-variant-transforms/blob/master/gcp_variant_transforms/testing/integration/vcf_to_bq_tests/presubmit_tests/large_tests/test_non_splittable_gzip.temp_deleted_json) which is failing due to Dataflow graph size. Any other test we need to restore?

> if int tests pass

Yes, I had a test run and was green.

PR #600, opened 2020-05-20
Add integeration tests for sample opt tables

samanvp:
We modify the following two tests to verify the sample lookup optimal output tables:
 * platinum_partition_with_residual
 * platinum_partition_without_residual

We also modify the sharding config files so that both include a shard that does not match to any row. We expect these empty output tables to be deletes after we added that functionality in #596

### DIFF
--- a/gcp_variant_transforms/testing/data/sharding_configs/integration_with_residual.yaml
+++ b/gcp_variant_transforms/testing/data/sharding_configs/integration_with_residual.yaml
@@ -63,7 +63,7 @@
 -  output_table:
      table_name_suffix: "empty"
      regions:
-       - "does not match to any reference_name value"
+       - "nonexistence_chromosome"
      total_base_pairs: 249,240,615
 -  output_table:
      table_name_suffix: "all_remaining"
--- a/gcp_variant_transforms/testing/data/sharding_configs/integration_with_residual.yaml
+++ b/gcp_variant_transforms/testing/data/sharding_configs/integration_with_residual.yaml
@@ -61,6 +61,11 @@
        - "m"
      total_base_pairs: 57,227,415
 -  output_table:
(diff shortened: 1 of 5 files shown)


gcp_variant_transforms/testing/data/sharding_configs/integration_with_residual.yaml
```diff
@@ -60,6 +60,11 @@
        - "chrM"
        - "m"
      total_base_pairs: 57,227,415
+-  output_table:
+     table_name_suffix: "empty"
+     regions:
+       - "does not match to any reference_name value"
```
tneymanov, line 66: How does this comment work? Isn't region part going to be looking for it?

gcp_variant_transforms/testing/data/sharding_configs/integration_without_residual.yaml
```diff
@@ -28,3 +28,8 @@
      regions:
        - "chr1:200,000,000-999,999,999"
      total_base_pairs: 249,240,615
+-  output_table:
+     table_name_suffix: "empty"
+     regions:
+       - "does not match to any reference_name value"
```
tneymanov, line 34: same here

gcp_variant_transforms/testing/data/sharding_configs/integration_with_residual.yaml
samanvp: Modified to make it more clear I meant empty shard.

gcp_variant_transforms/testing/data/sharding_configs/integration_without_residual.yaml
samanvp: Modified to make it more clear I meant empty shard.

PR #601, opened 2020-05-25
Generate Sample Info table through avro file.

tneymanov:
I looked into generating schema from one source for table and for the avro file, but avro converter needs TableSchema variable, while bq command needs actual json. We should discuss if it makes sense to combine them.

Also, all my attempts to store DATETIME in BQ failed, unfortunately. I don't think that's possible. I used TIMESTAMP approach, which is fine, because BQ automatically converts the output into readable datetime that has UTC timezone. The problem is that gnomad has those fields as datetime, so from now on there is going to be a disconnect.

### DIFF
--- a/gcp_variant_transforms/transforms/sample_info_to_bigquery.py
+++ b/gcp_variant_transforms/transforms/sample_info_to_bigquery.py
@@ -14,15 +14,16 @@
 
 from typing import Dict, Union  # pylint: disable=unused-import
 
-from datetime import datetime
+import time
 
 import apache_beam as beam
+import avro
 
 from gcp_variant_transforms.beam_io import vcf_header_io  # pylint: disable=unused-import
 from gcp_variant_transforms.beam_io import vcf_parser
-from gcp_variant_transforms.libs import bigquery_util
 from gcp_variant_transforms.libs import hashing_util
 from gcp_variant_transforms.libs import sample_info_table_schema_generator
+from gcp_variant_transforms.libs import schema_converter
 
 SampleNameEncoding = vcf_parser.SampleNameEncoding
 _DATETIME_FORMAT = "%Y-%m-%d %H:%M:00.0"
@@ -36,7 +37,7 @@ class ConvertSampleInfoToRow(beam.DoFn):
     self._sample_name_encoding = sample_name_encoding
 
   def _get_now_to_minute(self):
-    return datetime.now().strftime(_DATETIME_FORMAT)
+    return int(time.time()) / 60 * 60000000
 
   def process(self, vcf_header):
     # type: (vcf_header_io.VcfHeader, bool) -> Dict[str, Union[int, str]]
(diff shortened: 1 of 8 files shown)


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -309,7 +309,7 @@ def _update_google_cloud_job_name(google_cloud_options, job_name):
 
 
 def _merge_headers(known_args, pipeline_args,
-                   pipeline_mode, annotated_vcf_pattern=None):
+                   pipeline_mode, avro_root_path, annotated_vcf_pattern=None):
```
samanvp, line 312: In this method, we create sample info table (line 335). We should move it to where we create all other tables before AVRO load step.

gcp_variant_transforms/transforms/sample_info_to_bigquery.py
```diff
@@ -71,22 +72,16 @@ def __init__(self, output_table_prefix, sample_name_encoding, append=False):
         sample_id would only use sample_name in to get a hashed name; otherwise
         both sample_name and file_name will be used.
     """
-    self._output_table = bigquery_util.compose_table_name(
-        output_table_prefix, bigquery_util.SAMPLE_INFO_TABLE_SUFFIX)
+    self._output_path = output_table_prefix
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/transforms/sample_info_to_bigquery.py
```diff
@@ -71,22 +72,16 @@ def __init__(self, output_table_prefix, sample_name_encoding, append=False):
...
     self._append = append
```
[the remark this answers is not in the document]
tneymanov: Done.

gcp_variant_transforms/transforms/sample_info_to_bigquery.py
```diff
@@ -71,22 +72,16 @@ def __init__(self, output_table_prefix, sample_name_encoding, append=False):
...
     self._sample_name_encoding = sample_name_encoding
-    self._schema = sample_info_table_schema_generator.generate_schema()
```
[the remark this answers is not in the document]
tneymanov: Sure, Done.

gcp_variant_transforms/transforms/sample_info_to_bigquery.py
```diff
@@ -71,22 +72,16 @@ def __init__(self, output_table_prefix, sample_name_encoding, append=False):
...
-                self._output_table,
-                schema=self._schema,
-                create_disposition=(
-                    beam.io.BigQueryDisposition.CREATE_IF_NEEDED),
-                write_disposition=(
-                    beam.io.BigQueryDisposition.WRITE_APPEND
-                    if self._append
-                    else beam.io.BigQueryDisposition.WRITE_TRUNCATE))))
+            | 'WriteToAvroFiles' >> beam.io.WriteToAvro(
+                self._output_path, self._avro_schema))
```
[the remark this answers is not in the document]
tneymanov: Yes, it does, we append it in vcf_to_bq, when initializing this ptransform.

gcp_variant_transforms/vcf_to_bq.py
tneymanov: Umm, disagree - we have headers here, we would need to recreate them, which would be an unnecessary computation. WDYT?

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -540,6 +539,7 @@ def run(argv=None):
             total_base_pairs, schema_file)
         logging.info('Integer range partitioned table %s was created.',
                      table_name)
+    suffixes.append(bigquery_util.SAMPLE_INFO_TABLE_SUFFIX)
```
samanvp, line 542: Somewhere around this line we need to create `sample_info` table:
```
bigquery_util.create_sample_info_table(known_args.output_table)
```

samanvp: I checked the failed tests and I found the culprit:
We treat `sample_info` table similar to other variant tables and that causes the failure (for example when we try to flatten it, it fails because it does not have a `call` column).

Here is the solution:
After AVRO load is done:
```
not_empty_variant_suffixes = load_avro.start_loading()
```
We need to remove `bigquery_util.SAMPLE_INFO_TABLE_SUFFIX` from both `suffixes` and `not_empty_variant_suffixes`.

gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -540,10 +537,14 @@ def run(argv=None):
...
         logging.info('Integer range partitioned table %s was created.',
                      table_name)
+    if not known_args.append:
+      bigquery_util.create_sample_info_table(known_args.output_table)
+    suffixes.append(bigquery_util.SAMPLE_INFO_TABLE_SUFFIX)
     load_avro = bigquery_util.LoadAvro(avro_root_path,
                                        known_args.output_table,
                                        suffixes, not known_args.append)
     not_empty_variant_suffixes = load_avro.start_loading()
+    not_empty_variant_suffixes.remove(bigquery_util.SAMPLE_INFO_TABLE_SUFFIX)
```
samanvp, line 547: You need to remove it from both this list and `suffixes`, otherwise we will end up with an extra table when `sample_lookup_optimized_output_table` is set (which should fail some of our tests where we expect 51 output tables).
Also for inputs such a gnomAD where there is no samples, `sample_info` table will be empty so this list will not have that element. I am not sure what is the `remove` behavior in that case.

PR #602, opened 2020-05-25
Store table schema in Avro location.

tneymanov:
Unfortunately, bq commands don't work if schema is on gs, so the logic has to be duplicated.

### DIFF
--- a/gcp_variant_transforms/vcf_to_bq.py
+++ b/gcp_variant_transforms/vcf_to_bq.py
@@ -403,6 +403,8 @@ def _run_annotation_pipeline(known_args, pipeline_args):
 
 
 def _write_schema_to_temp_file(schema, path):
+  if not path or not path.startswith('gs://'):
+    raise ValueError('Schema must be stored on a GS bucket.')
   schema_json = schema_converter.convert_table_schema_to_json_bq_schema(schema)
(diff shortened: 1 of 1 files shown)


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -402,11 +402,14 @@ def _run_annotation_pipeline(known_args, pipeline_args):
   return annotated_vcf_pattern
 
 
-def _write_schema_to_temp_file(schema):
+def _write_schema_to_temp_file(schema, path):
   schema_json = schema_converter.convert_table_schema_to_json_bq_schema(schema)
   schema_file = tempfile.mkstemp(suffix=_BQ_SCHEMA_FILE_SUFFIX)[1]
   with filesystems.FileSystems.create(schema_file) as file_to_write:
     file_to_write.write(schema_json)
+  gs_file_path = path + _BQ_SCHEMA_FILE_SUFFIX
```
samanvp, line 410: Do you want to add a simple assert `startswith('gs://')`? 

Not super strong check, however, this is what we do to validate flags:
https://github.com/googlegenomics/gcp-variant-transforms/blob/5ddabc6647bbe552b5c259a46d2286800630caf5/gcp_variant_transforms/options/variant_transform_options.py#L426

tneymanov: Done.

PR #603, opened 2020-05-27
Refactor bigquery_util

samanvp:
into 3 smaller modules to make it's maintanance more manageable.

PR #604, opened 2020-05-28
Rollback logic for sample opt tables

samanvp:
This PR makes the rollback logic to work with sample optimized tables. It also ensures that we properly rollback all newly created tables in case or exception at any stage of running vcf_to_bq.

Note: This PR is synced with #603

### DIFF
--- a/gcp_variant_transforms/vcf_to_bq.py
+++ b/gcp_variant_transforms/vcf_to_bq.py
@@ -95,6 +95,11 @@ _SHARDS_FOLDER = 'shards'
 _GCS_RECURSIVE_WILDCARD = '**'
 SampleNameEncoding = vcf_parser.SampleNameEncoding
 
+_newly_created_tables = []
+def _record_newly_created_table(full_table_id):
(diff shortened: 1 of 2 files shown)


gcp_variant_transforms/vcf_to_bq.py
```diff
@@ -601,4 +605,17 @@ def run(argv=None):
 
 if __name__ == '__main__':
   logging.getLogger().setLevel(logging.INFO)
-  run()
+  try:
+    run()
+  except Exception as e:
+    if _newly_created_tables:
```
tneymanov, line 611: I think it's possible to not have any tables in _newly_created_tables but still have append off. This would happen if errors happens before any table is generated.

samanvp: I agree, but it was getting too complicated to pass `known_args.append` here.
So I decided to show this warning just based on empty `_newly_created_tables` (which means in some cases it's wrong).

PR #605, opened 2020-05-31
Keep empty tables after AVRO copy stage

samanvp:
Based on feedback from one of our customers, we decided to keep emtpy tables after AVRO copy stage. This is mainly for supporting cases were first a VCF file is imported and then we keep appending new rows to the these newly created tables by running Variant Transforms with `--append` flag. In such cases, if the first VCF file does not have variants in all chromosome we should still keep all empty output tables, so future append operation will be possible. Otherwise the append run will fail due to the following check:
```
ValueError: Table PROJECT_ID:DATASET_ID.BASE_TABLE_ID__chrXX does not exist, cannot append to it.
```
Where `chrXX` is the empty table in the fist run.

Note, this PR is technically rolls-back the change in #596


cloudbuild_CI.yaml
```diff
@@ -41,10 +41,7 @@ steps:
       - '--skip_build'
       - '--project ${PROJECT_ID}'
       - '--image_tag ${COMMIT_SHA}'
-      - '--run_unit_tests'
-      - '--run_preprocessor_tests'
-      - '--run_bq_to_vcf_tests'
-      - '--run_all_tests'
+      - '--run_presubmit_tests'
```
tneymanov, line 44: Restore.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -178,6 +181,42 @@ def table_exist(client, project_id, dataset_id, table_id):
...
+  query_job = client.query(query)
+  num_retries = 0
+  while True:
+    try:
+      results = query_job.result(timeout=300)
+    except TimeoutError as e:
+      logging.warning('Time out waiting for query: %s', query)
+      if num_retries < BQ_NUM_RETRIES:
+        num_retries += 1
+        time.sleep(90)
```
tneymanov, line 197: Lets have a const variable for timeout amount.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -178,6 +181,42 @@ def table_exist(client, project_id, dataset_id, table_id):
...
+      else:
+        raise e
+    else:
+      if results.total_rows == 1:
+        break
+      else:
+        logging.error('Query did not returned expected # of rows: %s', query)
+        if num_retries < BQ_NUM_RETRIES:
+          num_retries += 1
+          time.sleep(90)
```
tneymanov, line 207: Same.

gcp_variant_transforms/libs/bigquery_util.py
```diff
@@ -178,6 +181,42 @@ def table_exist(client, project_id, dataset_id, table_id):
...
+        else:
+          raise ValueError(
+            'Expected 1 row in query result, got {}'.format(results.total_rows))
+
+  row = list(results)[0]
+  col_names = row.keys()
+  if set(col_names) != {'num_rows'}:
```
tneymanov, line 214: Since output column is repeated in several places, lets put it into a separate const var as well.

gcp_variant_transforms/libs/partitioning.py
```diff
@@ -79,13 +79,22 @@ def __init__(self, base_table_id, suffixes):
...
-    self._schema_table_id = bigquery_util.compose_table_name(
-        self._base_table, suffixes[0])
     self._column_names = []
     self._sub_fields = []
     self._client = bigquery.Client(project=self._project_id)
 
+    self._find_one_non_empty_table()
+
+  def _find_one_non_empty_table(self):
+    # Any non empty input table can be used as the source for schema extraction.
```
tneymanov, line 89: Doesn't empty table also have the required schema?

cloudbuild_CI.yaml

PR #612, opened 2020-06-11
pin pysam version to < 0.16.0

samanvp:
Due to issues related to this integration test:
`test_allow_malformed_records`

It seems pysam gets stock in some kind of halt state when it parses that malformed VCF file (which is missing `start_position` value in one of its lines).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ PYSAM_DEPENDENCY_COMMANDS = [
      'zlib1g-dev']
 ]
 
-PYSAM_INSTALLATION_COMMAND = ['pip', 'install', 'pysam>=0.15.3']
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -27,7 +27,7 @@
      'zlib1g-dev']
 ]
 
-PYSAM_INSTALLATION_COMMAND = ['pip', 'install', 'pysam>=0.15.3']
+PYSAM_INSTALLATION_COMMAND = ['pip', 'install', 'pysam<0.16.0']
```
tneymanov, line 30: How about ==0.15.3 instead? we need that for UTF-8 compatibility.

samanvp: Discussed offline, modified to have both limitations.

PR #614, opened 2020-06-12
Update our base docker images

samanvp:
`golang` from 1.11 to latest (1.13)
`cloud-sdk:slim` from 286.0.0 to latest (296.0.1)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,11 +15,11 @@
 # To build a new docker image, run the following from the root source dir:
 # $ docker build . -f docker/Dockerfile -t $IMAGE_NAME
 
-FROM golang:1.11
+FROM golang:latest
 
 RUN go get -ldflags '-extldflags "-fno-PIC -static"' -buildmode pie -tags 'osusergo netgo static_build' github.com/googlegenomics/pipelines-tools/pipelines
 
-FROM google/cloud-sdk:286.0.0-slim
+FROM google/cloud-sdk:slim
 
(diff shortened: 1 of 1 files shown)


samanvp:
We need to use the latest base images before our final release.

The reason that I decided to switch to `:latest` instead of pinning to a specific version was last week's failure of [this package](https://pypi.org/project/google-auth/#history). Version 1.17.1 had an issue and they released 1.17.2 almost 12 hours after the previous one to fix the issue. 

The fact that we are relying on many pip packages without pining them to a specific version, made me think. For example, here are only `google` images we have:
```
$ pip freeze | grep google
google-api-core==1.20.0
google-api-python-client==1.7.11
google-apitools==0.5.28
google-auth==1.16.1
google-auth-httplib2==0.0.3
google-cloud-bigquery==1.24.0
google-cloud-bigtable==1.0.0
google-cloud-core==1.3.0
google-cloud-datastore==1.7.4
google-cloud-dlp==0.13.0
google-cloud-language==1.3.0
google-cloud-pubsub==1.0.2
google-cloud-spanner==1.13.0
google-cloud-storage==1.28.1
google-cloud-videointelligence==1.13.0
google-cloud-vision==0.42.0
google-resumable-media==0.5.1
googleapis-common-protos==1.52.0
grpc-google-iam-v1==0.12.3
```
 I started questioning our earlier decision to pin our base images. Base images go though much stronger tests before release that's why relying on the latest version is not as risky.

PR #615, opened 2020-06-16
Modify sample optimized tables

samanvp:
To make their schema more consistent with variant optimized tables.

In this PR, instead of flattening all fields under `call` column, we keep them as one column with type `STRUCT`. Under this new change the only differences between variant optimized tables and sample optimized tables are:
* call column is `REPEATED` for variant tables but not for sample tables.
* sample tables have a new `INT64` column (added before `call` column) named `sample_id` which is used to partition sample tables based on its value.

Note that `sample_id` column is redundant and its value is exactly equal to `call.sample_id`. Even though this redundant column makes our tables larger (higher storage cost) but similarity of sample optimized tables and variant optimized table is a desirable property that worth this extra cost.

We adopted this schema based on feedback from one of our users. Thanks Micheal!


cloudbuild_CI.yaml
```diff
@@ -44,7 +44,7 @@ steps:
       - '--run_unit_tests'
       - '--run_preprocessor_tests'
       - '--run_bq_to_vcf_tests'
-      - '--run_all_tests'
+      - '--run_presubmit_tests'
```
tneymanov, line 47: Hmm, were presubmit ones disabled?

Restore all_tests please.

PR #609, opened 2020-06-04
Add column descriptions to sample optimized tables

samanvp:
In this PR we copy column descriptions from variant optimized tables to sample optimized schema json file.

PR #616, opened 2020-06-17
Modify the way we generate schema of sample optimized tables

samanvp:
Instead of the current method:
 * a query to select 1 row and write it to a temporary table
  and extract the schema of the temporary table

In this PR we extract the schema of sample optimized tables by modifying the schema of variant optimized tables. This was not possible before #615 

This new method has 2 main benefits:
 1) It's simpler than previous method.
 2) It also copies the column description (so there is no need for #609 )

### DIFF
--- a/gcp_variant_transforms/libs/partitioning.py
+++ b/gcp_variant_transforms/libs/partitioning.py
@@ -15,6 +15,7 @@
 """Utilities to create integer range partitioned BigQuery tables."""
 
 from concurrent.futures import TimeoutError
+import json
 import logging
 import math
 import os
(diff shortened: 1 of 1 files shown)


gcp_variant_transforms/libs/partitioning.py
```diff
@@ -182,59 +181,64 @@ def _copy_to_flatten_table(self, output_table_id, cp_query):
...
-      logging.info('Successfully extracted the schema of flatten table.')
-    if bigquery_util.delete_table(full_table_id) == 0:
-      logging.info('Successfully deleted temporary table: %s', full_table_id)
-    else:
-      logging.error('Was not able to delete temporary table: %s', full_table_id)
-    return result == 0
+    full_table_id = '{}.{}.{}'.format(
+        self._project_id, self._dataset_id, self._schema_table_id)
+    try:
+      variant_table = self._client.get_table(full_table_id)
```
tneymanov, line 231: Do we need to get schema from BQ table? Don't we already have it computed at this point?

samanvp: Probably we could get it from BEAM, however, since we are creating sample optimized tables from variant optimized tables I can imagine in future we turn this into an independent functionality. That's why I thought we better not to have dependency on other parts of code.

PR #620, opened 2020-06-23
Avoid creating tables when we load AVRO files

samanvp:
In the current implementation, we create partitioned tables using `bq mk table ...` command before we load AVRO files. Thus we need to ensure table exists before we start loading AVRO files into it.

Note, if we resolve issue #582  adding `create_disposition` flag would not be needed.


cloudbuild_CI.yaml
```diff
@@ -44,7 +44,7 @@ steps:
       - '--run_unit_tests'
       - '--run_preprocessor_tests'
       - '--run_bq_to_vcf_tests'
-      - '--run_all_tests'
+      - '--run_presubmit_tests'
```
tneymanov, line 47: Fix.

PR #622, opened 2020-06-24
Modify our docker entrypoint

samanvp:
to share `--temp_location` flag between Dataflow and luncher VM. This will simplify running VT using our docker image.

### DIFF
--- a/docs/bigquery_to_vcf.md
+++ b/docs/bigquery_to_vcf.md
@@ -15,36 +15,41 @@ source.
 
 Run the script below and replace the following parameters:
 
-* `GOOGLE_CLOUD_PROJECT`: This is your Google Cloud project ID where the job
+* Dataflow's [required inputs](https://cloud.google.com/dataflow/docs/guides/specifying-exec-params#configuring-pipelineoptions-for-execution-on-the-cloud-dataflow-service):
+  * `GOOGLE_CLOUD_PROJECT`: This is your Google Cloud project ID where the job
   should run.
+  * `GOOGLE_CLOUD_REGION`: You must choose a geographic region for Cloud Dataflow
+  to process your data, for example: `us-west1`. For more information please refer to
+  [Setting Regions](docs/setting_region.md).
+  * `TEMP_LOCATION`: This can be any folder in Google Cloud Storage that your
+  project has write access to. It's used to store temporary files and logs
+  from the pipeline.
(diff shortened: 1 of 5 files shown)


samanvp:
@bgood please let us know what you think.

docs/setting_region.md
```diff
@@ -15,41 +15,46 @@ are located in the same region:
...
-to restrict your processing job to Europe, update the region as follows:
+to restrict your processing job to Europe west, set the region as follows:
 
 ```bash
 COMMAND="/opt/gcp_variant_transforms/bin/vcf_to_bq ...
 
 docker run gcr.io/cloud-lifesciences/gcp-variant-transforms \
   --project "${GOOGLE_CLOUD_PROJECT}" \
-  --region "${GOOGLE_CLOUD_REGION}" \
+  --region "europe-west1" \
```
tneymanov, line 31: remove quotes around europe-west1

docs/setting_region.md
```diff
@@ -15,41 +15,46 @@ are located in the same region:
...
 ```
 
-Note that values of `--project` and `--region` flags will be automatically
-passed as `COMMAND` args in [`piplines_runner.sh`](docker/pipelines_runner.sh).
-Alternatively, you can set your default region using the following command:
+Note that values of `--project`, `--region`, and `--temp_location` flags will be automatically
+passed as `COMMAND` flags in [`piplines_runner.sh`](docker/pipelines_runner.sh).
+
+Instead of setting `--region` flag for each run, you can set your default region using the following command.
+In this case you do not need to set the `--region` flag any more. For more
```
tneymanov, line 40: some nits:
In this case, you will not need to set the `--region` flag any more. For more
information, please refer to this [cloud SDK page](https://cloud.google.com/sdk/gcloud/reference/config/set).

Also, should we add that --project also won't be needed?

docs/bigquery_to_vcf.md
```diff
@@ -122,7 +123,8 @@ python -m gcp_variant_transforms.bq_to_vcf \
   --output_file gs://BUCKET/loaded_file.vcf \
   --representative_header_file gs://BUCKET/representative_header_file.vcf \
   --project "${GOOGLE_CLOUD_PROJECT}" \
-  --temp_location gs://BUCKET/temp \
+  --region "${GOOGLE_CLOUD_REGION}" \
```
tneymanov, line 126: Can't comment on the line above, but I think we still need temp_location for DirectRunner, no?

PR #623, opened 2020-06-24
Adding sub network and private address as options to the gcp-variant-transforms container

bgood:
In this PR, we've added optional parameters for for sub network and private addresses only.  These are needed launch jobs in projects that have custom networks and don't allow for public IPs.

### DIFF
--- a/docker/pipelines_runner.sh
+++ b/docker/pipelines_runner.sh
@@ -22,7 +22,7 @@ set -euo pipefail
 #################################################
 function parse_args {
   # getopt command is only for checking arguments.
-  getopt -o '' -l project:,temp_location:,docker_image:,region:,zone:,network:,subnetwork:,privateAddress: -- "$@"
+  getopt -o '' -l project:,temp_location:,docker_image:,region:,subnetwork:,privateAddress: -- "$@"
   while [[ "$#" -gt 0 ]]; do
     case "$1" in
       --project)
@@ -39,14 +39,6 @@ function parse_args {
 
       --region)
         region="$2"
-        ;;
-
-      --zone)
-        zone="$2"
-        ;;
-
-      --network)
-        network="$2"
         ;;
 
       --subnetwork)
@@ -76,8 +68,6 @@ function main {
   vt_docker_image="${vt_docker_image:-gcr.io/cloud-lifesciences/gcp-variant-transforms}"
   region="${region:-$(gcloud config get-value compute/region)}"
   temp_location="${temp_location:-''}"
-  zone="${zone:-}"
-  network="${network:-}"
   subnetwork="${subnetwork:-}"
   privateAddress="${privateAddress:-}"
   extra_args=""
@@ -94,30 +84,20 @@ function main {
     exit 1
   fi
 
-  if [[ ! -v command ]]; then
-    echo "Please specify a command to run Variant Transforms."
-    exit 1
-  fi
-
   # Build up the extra args is they are provided
-  if [[ ! -z "${zone}" ]]; then
-    echo "Adding --zone=${zone} to extra_args"
-    extra_args="${extra_args} --zone=${zone}"
-  fi
-
-  if [[ ! -z "${network}" ]]; then
-    echo "Adding --network=${network} to extra_args"
-    extra_args="${extra_args} --network=${network}"
-  fi
-
   if [[ ! -z "${subnetwork}" ]]; then
-    echo "Adding --subnetwork=${subnetwork} to extra_args"
-    extra_args="${extra_args} --subnetwork=${subnetwork}"
+    echo "Adding --subnetwork ${subnetwork} to extra_args"
+    extra_args="${extra_args} --subnetwork ${subnetwork}"
   fi
 
   if [[ ! -z "${privateAddress}"  && "${privateAddress}" == "true" ]]; then
     echo "Adding --private-address to extra_args"
     extra_args="${extra_args} --private-address"
+  fi
+
+  if [[ ! -v command ]]; then
+    echo "Please specify a command to run Variant Transforms."
+    exit 1
   fi
 
   pipelines --project "${google_cloud_project}" run \
--- a/docker/pipelines_runner.sh
+++ b/docker/pipelines_runner.sh
@@ -22,7 +22,7 @@ set -euo pipefail
 #################################################
 function parse_args {
   # getopt command is only for checking arguments.
-  getopt -o '' -l project:,temp_location:,docker_image:,region: -- "$@"
+  getopt -o '' -l project:,temp_location:,docker_image:,region:,zone:,network:,subnetwork:,privateAddress: -- "$@"
   while [[ "$#" -gt 0 ]]; do
     case "$1" in
       --project)
@@ -39,6 +39,22 @@ function parse_args {
 
       --region)
         region="$2"
+        ;;
+
+      --zone)
+        zone="$2"
+        ;;
+
+      --network)
+        network="$2"
+        ;;
+
+      --subnetwork)
+        subnetwork="$2"
+        ;;
+
+      --privateAddress)
+        privateAddress="$2"
         ;;
 
       *)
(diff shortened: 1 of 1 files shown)


docker/pipelines_runner.sh
```diff
@@ -88,7 +110,7 @@ function main {
     --machine-type "g1-small" \
     --pvm-attempts 0 \
     --attempts 1 \
-    --disk-size 10
+    --disk-size 10 ${extra_args}
```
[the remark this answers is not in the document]
samanvp: I did a quick check and it seems Dataflow has `--use_public_ips` (or ` --no_use_public_ips`). 
So if we want to do what I suggested we need to construct two `extra_args`: one for Dataflow one for pipelines-tool.
If you think this makes things more complex and does not help with users please feel free to ignore.

docker/pipelines_runner.sh
```diff
@@ -41,6 +41,14 @@ function parse_args {
         region="$2"
         ;;
 
+      --subnetwork)
+        subnetwork="$2"
+        ;;
+
+      --privateAddress)
+        privateAddress="$2"
```
samanvp, line 49: I don't like the fact that this flag is CamleCase while all others are snake_case.
I think it's because [Dataflow](https://cloud.google.com/dataflow/docs/guides/specifying-exec-params) uses snake_case while [pipelines-tools](https://github.com/googlegenomics/pipelines-tools/blob/master/pipelines/internal/commands/run/run.go) uses CamleCase.

How about this: Let's stick to the naming of Dataflow. So the input flag that we parse will be `[no_]use_public_ips` and when we construct the `extra_args` we use `privateAddress`?

bgood: Yes that bothered me as well, however, I matched the casing on pipeline-tools because Dataflow uses different names and casing depending on Python or Java (https://cloud.google.com/dataflow/docs/guides/specifying-networks#java).  

Since it is inconsistent in Dataflow and pipeline-tools, how about we use `public_ips true/false` for the input argument and shape it as needed in the script?

docker/pipelines_runner.sh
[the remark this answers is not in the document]
bgood: Yes, unfortunately it is different and per my other comment is it even different for Java vs. Python.

I didn't add them or do the extra parsing since users are putting much of the Dataflow command together already.  That said, I would be happy to add a second optional set of arguments to the script for Dataflow if we can resolve the Java vs. Python issue.

docker/pipelines_runner.sh
```diff
@@ -57,9 +65,12 @@ function main {
   parse_args "$@"
 
   google_cloud_project="${google_cloud_project:-$(gcloud config get-value project)}"
-  vt_docker_image="${vt_docker_image:-gcr.io/cloud-lifesciences/gcp-variant-transforms:${COMMIT_SHA}}"
+  vt_docker_image="${vt_docker_image:-gcr.io/cloud-lifesciences/gcp-variant-transforms}"
   region="${region:-$(gcloud config get-value compute/region)}"
   temp_location="${temp_location:-''}"
+  subnetwork="${subnetwork:-}"
+  privateAddress="${privateAddress:-}"
+  extra_args=""
```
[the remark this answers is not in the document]
bgood: I like it, will update.

docker/pipelines_runner.sh
bgood: Also using a flag and a true/false value simplifies the args parsing.

samanvp: To be honest I think adding a third naming convention only makes things more complicated. I think I prefer to stick with Dataflow Python SDK's standard, let's use `use_public_ips`.

And also true/false value to simplify the parsing is a good idea.

docker/pipelines_runner.sh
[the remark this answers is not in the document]
bgood: In the example usage I have seen the user is needing to build up a fair bit of the Dataflow command as it is, so it is easy for users to assume they can extra arguments to the Dataflow command as they need.  I think it would be nice to pass `temp_location`, `subnetwork` and `use_public_ip`, however, it can be confusing.

In fact at first not understanding the flow of Container -> VM -> Dataflow and the arguments made it tricky to understand what the problem was in the case of having a custom network because we thought having the subnet flag in the Dataflow command was enough.

docker/pipelines_runner.sh
bgood: Sounds good, will also make that change.

docker/pipelines_runner.sh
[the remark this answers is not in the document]
samanvp: Thanks for the feedback. So based on your suggestion #622 is a good idea.

After your PR merged we can add a small section to our docs and explain these two new flags. In that example we can show these flags need to be added to both Dataflow args and launcher VM args. 
If you will, I'd appreciate if you help us with that change too. Specially the wording related to these requirements to be HIPAA compliant.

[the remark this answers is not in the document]
bgood: Yes, I think it makes things easier, so I'll add some logic to this PR for setting the corresponding Dataflow flags.

Also happy to help with other PR.

bgood:
Updates made as discussed.  Let me know what you think especially regarding the `subnetwork` flag and the assembling of the Dataflow and pipelines-tool formatted versions of that flag.

docker/pipelines_runner.sh
```diff
@@ -73,13 +85,26 @@ function main {
     exit 1
   fi
 
+  # Build up the extra args is they are provided
```
samanvp, line 88: nit: 
s/extra/optional/ 
s/is/if/

docker/pipelines_runner.sh
```diff
@@ -73,13 +85,26 @@ function main {
...
+  if [[ ! -z "${subnetwork}" ]]; then
+    echo "Adding --subnetwork ${subnetwork} to optional_args"
+    pt_optional_args="${pt_optional_args} --subnetwork projects/${project}/regions/${region}/subnetworks/${subnetwork}"
+    df_optional_args="${df_optional_args} --subnetwork https://www.googleapis.com/compute/v1/projects/${project}/regions/${region}/subnetworks/${subnetwork}"
+  fi
+
+  if [[ ! -z "${use_public_ips}"  && "${use_public_ips}" == "true" ]]; then
```
samanvp, line 95: Shouldn't  this condition be `use_public_ips == "false"` ?

docker/pipelines_runner.sh
```diff
@@ -73,13 +85,26 @@ function main {
...
+    echo "Adding --private-address and --no_use_public_ips to optional_args"
```
samanvp, line 96: s/private-address/privateAddress/

docker/pipelines_runner.sh
```diff
@@ -73,13 +85,26 @@ function main {
     exit 1
   fi
 
+  # Build up the extra args is they are provided
+  if [[ ! -z "${subnetwork}" ]]; then
+    echo "Adding --subnetwork ${subnetwork} to optional_args"
+    pt_optional_args="${pt_optional_args} --subnetwork projects/${project}/regions/${region}/subnetworks/${subnetwork}"
+    df_optional_args="${df_optional_args} --subnetwork https://www.googleapis.com/compute/v1/projects/${project}/regions/${region}/subnetworks/${subnetwork}"
```
samanvp, line 92: I have never used subnetwork so I am not entirely sure why these strings are different.
Also, I am not sure if the value we expect to receive by --subnetwork flag (and then plug it inside those two strings) is a usual way to provide this flag.

I will leave this to your expertise.

docker/pipelines_runner.sh
bgood: Yes it should, my bad.

Good catch!

docker/pipelines_runner.sh
bgood: The external flag for the pipelines tool is `private-address` but the var name is `privateAddress`.

https://github.com/googlegenomics/pipelines-tools/blob/master/pipelines/internal/commands/run/run.go#L159

docker/pipelines_runner.sh
bgood: The Dataflow param format is different from the GCE format.  My thought was to hide that in the script and just allow users to private the "short name" for the subnetwork.

We could very easily take the GCE version and perpend with the `https://www......` for Dataflow.  In my opinion there isn't much difference from an end-user perspective.

docker/pipelines_runner.sh
bgood: Done

docker/pipelines_runner.sh
bgood: Done.

docker/pipelines_runner.sh
samanvp: Sounds good.

docker/pipelines_runner.sh
samanvp: my bad, thanks for correcting me.

PR #627, opened 2020-06-29
Modify how we send subnetwork to pipelines-tools

samanvp:
It seems pipelines-tools only need subnetwork value and it generates the correct format itself:

`projects/${google_cloud_project}/regions/${region}/subnetworks/${subnetwork}`

### DIFF
--- a/docker/pipelines_runner.sh
+++ b/docker/pipelines_runner.sh
@@ -104,7 +104,7 @@ function main {
 
   if [[ ! -z "${subnetwork}" ]]; then
     echo "Adding --subnetwork ${subnetwork} to optional_args"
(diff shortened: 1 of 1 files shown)


samanvp:
FYI @bgood 
Following #623 we received feedback from one of our users indicating that pipelines-tools failed to create the instance because the subnetwork flag was set to:
```
projects/${google_cloud_project}/regions/${region}/subnetworks/projects/${google_cloud_project}/regions/${region}/subnetworks/${subnetwork}
```

It seems this is because pipelines tools also constructs the string for subnetwork. To solve this we need to send only `${subnetwork}` value as input flag to pipelines tools.

bgood:
You will also need to update the `df_optional_args` to match that change.

PR #628, opened 2020-06-29
Revert "Modify how we send subnetwork to pipelines-tools (#627)"

samanvp:
This reverts commit 527e1d0688fdab4502a2bcdd4b0aa9309e6d5531.


samanvp:
Sorry Tural, it turns out the code was correct the way it was. User was not running `docker run ...` command with correct inputs.

PR #638, opened 2020-07-09
Replace RemoveDuplicates with Distinct

samanvp:
While running bq_to_vcf we get the following message:
```
BeamDeprecationWarning: RemoveDuplicates is deprecated since 2.12. Use Distinct instead.
```


cloudbuild_CI.yaml
```diff
@@ -44,7 +44,7 @@ steps:
       - '--run_unit_tests'
       - '--run_preprocessor_tests'
       - '--run_bq_to_vcf_tests'
-      - '--run_all_tests'
+      - '--run_presubmit_tests'
```
tneymanov, line 47: Restore.

PR #645, opened 2020-07-20
Apply initial setup changes for Python3 migration.

tneymanov:


docker/Dockerfile
```diff
@@ -45,30 +45,31 @@ RUN apt-get update && apt-get install -y \
...
 # Install dependencies.
-RUN python -m pip install --upgrade pip && \
-    python -m pip install --upgrade virtualenv && \
-    virtualenv /opt/gcp_variant_transforms/venv && \
-    . /opt/gcp_variant_transforms/venv/bin/activate && \
+RUN 
+RUN python3 -m venv /opt/gcp_variant_transforms/venv3 && \
+    . /opt/gcp_variant_transforms/venv3/bin/activate && \
     cd /opt/gcp_variant_transforms/src && \
+    python -m pip install --upgrade wheel \
```
samanvp, line 56: Why we need `wheel`? I saw we also added it to the Travis yaml file.

tneymanov: So in py3 pip doesn't build virtualenv if wheel is missing. Im not entirely sure why this happens, but this is a common issue

https://stackoverflow.com/questions/14296531/what-does-error-option-single-version-externally-managed-not-recognized-ind

Note that just install upgrading setuptools didn't work, as well as adding wheel to setup.py.

PR #648, opened 2020-08-12
Add Life Sciences API location flag

samanvp:
To enable users to spceficy where the Life Sciences API is running and scheduling the jobs from.
This isn't where the data is processed, just some of the metadata like the commands run, input/output names, etc.
You can see the currently available locations here:
https://cloud.google.com/life-sciences/docs/concepts/locations


samanvp:
@tneymanov  I manually tested image made with this PR and the code work as intended.

tneymanov (approved):

PR #656, opened 2020-09-08
Set output-interval in pipelines_runner.sh to 60s.

mbookman:
The default value for the pipelines-tool's --output-interval is 0 (see [run.go](https://github.com/googlegenomics/pipelines-tools/blob/41825beb7587be6e5dcd364f2fc19ca64bd838cc/pipelines/internal/commands/run/run.go)):

```
  outputInterval = flags.Duration("output-interval", 0, "if non-zero, specifies the time interval for logging output during runs")
```

Thus without a value for `--output-interval`, the runner log is not pushed to the GCS `--temp_location` until the job completes. This makes `gcp-variant-transforms` more opaque and harder to monitor and debug.


docker/pipelines_runner.sh
```diff
@@ -117,6 +117,7 @@ function main {
   pipelines --project "${google_cloud_project}" run \
     --command "/opt/gcp_variant_transforms/bin/${command} ${df_required_args} ${df_optional_args}" \
     --output "${temp_location}"/runner_logs_$(date +%Y%m%d_%H%M%S).log \
+    --output-interval 60s \
```
samanvp, line 120: nit: I am not sure if this can have any negative impact on running the main task, specially because we are using a very small machine for this worker (`g1-small`).

What do you think about modifying it to 3 or 5 minutes intervals?

samanvp (approved):
Except for a small nit this PR LGTM.

docker/pipelines_runner.sh
mbookman: I think your update to 200 seconds is fine. LGTM.

samanvp: Thanks @mbookman for your reply and also for your contribution.

I am merging this PR now.

PR #663, opened 2020-09-10
Large threshold reduce to 500

samanvp:
Reduce the _LARGE_DATA_THRESHOLD from 50,000 to 500.
After deprecating  `--optimize_for_large_inputs` we need to modify this threshold to be more conservative about what we call large.


tneymanov (approved):
Thanks Saman, make sure to sync to resolve the errors.

PR #670, opened 2020-09-18
Fix README.md to remove dev update and fix broken URL.

tneymanov:


samanvp (approved):
Thanks for bringing these issues to our attention @moschetti

PR #671, opened 2020-09-19
Update Beam

samanvp:
from 2.22.0 to 2.24.0


tneymanov (approved):
Start CB on my project, please don't submit until it's green.

PR #672, opened 2020-09-21
Fix venv docker issue.

### LINKED ISSUE #666: Fix deploy_and_run_tests.sh to be locally runnable.
Since January 2020, the script was not locally runnable for unknown reasons and required Cloud Build launch to run integration tests.

Additionally in #664 PR installation of local virtual environment has been removed, in favor of using the virtualenv generated in Dockerfile, due to inability to setup PySam library.

Research both issues and make deploy_and_run_tests script to be runnable locally.

tneymanov:
Applied the suggested workaround in https://github.com/pypa/virtualenv/issues/1342 to fix venv errors caused when 'deactivate' command invoked.

Once deactivate worked, assumed that the change also fixes #666 , which after some code tweaking, it did. CB passed.

### DIFF
--- a/deploy_and_run_tests.sh
+++ b/deploy_and_run_tests.sh
@@ -188,7 +188,9 @@ parse_args() {
 # Deactivates virtualenv, removes its directory, and deletes the image.
 #################################################
 clean_up() {
+  color_print "Removing integration test environment ${temp_dir}" "${GREEN}"
   deactivate
+  rm -rf "${temp_dir}"
   if [[ -z "${keep_image}" ]]; then
     # TODO(bashir2): Find a way to mark these images as temporary such that they are
     # garbage collected automatically if the test fails before this line.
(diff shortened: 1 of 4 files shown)


deploy_and_run_tests.sh
```diff
@@ -224,11 +226,26 @@ if [[ -z "${skip_build}" ]]; then
       --substitutions _CUSTOM_TAG_NAME="${image_tag}" .
 fi
 
-source /opt/gcp_variant_transforms/venv3/bin/activate;
+# Running integration tests in a temporary virtualenv
+temp_dir="$(mktemp -d)"	
+color_print "Setting up integration test environment in ${temp_dir}" "${GREEN}"
+# Since we have no prompt we need to disable prompt changing in virtualenv.
+export VIRTUAL_ENV_DISABLE_PROMPT="something"
```
samanvp, line 233: It seems this was another hack for old venv, do we still need this?

tneymanov: As per offline conversation, this is used to disable the prompt tag.
https://stackoverflow.com/questions/10406926/how-do-i-change-the-default-virtualenv-prompt/20026992

PR #674, opened 2020-09-22
Fix google-cloud-storage import.

tneymanov:


samanvp (approved):
Thanks for finding this.

PR #675, opened 2020-09-24
Enable users to set --service_account

samanvp:
and pass that value to pipelines API so that the newly created VM will be using that service account.


samanvp:
@tneymanov I manually tested image made with this PR and the code work as intended.

tneymanov (approved):

PR #678, opened 2020-10-01
Add gnomAD sample queries notebook

moschetti:
Also includes folder & readme structure to support future sample queries.


sample_queries/README.md
```diff
@@ -0,0 +1,13 @@
+# Sample Queries
+
+This folder contains sample queries showing the functionality of querying a
+dataset in BigQuery which was imported with Variant Transforms.
+
```
samanvp, line 5: How about adding this paragraph? Please feel free to modify it as you see suited.

By providing these queries we are hoping researchers can better utilize their BigQuery tables. Also this will provide a way for the community to contribute by modifying these queries or suggesting new ones. We highly encourage all users to send us their queries so we can share them here with all other researchers. Please feel free to [submit an issue](https://github.com/googlegenomics/gcp-variant-transforms/issues) or contact us via our public mailing list <gcp-life-sciences-discuss@googlegroups.com>.

samanvp (commented):
The only real changed needed is to move this directory under docs/. @tneymanov and I discussed this earlier and we think that would be a better long term solution.

I also added two nits.

moschetti:
> The only real changed needed is to move this directory under docs/. @tneymanov and I discussed this earlier and we think that would be a better long term solution.
> 
> I also added two nits.

moved & addressed nits

samanvp (approved):

samanvp:
I am going to merge this PR now, thanks @moschetti

PR #649, opened 2020-08-12
Add location flag docs

moschetti:
Add docs for `--location` flag added in https://github.com/googlegenomics/gcp-variant-transforms/pull/648, as well as docs for `--subnetwork` & `--use_public_ips`

PR #682, opened 2020-10-05
Add location flag to documentation

moschetti:
Replaces googlegenomics/gcp-variant-transforms#649


samanvp (approved):
Thanks again Andrew.

PR #683, opened 2020-10-07
Fix typos and update narrative in gnomad sample queries notebook.

mbookman:


samanvp (approved):
Thanks @mbookman for the suggestions.

PR #633, opened 2020-07-02
Set 1-based coordinate as our default

samanvp:
Variant Transforms was only able to generate BigQuery tables in 0-based coordinate. We added ability to choose between 0-based and 1-based in #586.
We decided to modify our default behavior to generate 1-based variants in the output tabels.

PR #686, opened 2020-10-08
Change the default value of use_1_based_coordinate to False

samanvp:
Based on the feedback from our users and also to keep all our docs, tutorials, and queries consistent with VT before v0.10.0 launch, we change the default behavior back to 0-based.

Reverts #633


samanvp:
FYI @dglazer

tneymanov (approved):

PR #688, opened 2020-10-08
Fix query 2.1 and update notebook name

moschetti:
- The condensed results block in Query 2.1 had invalid name, fixed that
- Changed notebook name to match header 1
- Re-ran everything from a new project so the query bytes processed are accurate (not 0)


samanvp (approved):
I am pretty sure I modified the name (without the intention). I will ask you offline how can I avoid this in future.

Thank you!

PR #689, opened 2020-10-15
Drop hom_ref calls in sample optimized tables

samanvp:
In order to make the size of sample optimized tables more manageable we are dropping homozygous reference calls. This means sample optimized tables will only include:
* variants
* no calls


samanvp:
I ran multiple manual tests and confirmed the modified query is generating the correct output.

I will update docs in a follow up PR.

tneymanov (approved):
LGTM

moschetti:
Per conversations with customers, this change would have a negative impact on the workflow. As the genotype probability could change with increasing number of samples based on the statistics of that variant. Therefore rejecting this PR as the benefits would not outweigh the potential impact.

PR #693, opened 2020-10-21
Add an option to supply custom_sample_info_table.

tneymanov:


gcp_variant_transforms/options/variant_transform_options.py
```diff
@@ -658,19 +664,27 @@ def validate(self, parsed_args, client=None):
...
-
-    if not bigquery_util.table_exist(client, project_id, dataset_id,
-                                     sample_table_id):
+    if parsed_args.custom_sample_info_table:
+      sample_project_id, sample_dataset_id, sample_table_id = (
+          bigquery_util.parse_table_reference(
+              parsed_args.custom_sample_info_table))
+    else:
+      sample_project_id = project_id
+      sample_dataset_id = dataset_id
```
samanvp, line 673: nit: Let's move these two lines to where `sample_table_id` is initiated.

samanvp (approved):
Thanks for clean PR.

gcp_variant_transforms/options/variant_transform_options.py
tneymanov: Done.

tneymanov (commented):
Will run int tests and submit afterwards. Thanks.

PR #697, opened 2021-05-14
update subnetwork flag documentation

rcowin-gcp:


moschetti (approved):
LGTM

moschetti:
Added commits to disable pylint `consider` flags as this PR is only changing docs. Can revisit implementation changes for those later.

PR #709, opened 2022-01-21
Fix yaml.load error

kbtian:
yaml.load as currently used was deprecated and requires specification of a loader.

### DIFF
--- a/gcp_variant_transforms/libs/variant_sharding.py
+++ b/gcp_variant_transforms/libs/variant_sharding.py
@@ -137,7 +137,8 @@ class VariantSharding():
     has_any_interval = False
     with FileSystems.open(config_file_path, 'r') as f:
       try:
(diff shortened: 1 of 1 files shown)


gcp_variant_transforms/libs/variant_sharding.py
```diff
@@ -137,7 +137,8 @@ def _validate_config_and_check_intervals(self, config_file_path):
     has_any_interval = False
     with FileSystems.open(config_file_path, 'r') as f:
       try:
-        shards = yaml.load(f)
+        yaml.warnings({'YAMLLoadWarning': False})
```
moschetti, line 140: If we fixed the issue, why are we also disabling the warning?

moschetti (approved):

lawrenae (approved):

PR #713, opened 2022-03-03
Pipline mode flag

kbtian:
Add a pipeline mode flag to override existing behavior. 

Default set to '' which will use existing behavior, but can be set to 'small', 'medium', or 'large' to force a pipeline mode.

### DIFF
--- a/gcp_variant_transforms/pipeline_common_test.py
+++ b/gcp_variant_transforms/pipeline_common_test.py
@@ -51,7 +51,7 @@ class PipelineCommonWithPatternTest(unittest.TestCase):
           input_pattern='nonexistent_file.vcf', input_file=None)
 
   def test_get_mode_small_pipeline_mode(self):
-    args = self._create_mock_args(input_pattern='*', input_file=None, pipeline_mode='medium')
+    args = self._create_mock_args(input_pattern='*', input_file=None, pipeline_mode='small')
     match_result = collections.namedtuple('MatchResult', ['metadata_list'])
     match = match_result([None for _ in range(100)])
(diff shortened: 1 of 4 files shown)


gcp_variant_transforms/pipeline_common_test.py
```diff
@@ -41,24 +41,46 @@ def _create_mock_args(self, **args):
...
+  def test_get_mode_large_pipeline_mode(self):
+    args = self._create_mock_args(input_pattern='*', input_file=None, pipeline_mode='large')
+    match_result = collections.namedtuple('MatchResult', ['metadata_list'])
+    match = match_result([None for _ in range(100)])
+    with mock.patch.object(FileSystems, 'match', return_value=[match]):
+      self.assertEqual(self._get_pipeline_mode(args), PipelineModes.LARGE)
+
   def test_get_mode_small(self):
-    args = self._create_mock_args(input_pattern='*', input_file=None)
+    args = self._create_mock_args(input_pattern='*', input_file=None, pipeline_mode='')
```
lawrenae, line 75: totally a nit: you don't need to pass `pipeline_mode` at all here as the argument is defaulted to the same, and this test (and the next few) ensure existing behavior continues, right?

lawrenae (changes requested):

gcp_variant_transforms/pipeline_common_test.py
kbtian: Yes, the tests are to ensure both existing and new behavior. The `pipeline_mode` needs to be passed in for test setup.

lawrenae (approved):
LGTM

PR #727, opened 2025-06-27
Upgrade dependencies and migrate to Batch API

ptdtan:
Upgraded the dependencies, including apache-beam to latest version 2.65.0, and python 3.12

- Built latest VEP docker image using latest VEP
- Migrated all the Life Science APIs to Batch API, including:
- Implemented the Variant-Effect-Prediction (VEP) pipeline using Batch API
- Implement the end-to-end pipeline runner, which is currently used in docker, via Batch API
- Implement a new VCFParser, based on the old Pysam-based parser. This fix is to avoid nested multiprocessing in the Dataflow environment, which causes workers disconnected.


pgrosu:
Hi Tan (@ptdtan),

Thank you for helping out the project and community with this nice update!

Paul

jairobe (approved):
LGTM

